Issue #65, opened 2013-02-24
Keys should not allow arbitrary characters: they should be `[a-zA-Z_]\w*` only


mrflip:
I'm surprised that keys are allowed to be arbitrary characters. What is the configuration file use case for this? Can those cases be just as well handled as a literal hash, or as an array of pairs? Allowing funny characters seems to go against the "simple as possible" ethos.

I think of keys as living in the control path, not data path, and thus predictability should win over expressiveness (no matter how fun it would to configure `☃.♛=シ`). I can't say for sure what would go wrong, but allowing nulls, vertical tab characters, funny unicode spaces and so forth sounds like an eventual security flaw. Unicode opens a lot of "there should only be one way to do anything" holes: for just one example, the strings `dīáçṙïtĭč` and `dīáçṙïtĭč` are semantically equivalent but not byte-comparable (one has combining diacritics, the other precomposed). Are two keys equal if they character comparable, or only if they are byte-comparable? 

The spec should require keys to be identifier-like: they must start with a letter or underbar, and contain only letter, underbar or number. That is:  `[a-zA-Z_]\w*`.

mojombo:
Resolved by #283.

PR #67, opened 2013-02-24
Keys should be identifier-like (addresses #65)

mrflip:
Keys should contain only letters, numbers or underscore (`_`) characters, and should
start with a letter or underscore. See #65 for discussion.


mojombo:
An alternative was merged in #283.

Issue #76, opened 2013-02-24
[RFC] adding a Schema validation


liuggio:
A great feature of XML is that you could validate a given XML file with another file.
In that file you could define the schema of the file with default values, optional values, children, data types etc.

In some env the schema is a killer feature, its usage should be suggested as best practice.

The schema file could be called 
TOSF - Schema File
TOSD - as XSD
TOLO - Liuggio Obvious

PR #116, opened 2013-02-25
Added Schema validator specification

liuggio:
see #76 

TOLS is like XSD, only more readable and simpler.

There are 3 simple rules to follow:
- A TOLS file is itself a valid TOML file.
- Each element could be validated.
- Explicit validation - if an element has not a validation scheme is valid by default.

Todo:
-  [ ] gather feedback about Specification
- [ ] gather feedback about Keywords
- [ ] gather feedback about Usages


liuggio:
:+1:

WiSaGaN:
Are we progressing toward 1.0? It seems to me if we are not going to be ready for 1.0 soon, this schema validator would be something useful to add.

liuggio:
@WiSaGaN if you think this will be merged I'll do a code rewiew and update and speed up this Issue 
But I think this will not be merged :) anyway

breerly:
Any chance of getting renewed interest here?

golddranks:
I wonder if there were any value in using a readymade solution for JSON, like JSON Schema? I'm not familiar enough with TOML so I don't know whether it roundtrips through JSON well enough, but if it does, I think that would be a simple way to add schema validation.

golddranks:
Anyway, having a schema language would be great in the sense that it would allow configuration file validation and autocompletion using external, generalised tools. I'd be more than happy to have that!

adamvoss:
You can now validate TOML (v0.4) documents against a JSON Schema using [**pajv**](https://www.npmjs.com/package/pajv).  I suppose you could even write the schema in TOML if you want then use [**any-json**](https://www.npmjs.com/package/any-json) to convert it to JSON.

I have documented the [implementation status of JSON Schema validation for TOML](https://json-schema-everywhere.github.io/toml).  The main issue I can think of with the available implementations is that they are JavaScript-based and thus cannot differentiate between `number` and `integer`.

@golddranks You can use any-json to experiment with round-tripping.

vietlq:
How is this going? I'm quite interested in using schema for TOML :)

AndreiPashkin:
What do everybody think about having ability to specify default values as a reference to other values in the config in schema?

eksortso:
@AndreiPashkin Well, defaults could be useful. But including defaults goes beyond simple schema validation. So I wouldn't recommend it for TOLS, at least right now.

But I can imagine an extension that could contain all sorts of data properties that are not natively handled by TOML (such as defaults, cardinality requirements, valid alternatives, etc.). A document with such features could be fed into a meta-parser, which then would handle some of the common but bothersome tasks that processing config files involves.

Save defaults for later, I'd say. Stick to the essentials in your spec, and release that as TOLS v1.0. Once it's released, you can then see how validations are being done with it, then add useful stuff later on.

LongTengDao:
Could we use TypeScript-like things? Now mostly program API docs are described in TypeScript format, which is expressive for type verify:

```typescript
import { String, Integer } from 'spec';

type table = {
  key? :String | Integer | true
};
```

Or:

```
[table]
key? = String | Integer
```

I think we can't get a nice validator, unless we add `|` '?' operators, and change `[table]` `[[array-of-tables]]` to inline notation, because here needs show branches.

verdie-g:
Hello, I'm a big fan of the XSD and I'm looking for something similar in toml. To try the robustness of tols I have tried writing tols' tols.

Maybe I missed something but First problem I encountered was that I don't know how to write a schema validating dynamic key names like we find in tols:
```toml
[owner] # the key can be anything
[database]
...
```

So I've skipped this part and jumped right into the definition of a field:
```toml
[primitive]
  primitive = "String"
  [primitve.range]
  in = ["String", "Integer", "Float", "Boolean", "Datetime", "Array", "Hash"]

[default]
# primitive = ??

[required]
  primitive = "Boolean"
  default = false

[length]
  primitive = "Hash"
  [length.min]
  primitive = "Integer"
  default = 0
  [length.max]
  primitive = "Integer"
  default = 9_223_372_036_854_775_807 # max 64 bit signed integer

[range]
  primitive = "Hash"
  [range.min]
  # primitive = ??
  [range.max]
  # primitive = ??
  [range.in]
    primitive = "Array"
    [range.in.content]
    # primitive = ??

[pattern]
  primitive = "String"
  pattern = "/^((?:(?:[^?+*{}()[\]\\|]+|\\.|\[(?:\^?\\.|\^[^\\]|[^\\^])(?:[^\]\\]+|\\.)*\]|\((?:\?[:=!]|\?<[=!]|\?>)?(?1)??\)|\(\?(?:R|[+-]?\d+)\))(?:(?:[?+*]|\{\d+(?:,\d*)?\})[?+]?)?|\|)*)$/" # https://stackoverflow.com/a/172316/5407910

[content]
  primitive = "Hash"
  [content.content]
  # recursive. need a way to reference a custom type

[occurence]
  primitive = "Hash"
  [occurence.min]
  primitive = "Integer"
  default = 1 # not documented so I chose 1 like the XSD
  [occurence.max]
  primitive = "Integer"
  default = 1
```

Some parts can't be validated:

- root elements
- why Hash when it is a Table
- occurence defaults
- add description field
- inclusive instead of exclusive
- what about Local Date-Time, Local Date, Local Time ?
- I would rename primitive to type
- doesn't support recursive structure
- doesn't support inheritance

verdie-g:
Hello, I'm a big fan of the XSD and I'm looking for something similar in toml. To try the robustness of tols I have tried writing tols' tols.

Maybe I missed something but the first problem I have encountered was that I don't know how to write a schema validating dynamic key names like we find in tols:
```toml
[owner] # the key can be anything
[database]
...
```

So I've skipped this part and jumped right into the validation rules of an element:
```toml
[primitive]
  primitive = "String"
  [primitve.range]
  in = ["String", "Integer", "Float", "Boolean", "Datetime", "Array", "Hash"]

[default]
# primitive = ??

[required]
  primitive = "Boolean"
  default = false

[length]
  primitive = "Hash"
  [length.min]
  primitive = "Integer"
  default = 0
  [length.max]
  primitive = "Integer"

[range]
  primitive = "Hash"
  [range.min]
  # primitive = ??
  [range.max]
  # primitive = ??
  [range.in]
    primitive = "Array"
    [range.in.content]
    # primitive = ??

[pattern]
  primitive = "String"
  pattern = "/^((?:(?:[^?+*{}()[\]\\|]+|\\.|\[(?:\^?\\.|\^[^\\]|[^\\^])(?:[^\]\\]+|\\.)*\]|\((?:\?[:=!]|\?<[=!]|\?>)?(?1)??\)|\(\?(?:R|[+-]?\d+)\))(?:(?:[?+*]|\{\d+(?:,\d*)?\})[?+]?)?|\|)*)$/" # https://stackoverflow.com/a/172316/5407910

[content]
  primitive = "Hash"
  [content.content]
  # recursive. need a way to reference a custom type

[occurence]
  primitive = "Hash"
  [occurence.min]
  primitive = "Integer"
  default = 1 # not documented so I chose 1 like the XSD
  [occurence.max]
  primitive = "Integer"
  default = 1
```

Tols has some limitations and can't validate some parts of tols:
1. Length only for String, Array, Hash
```toml
[foo]
primitive = "Integer"
[foo.length] # ???
```

2. Range only for Integer, Float, Datetime
```toml
[foo]
primitive = "Array"
[foo.range] # ???
```

3. Pattern only for String
```toml
[foo]
primitive = "Array"
pattern = "/.../" # ???
```

4. Range, default, in/notin primitives
```toml
[foo]
primitive = "Integer"
default = "xyz" # ???

[foo]
primitive = "Integer"
in = ["xyz"] # ???
```

5. Array subtable's keys not being integers
```toml
[foo]
primitive = "Array"
[foo.abc] # ???
```

6. String, Integer, Float, Boolean having subtables
```toml
[foo]
primitive = "Integer"
[foo.0] # no effect
```

Also there are several points that are missing or that bother me:
7. As I said I don't know how validate a table with key name unknown
8. Occurence defaults are not documented
9. Hash should be renamed to Table like in toml's spec
10. Min and max are exclusive which I think is extremely confusing
11. Local Date-Time, Local Date, Local Time from toml are missing
12. Primitive could be renamed to type
13. Recursive structure are not supported so it is not possible to write tols' schema
14. Inheritance could be a nice feature, it is something I use a lot with the XSD
15. `if an element has not a validation scheme is valid by default`. Does it mean if there is a typo in a key, the validator won't raise an error because there is no validation rules found?

muuvmuuv:
If it has not been written somewhere I would like to see editor autocompletion too, like with JSON `$schema`. A URL which holds the XSD/TOLS file to validate against.

PR #215, opened 2014-04-22
add a link to Chinese translation

weakish:
Add a link to Chinese translation of TOML README.


mojombo:
@weakish This is awesome, thanks! But do you have the raw Markdown available? And do you think you could update it for the 0.3.1 spec?

I've created two-letter-language-code subdirs in the repo to hold translations. This seems like the best solution to me at the moment.

mojombo:
Indeed, translations go into the `versions/LANGCODE` directory these days, and an 0.4 Chinese version was just merged.

Issue #219, opened 2014-06-24
Tuples and Inline Tables: A Motivation


wycats:
This builds on https://github.com/toml-lang/toml/pull/154, which would be very useful regardless of whether this proposal is accepted.

> Hurray for tuples! @mojombo 

Currently, there is no syntax for inline tables, which means that you're forced to create a new section for simple cases. In Cargo:

``` toml
[dependencies.hammer]

version = "1.0.0"
git = "https://github.com/wycats/hammer.rs"
```

We have a shorthand for the case where a dependency only needs to specify a version:

``` toml
[dependencies]

hammer = "1.0.0"
```

With tuples, we could do something like:

``` toml
[dependencies]

hammer = ("1.0.0", "git", "https://github.com/wycats/hammer.rs")
```

I would like to propose a syntax for inline tables:

``` toml
[dependencies]

hammer = ("1.0.0", git: "https://github.com/wycats/hammer.rs")
# or alternately
hammer = ("1.0.0", git = "https://github.com/wycats/hammer.rs")
```

I would be comfortable with restricting inline tables to the last element of a tuple, to eliminate the need for additional braces or something around the table.

PR #235, opened 2014-07-16
Add inline table syntax.

mojombo:
Concrete proposal for the inline table syntax I first proposed in Issue #219. Here's a quick taste:

```
# They should look familiar and obvious
name = { first = "Tom", last = "Preston-Werner" }
point = { x = 1, y = 2 }

# They make tuples unnecessary by allowing mixed types (they are just tables, after all).
# Unlike tuples, they are self-documenting via the key names.
address = { proto = "http", ip = "10.0.0.1", port = 8080 }

# Arrays of tables with less hassle
points = [ { x = 1, y = 2, z = 3 },
           { x = 7, y = 8, z = 9 },
           { x = 2, y = 4, z = 8 } ]
```

This is about as brief as I could think to make it, so if there are specific issues that need clarifying, please bring them up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,6 +363,39 @@ All table names and keys must be non-empty.
  = "no key name" # not allowed
 ```
 
+Inline Table
+------------
+
+Inline tables provide a more compact syntax for expressing tables. They are
+especially useful for grouped data that can otherwise quickly become verbose.
+Inline tables are enclosed in curly braces `{` and `}`. Within the braces, zero
+or more comma separated key/value pairs may appear. Key/value pairs take the
+same form as key/value pairs in standard tables. All value types are allowed,
+including inline tables.
+
+Inline tables are intended to appear on a single line. No newlines are allowed
+between the curly braces unless they are valid within a value. Even so, it is
+strongly discouraged to break an inline table onto multiples lines. If you find
+yourself gripped with this desire, it means you should be using standard tables.
+
+```toml
+name = { first = "Tom", last = "Preston-Werner" }
+point = { x = 1, y = 2 }
+```
+
+The inline tables above are identical to the following standard table
+definitions:
+
+```toml
+[name]
+first = "Tom"
+last = "Preston-Werner"
+
+[point]
+x = 1
+y = 2
+```
+
 Array of Tables
 ---------------
 
@@ -462,6 +495,14 @@ array must produce an error at parse time.
   # This table conflicts with the previous table
   [fruit.variety]
     name = "granny smith"
+```
+
+You may also use inline tables where appropriate:
+
+```toml
+points = [ { x = 1, y = 2, z = 3 },
+           { x = 7, y = 8, z = 9 },
+           { x = 2, y = 4, z = 8 } ]
 ```
 
 Seriously?


mojombo:
This feature is the last big thing to make a decision on before TOML 1.0. I've been hesitant to merge it because of the reasonable feedback and because it still doesn't feel quite as elegant as I strive for. That being said, it's still very likely to go in.

It's been a while since this was proposed. @wycats @alexchrichton - Has anything changed from a Cargo perspective? Have the last 6 months given you any more precise insight into whether this proposal still meets your needs well? I'd love an updated comment about the matter.

alexcrichton:
I don't think we've necessarily gained much more insight since 6 months ago, but our usage pattern has definitely changed quite a bit. Six months ago most `Cargo.toml` manifests looked like:

```toml
[dependencies.foo]
git = "https://github.com/foo/foo"

[dependencies.bar]
git = "https://github.com/bar/bar"
```

Most `Cargo.toml` manifests today look like:

```toml
[dependencies]
foo = "0.1"
bar = "0.2.3"
```

The actual syntax of declaring dependencies hasn't actually changed (for us these differences are just declaring dependencies from different sources). One interesting point we've seen, though, is that in isolation both of these syntaxes are pretty reasonable, but when put together the result is a little jarring:

```toml
[dependencies]
foo = "0.1.0"

[dependencies.bar]
git = "https://github.com/bar/bar"
```

For contrast, the mixed-source dependencies with inline tables would look like:

```toml
[dependencies]
foo = "0.1.0"
bar = { git = "https://github.com/bar/bar" }
```

As a personal opinion, I find the second inline table syntax flows quite nicely as it's easy to scan what the list of dependencies are without having to worry about where they're actually coming from.

So all in all, in terms of insights I don't think we've turned up much new (our syntax hasn't changed much since day 1). I think it'd be summed up by "this would still be quite nice for us" :)

wycats:
Just to echo what @alexcrichton said, the migration towards a registry for Cargo has had the expected results here. Mixing in a few `git` dependencies with a larger number of registry dependencies is awkward (as expected), and this would be nice for us.

mwanji:
As it looks like this might go in, I'm going to argue against it, for 3 reasons: overloaded tables, complexity and I don't think even Cargo needs it.

**Cargo**
The driving usecase is Cargo, but its own syntax has improved enough that inline tables might not add much benefit, but would add a lot of burden to TOML. I've only just read the [Cargo manifest definition](http://doc.crates.io/manifest.html), but it seems like at least some of the akwardness could be fixed by making Cargo a bit more flexible (please pardon my backseat programming).

For example:

```toml
[dependencies]
foo = "0.1.0"

[dependencies.bar]
git = "https://github.com/bar/bar"

[dependencies.baz]
path = "./path/to/baz"
```

could be

```toml
[dependencies]
foo = "0.1.0"
bar = "https://github.com/bar/bar"
baz = "./path/to/baz"
hammer = "github:wycats/hammer/wip" # I'm making this one up: scheme + repo ID + branch or sha-1
```
This seems fairly clear and in line with Cargo's evolution, without requiring the entire TOML ecosystem to support inline tables. Further, quoted keys can help provide support for dependency names with dots in them.

**Overloaded tables**
There are several ways of defining tables (normal, implicit, table arrays). The combinations can already be a bit tricky, adding another way makes it even trickier.

**Complexity**
Datetimes are seriously [being considered for removal](https://github.com/toml-lang/toml/issues/263), yet seem far simpler than inline tables.

kardianos:
I have looked into toml for my line of business configuration files. I intend to use it, especially once the spec gets nailed down to v1.  I'm a fan of how Go is optimized readability over totally simplest way to write something. 

I did a comparison with the above examples in my text editor. I think toml's advantage is that it repeats itself. If you don't want to repeat section headers you should use json or xml. The repeated section names were much easier to read.

I'm strongly in favor of not including this.

My use case is primarily application configuration files. Example information include http ports,  database connection strings, host name binding, and filepaths. I don't want the user to see or use "{}" or have someone later decide to "take advantage" of this feature, confusing end users later. That's why I'd choose toml in the first place. Believe it or not, nesting and comma separation is surprisingly confusing to many people.

It sounds like cargo may be better off with json then toml given the arguments. After all, programmers will be using cargo, not users.

-Daniel

pnathan:
> I don't want the user to see or use "{}" or have someone later decide to "take advantage" of this feature, confusing end users later. 

+1. A major use case for me is being able to build tools and utilities, then distribute them to people without having to worry that "forgot to put a }" becomes a support case root cause.  INI is a revealed preference for many many people; leaning towards that is very nice.

wycats:
Encoding all of the options for a dependency in a cryptic string micro-syntax is not an overall simplification. 

You can, of course, encode arbitrary data in strings, but when I start going down that path, I always ask myself whether an options hash would would be better.

mwanji:
> Encoding all of the options for a dependency in a cryptic string micro-syntax is not an overall simplification.

Sure, the last example probably goes overboard. However, I think the other three are in line with Cargo's evolution. In any case, my main point was to try to show that having to list a dependency's properties explicitly could be made a relatively rare situation.

mojombo:
One of the most important things that TOML will ever have is readability. In most cases obviousness and minimalism will align with this goal. But sometimes minimalism must be sacrificed in order to create something that is powerful and expressive enough to handle reality. Cargo represents one very powerful use case, that of a structure like A → B → C, where A is a table of many Bs, and each B is a table of few Cs. This circumstance is not currently handled well by TOML. And that's a shame, because I anticipate that the pain this creates will affect a lot of people. **After spending some time trying out various config files with and without this PR, I've decided that TOML needs it.** Fortunately, those that don't need this extra bit of power don't need to use it or present it to users. But those that DO need it will find their config files much more delightful and intuitively grouped.

Thank you to everyone that weighed in on this issue. Your insights and experiences have been extremely valuable.

mwanji:
Do inline tables support trailing commas, like arrays?

mojombo:
@mwanji No, not with the current spec.

mwanji:
Is it necessary to mention that bare & quoted keys can be used within inline tables, or is that too obvious?

mojombo:
@mwanji I think that's suitable addressed by "Key/value pairs take the same form as key/value pairs in standard tables."

mwanji:
Nested tables make seeing which tables have been defined a lot less obvious.

It's an issue that I don't think has been raised and that I came across while adding inline tables to my parser. For example: 
```toml
a = { a1 = 1, b = { b1 = 2, c = { c1 = 3 }}}

[a.b] # this is now illegal
[a.b.c] # this is now illegal
```
Are there cases where a nested inline table looks better than the non-nested syntax?

PR #236, opened 2014-07-18
Add ABNF description of TOML.

mojombo:
One of the most important things we'll need for the TOML 1.0 release is a proper grammar. I've worked one up here as a starting point. I chose RFC 4234 ABNF because it feels much more modern and suitable for a Unicode world than ISO EBNF. Also, JSON uses ABNF and so many people should already be familiar with it.

This grammar should be complete. It also includes scientific notation for floats, as I expect those will go in soon.

I'm certainly no ABNF expert, and I'd love any suggestions for making this grammar more readable.

### DIFF
--- a/toml.abnf
+++ b/toml.abnf
@@ -1,34 +1,38 @@
+;; WARNING: This document is a work-in-progress and should not be considered
+;; authoritative until further notice.
+
 ;; This is an attempt to define TOML in ABNF according to the grammar defined
 ;; in RFC 5234 (http://www.ietf.org/rfc/rfc5234.txt).
 
-;; WARNING: This document is a work-in-progress and should not be considered
-;; authoritative until further notice.
+;; You can try out this grammar interactively via the online ABNF tool at
+;; http://www.coasttocoastresearch.com/interactiveapg
+;; Note that due to the limitations of ABNF parsers, in order for multi-line
+;; strings to work in that tool, the following rules must be ammended to
+;; disallow the use of unescaped double- or single-quotes:
+;; ml-basic-unescaped = basic-unescaped
+;; ml-literal-char = literal-char
 
 ;; TOML
 
 toml = expression *( newline expression )
-expression = (
-  ws /
-  ( ws comment ) /
-  ( ws keyval ws [ comment ] ) /
-  ( ws table ws [ comment ] )
-)
+expression = ( ( ws comment ) /
+               ( ws keyval ws [ comment ] ) /
+               ( ws table ws [ comment ] ) /
+                 ws )
 
 ;; Newline
 
-newline = (
-  %x0A /              ; LF
-  %x0D.0A             ; CRLF
-)
+newline = ( %x0A /              ; LF
+            %x0D.0A )           ; CRLF
 
 newlines = 1*newline
 
 ;; Whitespace
 
-ws = *(
-  %x20 /              ; Space
-  %x09                ; Horizontal tab
-)
+wschar = ( %x20 /              ; Space
+           %x09 )              ; Horizontal tab
+
+ws = *wschar
 
 ;; Comment
 
@@ -45,7 +49,7 @@ key = unquoted-key / quoted-key
 unquoted-key = 1*( ALPHA / DIGIT / %x2D / %x5F ) ; A-Z / a-z / 0-9 / - / _
 quoted-key = quotation-mark 1*basic-char quotation-mark ; See Basic Strings
 
-val = integer / float / string / boolean / date-time / array / inline-table
+val = string / boolean / array / inline-table / date-time / float / integer
 
 ;; Table
 
@@ -77,7 +81,7 @@ int = DIGIT / digit1-9 1*( DIGIT / underscore DIGIT )
 
 ;; Float
 
-float = integer ( frac / frac exp / exp )
+float = integer ( frac / ( frac exp ) / exp )
 zero-prefixable-int = DIGIT *( DIGIT / underscore DIGIT )
 frac = decimal-point zero-prefixable-int
 decimal-point = %x2E               ; .
@@ -86,7 +90,7 @@ e = %x65 / %x45                    ; e E
 
 ;; String
 
-string = basic-string / ml-basic-string / literal-string / ml-literal-string
+string = ml-basic-string / basic-string / ml-literal-string / literal-string
 
 ;; Basic String
 
@@ -112,7 +116,7 @@ escape = %x5C                    ; \
 
 ;; Multiline Basic String
 
-ml-basic-string-delim = quotation-mark quotation-mark quotation-mark
+ml-basic-string-delim = 3quotation-mark
 ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
 ml-basic-body = *( ml-basic-char / newline / ( escape newline ))
 
@@ -121,15 +125,15 @@ ml-basic-unescaped = %x20-5B / %x5D-10FFFF
 
 ;; Literal String
 
-literal-string = apostraphe *literal-char apostraphe
+literal-string = apostrophe *literal-char apostrophe
 
-apostraphe = %x27 ; ' Apostraphe
+apostrophe = %x27 ; ' apostrophe
 
 literal-char = %x09 / %x20-26 / %x28-10FFFF
 
 ;; Multiline Literal String
 
-ml-literal-string-delim = apostraphe apostraphe apostraphe
+ml-literal-string-delim = 3apostrophe
 ml-literal-string = ml-literal-string-delim ml-literal-body ml-literal-string-delim
 
 ml-literal-body = *( ml-literal-char / newline )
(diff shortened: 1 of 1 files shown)


flaviut:
Is there a reason that `01` is invalid? It seems like an arbitrary restriction for which I cannot see any justification.

toml.abnf
```diff
@@ -0,0 +1,185 @@
...
+
+date-fullyear  = 4DIGIT
+date-month     = 2DIGIT  ; 01-12
+date-mday      = 2DIGIT  ; 01-28, 01-29, 01-30, 01-31 based on month/year
+time-hour      = 2DIGIT  ; 00-23
+time-minute    = 2DIGIT  ; 00-59
+time-second    = 2DIGIT  ; 00-58, 00-59, 00-60 based on leap second rules
+time-secfrac   = "." 1*DIGIT
+time-numoffset = ( "+" / "-" ) time-hour ":" time-minute
+time-offset    = "Z" / time-numoffset
```
flaviut: It should be clarified that `Z` can be uppercase or lowercase:

> NOTE: Per [ABNF] and ISO8601, the "T" and "Z" characters in this
> syntax may alternatively be lower case "t" or "z" respectively.

Same with `<exp>` and `<date-time>`.

toml.abnf
```diff
@@ -0,0 +1,185 @@
...
+
+partial-time   = time-hour ":" time-minute ":" time-second [time-secfrac]
+full-date      = date-fullyear "-" date-month "-" date-mday
+full-time      = partial-time time-offset
+
+date-time      = full-date "T" full-time
```
flaviut: > NOTE: ISO 8601 defines date and time separated by "T".
> Applications using this syntax may choose, for the sake of
> readability, to specify a full-date and full-time separated by
> (say) a space character.

Since non-technical users are part of the target audience, the improved readability might be appreciated. Should I make a separate issue for this (the specification is somewhat ambiguous here)?

toml.abnf
BurntSushi: Agreed.

toml.abnf
BurntSushi: Yes, please make a new issue.

BurntSushi:
> Is there a reason that `01` is invalid? It seems like an arbitrary restriction for which I cannot see any justification.

Is there a reason why it should be allowed? In many places, prefixing a number with a `0` is an instruction that the number should be interpreted as octal. TOML doesn't support octal, so it seems surprising to allow it.

Moreover, if TOML were ever to adopt octal numerals, this restriction would allow that to be a backwards compatible addition.

dhardy:
@BurntSushi out with the old `0755` octal notation. If you need octal, use something less ambiguous like `0c755`.

flaviut:
I agree with @dhardy. `0777` is terrible for beginners and confusing for everyone else.

BurntSushi:
`0777` is the prevailing convention.

BurntSushi:
(And that's a good enough reason to bar integrrs from starting with a `0` IMO. And my backwards compatible argument is still viable, because it gives us the flexibility to choose the syntax.)

toml.abnf
```diff
@@ -0,0 +1,185 @@
...
+
+table = std-table / array-table
+
+;; Standard Table
+
+std-table-open  = %x5B ws     ; [ Left square bracket
+std-table-close = ws %x5D     ; ] Right square bracket
+table-key-sep   = ws %x2E ws  ; . Period
+
+std-table = std-table-open key *( table-key-sep key) std-table-close
```
flaviut: This rule will match `[]`, which is expressly forbidden in the spec.

flaviut: Using `<key>` in this way does not follow the spec, which makes no mention of quoted table names.

toml.abnf
```diff
@@ -0,0 +1,185 @@
...
+basic-char = basic-unescaped / escaped
+escaped = escape ( %x22 /          ; "    quotation mark  U+0022
+                   %x5C /          ; \    reverse solidus U+005C
+                   %x2F /          ; /    solidus         U+002F
+                   %x62 /          ; b    backspace       U+0008
+                   %x66 /          ; f    form feed       U+000C
+                   %x6E /          ; n    line feed       U+000A
+                   %x72 /          ; r    carriage return U+000D
+                   %x74 /          ; t    tab             U+0009
+                   %x75 4HEXDIG )  ; uXXXX                U+XXXX
```
flaviut: This omits
> `\UXXXXXXXX - unicode         (U+XXXXXXXX)`

toml.abnf
mojombo: @flaviut I don't see how it will match `[]`. The `<key>` rule mandates that at least 1 character be present. Also, see #283 which clarifies key names to match the rules present here.

toml.abnf
mojombo: Good catch. Will fix.

toml.abnf
flaviut: My bad then, sorry. It seems like I read over the first `<key>` without noticing it.

mojombo:
A few updates on this:

1. I've spent some time writing a PEG definition of TOML to see if that would be better to use than ABNF. I like the ideas behind PEGs a lot, but I don't think it's a suitable grammar format. There does not seem to be any official RFC-like standard for PEGs and the [Bryan Ford paper](http://pdos.csail.mit.edu/papers/parsing:popl04.pdf) is woefully lacking in details like how to specify ranges of Unicode characters, specific numbers of character repetition, etc. I haven't seen a single real-world example of it being used as an official definition, anywhere, ever. So I'm going to stick with ABNF.
2. I'd like to merge the various parts of the ABNF into the full spec soon, so if anyone has additional changes to propose before that happens, now is the time!

alexcrichton:
Currently this add syntax for inline tables, so I just wanted to confirm, would you want to add that now or merge that part of the ABNF at a possible future date?

This is also super helpful @mojombo, thanks for doing this!

mojombo:
@alexcrichton I added that just to see how complex it would be. I'd merge that into the spec when Inline Tables go in.

toml.abnf
```diff
@@ -0,0 +1,187 @@
...
+;; Basic String
+
+basic-string = quotation-mark *basic-char quotation-mark
+
+quotation-mark = %x22            ; "
+
+basic-char = basic-unescaped / escaped
+escaped = escape ( %x22 /          ; "    quotation mark  U+0022
+                   %x5C /          ; \    reverse solidus U+005C
+                   %x2F /          ; /    solidus         U+002F
```
jakwings: Why `\/`?

toml.abnf
```diff
@@ -0,0 +1,187 @@
...
+ml-basic-string-delim = quotation-mark quotation-mark quotation-mark
+ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
+ml-basic-body = *( ml-basic-char / newline / ( escape newline ))
+
+ml-basic-char = ml-basic-unescaped / escaped
+ml-basic-unescaped = %x20-5B / %x5D-10FFFF
+
+;; Literal String
+
+literal-string = apostraphe *literal-char apostraphe
```
avakar: This should be spelled "apostrophe".

toml.abnf
```diff
@@ -0,0 +1,187 @@
...
+
+;; Array
+
+array-open  = %x5B ws  ; [
+array-close = ws %x5D  ; ]
+
+array = array-open array-values array-close
+
+array-values = [ val [ array-sep ] [ ( comment newlines) / newlines ] /
+                 val array-sep [ ( comment newlines) / newlines ] array-values ]
```
avakar: This disallows newline before the first element, e.g.

    # this is not allowed
    key = [
        1]

but allows it after:

    # this is ok
    key = [1
        ]

Is that intended?

impinball: Hopefully not. Could this be fixed? (It's surprising behavior)

andrusha97: It also disallows newlines between values and commas. Is it ok?

impinball: @mojombo Ping...

toml.abnf
```diff
@@ -0,0 +1,187 @@
...
+
+inline-table = inline-table-open inline-table-keyvals inline-table-close
+
+inline-table-keyvals = [ inline-table-keyvals-non-empty ]
+inline-table-keyvals-non-empty = key keyval-sep val /
+                                 key keyval-sep val inline-table-sep inline-table-keyvals-non-empty
+
+;; Built-in ABNF terms, reproduced here for clarity
+
+; ALPHA = %x41-5A / %x61-7A ; A-Z / a-z
```
TerjeBr: Why not use unicode `ALPHA`? Allow characters like 'Æ', 'ø', 'å', 'ß' to be in keywords. Allow f.ex. Arabic or Hebrew to be used in key words.

isiahmeadows:
@mojombo Ping?

toml.abnf
kargakis: Any news on this?

pradyunsg:
@mojombo Ping 2.0?

mojombo:
I'm pretty happy with the ABNF now, and I think it'll be super helpful for implementers. Time to merge! From now on, we should update the ABNF any time the TOML spec changes.

Bina1791 (approved):

Issue #265, opened 2014-11-11
Update examples to include entire version coverage.


leonelquinteros:
Example files on /tests directory doesn't seems to cover all features on last version (0.3.0). 

It would be nice to extend these to put examples of everything that any parser should handle for each version.

mojombo:
Closing because #266 was merged in.

PR #266, opened 2014-11-11
Example with full version coverage

leonelquinteros:
This is what i meant on https://github.com/toml-lang/toml/issues/265

Doing this i figured out that you are able to define the entire language using a TOML file. That would feel like writing a [LANG] compiler in [LANG] language to me! =)

But I'll let you that for you. Hope this one helps.


mojombo:
Nice. I like it. I'll probably move it to an `examples` directory, though.

PR #269, opened 2014-11-29
Specify nanosecond precision for datetimes

groves:
As far as I can tell, RFC 3339 doesn't specify a maximum precision for the fractional seconds portion of the time. This is tough when parsing, as if I have to support arbitrarily precise times, I can't store the fractional seconds in a fixed size field: it's possible for it to be larger than an int, or a long, or whatever else.

This change makes nanoseconds the maximum precision. With that I can store the fractional portion in a 32 bit int.

Nanoseconds seemed precise enough to me as that's the highest precision I've seen returned by system clocks and the highest precision I've seen returned by databases. Milliseconds could also be interesting since that'd allow the whole datetime to be stored in a long, but I don't think it's worth restricting some use that wants to store the extra precision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ false
 Datetime
 --------
 
-Datetimes are [RFC 3339](http://tools.ietf.org/html/rfc3339) dates.
+Datetimes are [RFC 3339](http://tools.ietf.org/html/rfc3339) dates with nanosecond precision.
 
 ```toml
 1979-05-27T07:32:00Z


mojombo:
I'm hesitant to make nanosecond resolution a requirement as many platforms/languages still do not support that level of granularity. This might be a case where each parser needs to determine what's reasonable and report an error if a user tries to be more specific than can be handled.

groves:
I didn't intend to require nanosecond precision as much as put an upper bound on the precision. That way as a parser author I can just handle all valid TOML files and don't have to qualify that if they use greater-than-nanos precision, I blow up. 

Using less precise times or omitting fractional seconds altogether is still fine in my use case. What if I changed the language to "Datetimes are [RFC 3339](http://tools.ietf.org/html/rfc3339) dates with up to nanosecond precision"?

mojombo:
It seems strange to restrict something like this in the spec. I've set minimum requirements for integers and floats to give authors an assurance that a specific number will always be accepted by parsers, but if a decade from now we're all using 128 bit machines and everyone wants to store those items in 128 bits, so much the better. I feel similarly about datetimes. What if ten years from now we're all using computers that handle and commonly store picoseconds? Would it not be better for the spec to specify a minimum or remain silent on the matter? This way, each parser can announce its capabilities (which may improve over time) and users can choose a parser that meets their requirement.

ChristianSi:
@mojombo :+1:

RFC 3339 doesn't prescribe a precision, neither should TOML.

groves:
Fair enough. I can just document the restriction on my parser.

PR #274, opened 2014-12-23
Unicode escapes must be scalar values

g2p:
Code points was not precise enough.

Explicitly disallow the JSON UTF-16 surrogate hack which would cause portability issues.

(Something Python 3 should have done, incidentally; as of Python 3.4 you can build UCS-2 strings that can't be printed to a UTF-8 stream)


mojombo:
Good call.

Issue #220, opened 2014-06-24
A Proposal for Keys


wycats:
This is related to #65, #67, #185, #90, #180, #62, #126, #83, and probably others.

# Motivation

The TL;DR is that keys are currently slightly ambiguous, but also don't support a number of commonly desired characters.

A couple of examples:

```toml
[ips.127.0.0.1]
[directories.Space Separated]
```

In Cargo, this comes up because we use dependency names as keys:

```toml
[dependencies]

hammer.rs = "1.0.0"

# or

[dependencies.hammer.rs] = "1.0.0"
```

# Proposal

* Unadorned keys are one or more [alphabetic](http://www.unicode.org/reports/tr44/#Alphabetic), [numeric value](http://www.unicode.org/reports/tr44/#Numeric_Value), `-`, `_`, `$`, `!` or `?` characters.
* Quoted keys are valid [Toml Strings](https://github.com/toml-lang/toml#string).

So this would be valid:

```toml
[dependencies]

"hammer.rs" = "1.0.0"

# or

[dependencies."hammer.rs"] = "1.0.0"
```

I know there have been many discussions on this topic before. I have read them and have tried to include their considerations in this proposal.

One note: I think including `-` as a valid unadorned identifier character is extremely important.

mk-pmb:
So with current TOML, can I have unicode characters 0 to 31 a key name? I'm not familiar enough with TOML yet to guess about `\n`, but `\r\t` and terminal control escape sequences will surely add to my experience of editing config files on a terminal with grep/sed/cat. Especially for terminals that support silent/password input mode.

> Name your tables whatever crap you please, just don't use a dot. Dot is reserved. OBEY.

Probably well intentioned, but I'd still favor a whitelist approach. With a blacklist, many webdevs might remember to care about non-breaking space, some who know that JSON is not a JavaScript subset may even remember about U+2028 (line separator) and U+2029 (paragraph separator), and might hope that Unicode Consortium won't add too fancy new whitespace ever.

> Or, to keep it simple: key parts must be comprised of arbitrary sequences of characters belonging to the Unicode Categories Letter (L.), Number (N.) and Mark (M.), as well as _ and - (underscore, hyphen).

I couldn't find a good enough twin for `.` in just letters and digits yet, so maybe it's safe (for now) to whitelist all letters and digits. Depending on font, U+05C5 (hebrew mark lower dot) might work in some terminals and editors.

mojombo:
[...]

PR #283, opened 2015-01-07
Clarify key name syntax.

mojombo:
After reflecting on the matter of key names for quite a while, and taking into consideration the many opinions in #65, #67, #220, etc, I've changed my mind on how I think they should be specified. This PR proposes two types of keys: bare and quoted. Bare keys are extremely restrictive `[A-Za-z0-9-_]+` while quoted strings can contain almost anything. I like this approach because it is easy to understand (and remember) and guides users to choose simple key names while still making complex names possible. This proposal should also make the parser's job very simple and eliminate any weirdness that could come from having to deal with undelimited Unicode.

cc @wycats, @BurntSushi, @ChristianSi

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,28 +319,46 @@ Keys are on the left of the equals sign and values are on the right. Whitespace
 is ignored around key names and values. The key, equals sign, and value must
 be on the same line (though some values can be broken over multiple lines).
 
-Key names may only consist of non-whitespace, non-newline characters excluding
-`=`, `#`, `.`, `[`, and `]`.
+Keys may be either bare or quoted. **Bare keys** may only contain letters,
+numbers, underscores, and dashes (`A-Za-z0-9_-`). **Quoted keys** follow the
+exact same rules as basic strings and allow you to use a much broader set of key
+names. Best practice is to use bare keys except when absolutely necessary.
 
 Key/value pairs within tables are not guaranteed to be in any specific order.
 
 ```toml
 [table]
 key = "value"
+bare_key = "value"
+bare-key = "value"
+
+"127.0.0.1" = "value"
+"character encoding" = "value"
+"ʎǝʞ" = "value"
 ```
 
-Dots are prohibited in key names because dots are used to signify nested tables!
+Dots are prohibited in bare keys because dots are used to signify nested tables!
 Naming rules for each dot separated part are the same as for keys (see above).
 
 ```toml
-[dog.tater]
+[dog."tater.man"]
 type = "pug"
 ```
 
 In JSON land, that would give you the following structure:
 
 ```json
-{ "dog": { "tater": { "type": "pug" } } }
+{ "dog": { "tater.man": { "type": "pug" } } }
+```
+
+Whitespace around dot-separated parts is ignored, however, best practice is to
+not use any extraneous whitespace.
+
+```toml
+[a.b.c]          # this is best practice
+[ d.e.f ]        # same as [d.e.f]
+[ g .  h  . i ]  # same as [g.h.i]
+[ j . "ʞ" . l ]  # same as [j."ʞ".l]
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -319,28 +319,46 @@ Keys are on the left of the equals sign and values are on the right. Whitespace
 is ignored around key names and values. The key, equals sign, and value must
 be on the same line (though some values can be broken over multiple lines).
 
-Key names may only consist of non-whitespace, non-newline characters excluding
-`=`, `#`, `.`, `[`, and `]`.
+Keys may be either bare or quoted. **Bare keys** may only contain letters,
```
ChristianSi: "letters" should be "ASCII letters" ('Ä' is a letter too).

BurntSushi:
I very much like this!

ChristianSi:
I made a little line note, otherwise this looks good :+1:

redhotvengeance:
:+1:

BurntSushi:
@wycats @alexcrichton do you have any thoughts on this? Will it cause any major churn in `Cargo.toml`? Unquoted keys allow for `-` and `_`, which I know a few packages use, so that's good I think.

(For others: [Cargo](http://doc.crates.io/guide.html) is [Rust's](http://www.rust-lang.org/) package manager, and it uses TOML as its configuration format. If they aren't already the biggest users of TOML, I'm assuming they will be soon. So I think it's good to check with them before we make backwards incompatible changes.)

mk-pmb:
:+1:

alexcrichton:
Bare keys being so restrictive was somewhat surprising to me, but I think we're fine on Cargo's end. I believe we currently have the exact same restrictions on package names (which can be used as a bare key), and all other keys are basically the same.

This sounds like a great idea, and thanks for the heads up!

flaviut:
:+1:, although in the interest of compatibility with existing ini files, I'd also like clarification on how <https://github.com/toml-lang/toml/issues/220#issuecomment-47051538> should work.

mojombo:
@flaviut As much as I'd love for TOML to be somehow compatible with INI files, I don't really see it happening. Since inception, TOML has diverged from INI (if that's even possible, given the non-standardization of INI), and we must make choices that are right for TOML even if that means giving up on the dream of TOML-as-superset-of-INI.

mojombo:
> I believe we currently have the exact same restrictions on package names (which can be used as a bare key), and all other keys are basically the same.

@alexcrichton Sweet. I'd love to verify that, just to make sure TOML will continue to be easy to use for Cargo. Think you can take a look and see, or point me to the proper place in the Cargo source? Thanks!

alexcrichton:
> Think you can take a look and see, or point me to the proper place in the Cargo source?

Sure, this is actually [validated in our registry](https://github.com/rust-lang/crates.io/blob/743bfc980b76c914d545b41d6f4cb601a5c1d213/src/krate.rs#L226-L231), crates.io, when crates are published.

mojombo:
@alexcrichton Excellent, thank you for checking! Do you know if @wycats is around? I'd love to have him weigh in before I merge this.

wycats:
@mojombo I think the biggest issue here is that `ünicode-crate = "1.0"` (which we don't support now, but could) would require quotes. I don't think the quotes are such a big deal, but would they be expected?

What do you think?

mojombo:
@wycats Yes, with this PR you'd need to specify `"ünicode-crate" = "1.0"`. Idealistically I'd love to support unicode in key names, but I think the complexity and ambiguity is just too much to make it worthwhile. Quoting lets a reader know that something unexpected might be going on in the key and force them to look twice, reducing unnecessary confusion and mistakes.

mojombo:
I'm going to merge this PR, as I strongly believe it's in the right direction. If we need to tweak it, we can do so in another PR. Thanks for the feedback, everyone. One step closer to 1.0!

PR #286, opened 2015-01-16
Clarification about newlines in multi-line strings and empty keys in quotes

alexcrichton:
I updated some of the language around multi-line basic strings to be clear that they allow bare newlines (I think this is ok, right?).

I also didn't see specific discussion of an empty key in quotes after #283, so I added a commit as well which uses them as an example of bad key names. I could revert this, however, if it's not desired!

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,7 +415,9 @@ All table names and keys must be non-empty.
 [a..b]
 [.b]
 [.]
+[""]
  = "no key name" # not allowed
+"" = 3           # not allowed
 ```
 
 Array of Tables
--- a/README.md
+++ b/README.md
@@ -164,9 +164,9 @@ key3 = """\
        """
 ```
 
-Any Unicode character may be used except those that must be escaped: backslash
-and the control characters (U+0000 to U+001F). Quotation marks need not be
-escaped unless their presence would create a premature closing delimiter.
+Any Unicode character or newline may be used except those that must be escaped:
+backslash and the control characters (U+0000 to U+001F). Quotation marks need
+not be escaped unless their presence would create a premature closing delimiter.
 
 If you're a frequent specifier of Windows paths or regular expressions, then
 having to escape backslashes quickly becomes tedious and error prone. To help,


ChristianSi:
I don't really get the point of the first commit: newlines are Unicode character just like any other, so of course they are allowed. Also, the "multi-line strings" already implies that they can contain newlines.

As for the second commit, I'd think that *any* string within quotation marks is valid, including the empty string. Certainly not a good idea to use that as part of a key but I don't see a reason why it should be illegal.

mojombo:
@alexcrichton I'm not sure this change is necessary. I agree with @ChristianSi that "Any Unicode character may be used except those that must be escaped" is quite clear about what's allowed.

As for empty quoted keys, I don't think they should be allowed. I can't think of a single good reason a config file would need a blank key, and allowing them could lead to accidentally incorrect parser behavior and confused consumers of config files. The ABNF I'm working on already disallows it:

```abnf
key = unquoted-key / quoted-key
unquoted-key = 1*( ALPHA / DIGIT / %x2D / %x5F ) ; A-Z / a-z / 0-9 / - / _
quoted-key = quotation-mark 1*basic-char quotation-mark ; See Basic Strings
```

I'm planning on including the ABNF in the actual spec soon, and that should be sufficient to clarify this case without requiring additional examples. How's that sound?

alexcrichton:
Hm, it appears my late-night self isn't parsing that sentence as easily as my early-morning self. I agree with @ChristianSi as well about the first change.

Also, an ABNF sounds fantastic, so I'll just close this in favor of that, thanks @mojombo!

mojombo:
@alexcrichton Cool. Feel free to check out the full proposed ABNF in #236.

ChristianSi:
I'm not so sure it was a good idea to close this. If empty quoted keys are forbidden, then the following sentence in the spec is wrong:

> Quoted keys follow the exact same rules as basic strings and allow you to use a much broader set of key names.

Since `""` is certainly a valid basic string, isn't it?

I'd propose to change to sentence to something like:

> Quoted keys follow the same rules as basic strings, except that the empty string (`""`) is forbidden. They allow you to use a much broader set of key names.

Otherwise there would be a contraction between the ABNF and the written spec which would be a bad thing.

Issue #295, opened 2015-02-05
Thousands separators: +1_200_300.45


hamstergene:
It would be good to allow thousand separators in numeric literals:

    x = 2_000_089
    y = 1_999.95

Only single `_` between digits must be allowed. Combinations like `_1`, `2_` and `3___4` should be illegal.

mojombo:
This seems reasonable. Some kind of separator greatly enhances readability of large numbers, and I see that as highly desirable for config files. @BurntSushi thoughts?

BurntSushi:
I personally love this. Rust has it and I use it all the time.

Config files are meant to be read and written by humans and I think this greatly improves readability.

mojombo:
Cool. I'll work up a PR.

mojombo:
Please continue discussion in #296.

PR #296, opened 2015-02-05
Allow underscores in numbers.

mojombo:
Makes these possible. Addresses #295.

```
# integers
1_000
5_349_221
1_2_3_4_5     # valid but inadvisable

# floats
9_224_617.445_991_228_313
1e4_000_000
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,15 @@ Negative numbers are prefixed with a minus sign.
 -17
 ```
 
+For large numbers, you may use underscores to enhance readability. Each
+underscore must be surrounded by at least one digit.
+
+```toml
+1_000
+5_349_221
+1_2_3_4_5     # valid but inadvisable
+```
+
 Leading zeros are not allowed. Hex, octal, and binary forms are not allowed.
 Values such as "infinity" and "not a number" that cannot be expressed as a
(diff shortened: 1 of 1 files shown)


ChristianSi:
:+1:

mojombo:
@BurntSushi Any further comments on this? Otherwise I'll merge it in.

mwanji:
Isn't `1e4_000_000` outside of the range of a float?

mojombo:
@mwanji Yes, that's true! I'll push a fix.

mojombo:
@BurntSushi I'd love to see your +1 on this before I merge it.

BurntSushi:
@mojombo :+1:

fbernier:
Sorry for reviving this old PR but I think something should be clarified:

`Each underscore must be surrounded by at least one digit.`

Does that mean that `45_` should be valid? If not, it should be mentioned that it can't start or end with an underscore.

mojombo:
@fbernier Ah, I see what you're saying. I guess I thought "surrounded" implied a digit on both sides. How about:

`Each underscore must be surrounded by at least one digit on each side.`

fbernier:
Sounds good to me!

Issue #263, opened 2014-11-11
Datetime literals are not appropriate for a minimal language


thezerobit:
If this is to be taken at face value: "TOML aims to be a minimal configuration file format", then embedding all of RFC 3339 to get datetime literals makes no sense. Since the string values of TOML keys are restricted, I don't see TOML as being appropriate as a general data serialization protocol (in which a native datetime literal would be useful), so I assume it really is meant to be for human-edited config files. I've written/edited plenty of configuration files, and I can't remember when I ever needed to encode a datetime by hand.

oliverjanik:
I vote to keep the datetime in. JSON doesn't have a datetime type and it's a sore every day.

cies:
Do you use json for config files when you miss datetime?

steakknife:
It might be simpler to borrow from a standard like ISO 8601 only in the UTC format of `yyyy-MM-ddThh:mm:ss[.sssssss]Z` for datetime.  If a date type were essential, exclude the T part.

cies:
That is not easily a hieved when writing a parser.. Hence my point: dates
are not very minimal in a conf lang.

And when one needs dates in a conf lang, chances are big you need more
(durations, tzones, repetitions), or less (only year+month+day) then
currently in the toml spec.

mojombo:
There are some fair points raised here. When I originally wrote the spec, I included Date because it's a common language type and it didn't seem like it would add too much complexity. But everything added takes away from everything else. I do see TOML as a configuration format (not a serialization format), and I've also rarely needed to specify a date in a config file. They are indeed a bit pesky to parse (though not too bad, really), and tend to produce arguments around them. Honestly, I could be convinced to drop Date support, but I don't think I'm quite there yet.

@BurntSushi what are your thoughts?

cies:
Thanks for having a look at this.  For me the parsing is certainly not the
biggest problem. Its more the utility: they only serve a very narrow
purpose (non tzone, non duration, fixed precision). I cannot imagine i will
ever need exactly this datetime type in a config file; therefor imho not
part of a "minimal" config file format spec.

mojombo:
I asked on Twitter if anyone was using datetimes in config files, and the most useful response I got was this from @Keithamus:

@mojombo They’re useful for fixture files!

[...]

PR #297, opened 2015-02-06
Allow less specific datetimes.

mojombo:
This allows three kinds of datetimes:

```bash
# Full RFC 3339
1979-05-27T07:32:00Z
1979-05-27T00:32:00-07:00
1979-05-27T00:32:00.999999-07:00

# No local offset
1979-05-27T07:32:00
1979-05-27T00:32:00.999999

# No local offset or time
1979-05-27
```

See #263 for why this is being proposed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,13 +261,35 @@ false
 Datetime
 --------
 
-Datetimes are [RFC 3339](http://tools.ietf.org/html/rfc3339) dates.
+There are three ways to express a datetime. The first is simply by using the
+[RFC 3339](http://tools.ietf.org/html/rfc3339) spec.
 
 ```toml
 1979-05-27T07:32:00Z
 1979-05-27T00:32:00-07:00
 1979-05-27T00:32:00.999999-07:00
 ```
+
+You may omit the local offset and let the parser or host application decide that
+information. A good default is to use the host machine's local offset.
+
+```toml
+1979-05-27T07:32:00
+1979-05-27T00:32:00.999999
+```
+
+If you only care about the day, you can omit the local offset and the time,
+letting the parser or host application decide both. Good defaults are to use the
+host machine's local offset and 00:00:00.
+
+```toml
+1979-05-27
+```
+
+The precision of fractional seconds is implementation specific and no minimum is
+expected. If the implementation's precision is exceeded, the datetime should
+still be accepted without error (though a warning should be issued, if
+possible).
 
 Array
 -----


cies:
> The precision of fractional seconds is implementation specific and no minimum is
> expected. If the implementation's precision is exceeded, the datetime should
> still be accepted without error (though a warning should be issued, if
> possible).

So the following is valid?

1979-05-27T00:32:00.

How about dates before epoch? Or before the year zero?

This also introduces the concept of a parse warning... Before it would be valid or invalid, after this it may become valid+warning. Everyone app that uses the TOML format should now decide if/how it wants to deal with warnings (tell the user? log it somewhere?).

Please consider that date/time/timezones/timestamps is a can of worms that we do not want to open upon a *minimal* config format.

Quick proof: http://en.wikipedia.org/wiki/Leap_second

More proof in [this blog post](http://infiniteundo.com/post/25326999628/falsehoods-programmers-believe-about-time) and [sequel](http://infiniteundo.com/post/25509354022/more-falsehoods-programmers-believe-about-time), most of the falsehoods mentioned are not worth considering in the context of a serialization format, some of them certainly are.

ChristianSi:
Looks good to me :+1:

@cies As for your questions, it's best to consult RFC 3339. For examples, leap seconds are addressed by the RFC (seconds in the grammar range from 0 to 60 because of that).

Warnings are optional ("if possible") so I don't see that as a big issue.

mojombo:
@spf13 Can you weigh in on this proposal? Would it solve your needs?
@parkr I'd love to hear from you as well, from a Jekyll perspective.

spf13:
I think this would work quite well. One of the odd behaviors today is the very strict format. I think this would be a real boost to adoption. I can see the no offset being commonly used. It would definitely work well for Hugo's current use case.

mojombo:
I'm merging this right now, as I think it will make TOML more pragmatically useful, but I'd enjoy a discussion about whether the bit about warnings should be changed to just be an error.

spf13:
Thanks @mojombo . I'm happy to pass along any feedback I hear about it.

ChristianSi:
Looks good :+1:

I'm in favor of leaving the warning as is, it feels like an awful waste lo let a parser die just because it has to crop 12.345678 secs to 12.345 (or 12.346) secs.

Warnings might also be appropriate in other situations, e.g. when a key occurs twice in a table. The spec says "Doing so is invalid" -- leaving it up to the implementor whether to throw a parsing error or emit a warning, I guess.

ChristianSi:
On second thought, maybe change

> The precision of fractional seconds is implementation specific and no minimum is expected.

to

> The precision of fractional seconds is implementation specific, but at least millisecond precision is expected.

?

Rounding 12.345678 secs to 12 secs *would* feel a bit rogue to me.

If millisecond precision is the common minimum, then I think it would be OK to make the warning optional ("may" instead of "should").

HuwCampbell:
It's quite dangerous to treat raw days as is suggested. For example in Brazil on the 18th of October, there is no 00:00:00 time at all, as with daylight savings it skips straight to 1 am (this gem has caused, for example, popular issue trackers to not allow issues to be defined on this day).

One can have a raw days implementation, but it is very different to a day's midnight in a timezone (it's a completely different datatype). ISO 8601 also allows this (but is not what you're implementing anymore).

This change makes TOML not precise, which is bad:tm:.

HuwCampbell:
I think warnings opens up a bit of a can of worms too; could one, for instance, turn on -Wall and fail?

mojombo:
@HuwCampbell That's fair, but I still think this change is needed. How about suggesting noon as the time-point for raw days? Every day, regardless of daylight savings or other leap-changes has a noon.

mojombo:
>The precision of fractional seconds is implementation specific, but at least millisecond precision is expected.

@ChristianSi I like this idea, it's consistent with the float spec (some minimum precision is expected), and just leave out the warning suggestion. Any implementation is free to warn on any number of cases that may cause issues for it. I'll work up a PR.

mojombo:
@ChristianSi See #360 for the change regarding milliseconds.

mojombo:
@HuwCampbell See #361 for the noon-time change.

PR #328, opened 2015-05-01
This README needs an index yo!

santiagobasulto:
I think that having an index might be useful, it certainly is for me.

area de trabajo 1_105

This is the code I used to generate it, idk if it's worth to commit it somewhere:

```javascript
$('article.markdown-body').find('h2').map(function(i, v){
  return ("[" + $(v).text().trim() + "]" + "(#" + $(v).find('a').attr('id') + ")");
}).toArray().join("\n- ")
```


santiagobasulto:
@mojombo, @BurntSushi can you guys check this?

BurntSushi:
@santiagobasulto Thanks for the reminder! I like the ToC. :-)

PR #332, opened 2015-06-16
Clarify on the use of bare integers as keys.

avakar:
I totally missed on the fact that plain integers can be used as bare keys when writing my parser. This actually might make two-phase parsing (i.e. lexing then parsing) an unfavorable approach since there is an ambiguity on whether `1234` is a key or a number. It's therefore good to notice in advance, hence this clarification.


BurntSushi:
Thanks!

PR #333, opened 2015-06-21
Subsume comments in newline

avakar:
I've introduced a new symbol for "extended whitespace" in which newlines would not be significant and rearranged `newline` to support it. This patch fixes the whitespace problem in inline arrays which I wrote to you about in https://github.com/toml-lang/toml/pull/236#discussion-diff-32894591.

### DIFF
--- a/toml.abnf
+++ b/toml.abnf
@@ -3,21 +3,17 @@
 
 ;; TOML
 
-toml = expression *( newline expression )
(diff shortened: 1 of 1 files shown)


mojombo:
Good catch. This is resolved in a slightly different manner in the latest ABNF. Cheers!

Issue #330, opened 2015-06-12
How soon is this going to go stable?


isiahmeadows:
It is used in production in many areas, such as Rust's package manager Cargo, but the spec itself is labelled as unstable. This is a concerning situation, for obvious reasons.

I'm not saying it's a bad config format (it's not). I'm just curious about how close this is to going stable (v1.0) and what's holding it back.

/cc @BurntSushi

jakwings:
The official validator hasn't been updated for a long time. And I've observe some difference between some JavaScript parsers… [Here is my test suit.](https://github.com/jakwings/toml-j0.4/tree/master/test)

I think we should get a (somewhat) standard plain-text AST representation for validating.

BurntSushi:
I think that except for #263, most outstanding issues are related to adding things to TOML. I'm hoping we can convince @mojombo that the path to 1.0 does not include any breaking changes at this point.

I think we really wanted to get #236 merged before declaring 1.0, but other than that, I don't see any major reasons to wait much longer.

pnathan:
I would like to see #236 wrapped up and the EBNF formally checked as the final milestone for 1.0. It's been a bit of a long road, and I'm ready to see it over. :)

njsmith:
I wanted to ping this issue, because there's a discussion going on in python-land right now about what config file format to use for source packages (similar to the Cargo.toml use case), and the argument so far has gone like "toml is obviously the technically superior option", "yes but look at that terrifying line about stability at the top of the spec, we can't possibly use this, so how about yaml".

BurntSushi:
@mojombo How do you feel about getting #236 and #297 merged, and then releasing a 1.0?

impinball:
@njsmith 

> [...] "toml is obviously the technically superior option", "yes but look at that terrifying line about stability at the top of the spec, we can't possibly use this, so how about yaml"

This is *exactly* why I filed the bug. It's very nice for config and data, but it needs stability quick.

mojombo:
@BurntSushi OK LET'S DO THIS.

alexeyr:
Please consider including #362 as well.

nexocentric:
Most definitely waiting for it to go stable. Until then... making do with yaml.

rsyring:
FWIW, here is another thread where the Python team is leaning towards TOML, but there are concerns about stability.   https://mail.python.org/pipermail/distutils-sig/2016-May/028786.html

[...]

PR #334, opened 2015-06-22
[RFC] TOML v0.4.x grammar in ABNF

jakwings:
This is a specification for the supposed TOML v0.4.x in [ABNF](https://en.wikipedia.org/wiki/Augmented_Backus%E2%80%93Naur_Form).

Basically it is based on [v0.4.0](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v0.4.0.md), or it is just what I think v0.4.0 should be. Nothing new is added to the grammar. But this is the best I can do to clarify my idea. And I think these grammar rules are quite concise and not hard to understand.

I hope this can help to establish the new TOML v1.0.0 in the future. Welcome to criticize the grammar here!

Ping: #236 #330

### DIFF
--- a/versions/en/toml-v0.4.x.abnf
+++ b/versions/en/toml-v0.4.x.abnf
@@ -1,0 +1,321 @@
+;; The specification of TOML v0.4.x format in ABNF.
+;; The ABNF grammar is defined in RFC 5234 (http://www.ietf.org/rfc/rfc5234.txt)
+
+
+TOML = Expressions;
+
+Expressions
+  = *( Whitespace / Newline / Comment )
+    [ Expression *( Whitespace / Comment )
+      [ Newline Expressions ]  ; Multiple expressions stay in different lines.
+    ]
+
+Expression
+  = KeyValue
+  / TableHeader
+  / TableArrayHeader
+
+Newline
+  = %x0A     ; LF
+  / %x0D.0A  ; CRLF
+
+Whitespace
+  = %x09  ; Horizontal Tab "	"
+  / %x20  ; Space " "
+
+Comment
+  = CommentStarter *CommentCharacter
+
+CommentStarter
+  = %x23  ; Number sign #
+
+; Control-characters except horizontal tab, are not allowed.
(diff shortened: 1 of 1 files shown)


karupanerura:
:+1:

Profpatsch:
There is another, more recent try: https://github.com/toml-lang/toml/pull/236

jakwings:
@Profpatsch I know, and the grammar file and discussions there are a little messy, so I write my own with more details.

jakwings:
Closed because I think TOML is not smart enough. There is still time to make a good new format.

Profpatsch:
@jakwings Interesting, what brings you to that conclusion?

jakwings:
First, the most important reason: TOML is not that popular and easy to parse as I think and there are still many variants of ini format (for ease of use, performance rather than consistency). Everyone can learn the basic of ini-like format in a few minutes. If TOML can't be the best default, and it is really easy to write a parser, why should I stick to it?

Biggest flaws in TOML: DateTime; not easy to parse `[[section array]]`. (I won't discuss here.)

Profpatsch:
I see.

versions/en/toml-v0.4.x.abnf
```diff
@@ -0,0 +1,321 @@
...
+  = %x5C  ; \
+
+Comma
+  = %x2C  ; ,
+
+Underscore
+  = %x5F  ; _
+
+Letter
+  = %x41-5A  ; "A" to "Z"
```
TerjeBr: Why not allow all unicode letter characters?

http://www.fileformat.info/info/unicode/category/index.htm

Just include all unicode classes with the name Letter, <something>

jakwings: Well, you'd better ask the original designers of TOML 0.4.0, because I just wanted to get more details of 0.4.0 depicted.
PS. The chart of Unicode characters is just a mess. Be careful. Also, my grammar still have bugs, but I've no interest to improve it any more.

PR #341, opened 2015-07-17
Clarify that comments are OK in multiline arrays

### LINKED ISSUE #340: Are comments OK in multiline arrays?
The spec says:

> Arrays can also be multiline. So in addition to ignoring whitespace, arrays also ignore newlines between the brackets. 

With an example:

```
arr8 = [
  1,
  2, # this is ok
]
```

The description doesn't appear to allow comments in multiline arrays (ignoring whitespace and newlines doesn't imply ignoring comments). I'm not sure whether the example is supposed to be valid as written, or whether the comment is for illustration and should be removed.

The description of comments being "the rest of the line" implies that this is valid, but that could also be read to allow comments in multiline strings, which is (I hope) not the case.

takluyver:
Fixes #340

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,8 +326,8 @@ arr6 = [ 1, 2.0 ] # note: this is NOT ok
 ```
 
 Arrays can also be multiline. So in addition to ignoring whitespace, arrays also
-ignore newlines between the brackets. Terminating commas are ok before the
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks!

PR #360, opened 2015-10-20
Add millisecond precision expectation to datetime.

mojombo:
It makes sense for all implementations to support millisecond precision for datetimes. This also removes the suggestion of a warning, as that opens up a whole new can of worms. Every implementation may provide warnings on anything it wishes without the spec weighing in on the matter.


ChristianSi:
Looks good :+1:

PR #361, opened 2015-10-20
Suggest noon for raw day datetimes.

mojombo:
Suggesting 00:00:00 for datetimes that have only their day specified might cause problems due to daylight savings time and leap changes. Suggesting noon should avoid any of those problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,7 +320,7 @@ information. A good default is to use the host machine's local offset.
 
 If you only care about the day, you can omit the local offset and the time,
 letting the parser or host application decide both. Good defaults are to use the
-host machine's local offset and 00:00:00.
+host machine's local offset and 12:00:00.
 
 ```toml
 1979-05-27


mwanji:
It might be better to suggest using a date library that handles these sorts of things properly.

For example, in Java 8 there is [LocalDate#atStartOfDay](http://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html#atStartOfDay-java.time.ZoneId-) which gets the first existing time of the given date.

ChristianSi:
Looks good -- though #362 is not unreasonable either.

HuwCampbell:
So better (in that it won't crash), but still very problematic.
As an example: company in Sydney has an offer with an expiry date. They encode this in TOML. Customer in Sydney finds they can't use their coupon past midday, even worse, customer in Perth can't use theirs past 9am.

mojombo:
No longer relevant due to #414.

PR #362, opened 2015-10-21
Treat Dates and Datetimes as separate types

skystrife:
I think the current direction of handling DateTimes with no time (PR #297) is going to lead to unnecessary confusion. While it's true that every day ought to have a "noon" (as suggested for PR #361), I don't think that it's going to be intuitive for users that the following TOML is parsed as being today at noon:

``` toml
backtothefuture = 2015-10-21
```

To me, this simply says that key `backtothefuture` represents the date "Wednesday October 21st, 2015".

I think, if we want to support time-less dates, let's do exactly that and be explicit about it. We already have specific types for these things in many programming languages (e.g., ruby and python both have `Date` as well as `DateTime`), and if we don't specify a time, in the config, let's just not model it. This gets rid of the problem of trying to have a "default" time for time-less DateTimes, and I think is intuitive and not complicated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,10 +299,20 @@ bool1 = true
 bool2 = false
 ```
 
+Date
+----
+
+Dates can be specified in `YYYY-MM-DD` format. A date does not model time;
+for this, use a datetime (below).
+
+```toml
+1979-05-27
+```
+
 Datetime
 --------
 
-There are three ways to express a datetime. The first is simply by using the
+There are two ways to express a datetime. The first is simply by using the
 [RFC 3339](http://tools.ietf.org/html/rfc3339) spec.
 
 ```toml
@@ -317,14 +327,6 @@ information. A good default is to use the host machine's local offset.
 ```toml
 1979-05-27T07:32:00
 1979-05-27T00:32:00.999999
-```
-
-If you only care about the day, you can omit the local offset and the time,
-letting the parser or host application decide both. Good defaults are to use the
-host machine's local offset and 00:00:00.
-
-```toml
-1979-05-27
 ```
 
 The precision of fractional seconds is implementation specific, but at least


skynet:
:+1:

ChristianSi:
Strictly speaking, timezone-less datetimes could (and possibly should) be distinguished from those with timezone information, too. The latter can alternatively be represented as seconds since the epoch, without ambiguity, while that's not true for the former. Also, depending on the timezone one is in, some timezone-less datetimes will not even represent an existing point in time, or they will ambiguously represent two points in time -- both because of DST.

So a case could be made to represent these as three different types, as SQL does: *timestamp without time zone, timestamp with time zone, date.*

My inclination is that it might be better to let implementations decide how exactly to deal with dates+timestamps serialized in a TOML document. They could convert everything into a timestamp (just one type); or they could distinguish dates from timestamps, but supply timezone information for all of the latter (two types, as you suggest); or they could treat all three as distinct (like SQL). My understanding of "let the parser *or host application* decide" in the TOML spec (emphasis added) is that's it's up to the parser whether or not it converts everything into a timestamp. If it doesn't, it's up to the application to deal with that.

I think that's pretty reasonable and it may be best to leave it that way.

skystrife:
I'm fine with the *host application* doing whatever they want to with the result given to them by the parser. What I'm not so sure about is the *parser* deciding to prepopulate values automatically in the absence of information.

If it does so, as an application writer I can't distinguish a "tomorrow at noon" date that's the result of someone writing `2015-10-23` or someone writing `2015-10-23T12:00:00`, meaning that the host application is no longer empowered to make whatever decision it wanted to about what the timestamp of a value `2015-10-23` should be since the parser already chose for them.

In general, I'm concerned about having parsers making "default" choices for people. I'm inclined to agree with you on the notion of having a Datetime-with-zone and Datetime as separate types for that reason, but I felt that might complicate things a bit too much. I can add that distinction if more people think doing so is a good idea.

I guess my point is that if I'm using a configuration format that I expect to be unambiguous, having parsers in different languages choosing to do different things with the exact same input is going to be frustrating to have to work around (and this will inevitably happen if we just have "suggestions" for defaults). I think it's probably better for the spec to be very clear about what parsers *must* do, rather than what they *can* do to avoid interoperability issues. If we can force the application code to decide what to do and prevent ambiguity at the parsing step, I think that's a net win for the simplicity of the format, even if it means we have one (or two) more types.

HuwCampbell:
Copy from #361, but whatever:
`Company in Sydney has an offer with an expiry date. They encode this in TOML. Customer in Sydney finds they can't use their coupon past midday, even worse, customer in Perth can't use theirs past 9am.`

Dates and times are not instants. TOML does instants precisely currently (0.4), which is great. If TOML wants to do `local` times, or dates, it should be explicit in doing so.

ChristianSi:
Reconsidering this, I tend to agree with @skystrife and the other commentators here: Dates are conceptually different from timestamps (datetimes) and hence should be considered as a different data type. Albert Einstein was born on 1879-03-14 (not: 1879-03-14T12:00:00+some timezone), last Western-style Easter Sunday was 2015-04-05 (not: 2015-04-05T12:00:00+some timezone).

Dates (days) last (typically) 24 hours and don't have a timezone. (My birthday date doesn't change when I move into a different timezone, and Easter Sunday is on the same date in Europe and the US, even though it starts a few hours earlier in Europe.)

Converting a date into a timestamp or vice versa will almost certainly be the wrong thing to do, and so TOML should not encourage that practice but treat both as distinct.

Having said that much, I would additionally suggest to undo the part of #297 that introduced timezone-less datetimes. Being able to write "1979-05-27T07:32:00" instead of "1979-05-27T07:32:00Z" or even "1979-05-27T00:32:00-07:00" is a very small advantage compared to what you lose:

* You know longer know which timestamp (seconds since epoch) is actually intended.
* A TOML document containing timezone-less datetimes maybe invalid in some timezones (where that datetime doesn't exist at all or is ambiguous because of a DST switch), but perfectly valid in others. But the validity of a document shouldn't depend on some settings of the computer on which it is opened!

RFC 3339 says:

> Since interpretation of an unqualified local time zone will fail in approximately 23/24 of the globe, the interoperability problems of unqualified local time are deemed unacceptable for the Internet.

Those are wise words, and I would therefore suggest to merge this PR and prepare another one that removes the recently-added support for timezone-less datetimes.

ChristianSi:
@mojombo @BurntSushi What do you think?

BurntSushi:
Datetimes were on the brink of *being removed* from TOML because their use cases weren't very compelling. We had a few folks chime in and say that, yes, indeed, on occasion they are useful to have.

I don't really see any reason to add an *additional* type at this moment in time. If an application needs a date, then it should interpret the time as a date.

alexeyr:
@BurntSushi What about at least _allowing_ parsers to distinguish them? It seems that currently parsers are required to treat `2015-04-05` as a date with some (unspecified, pending #361) time.

BurntSushi:
@alexeyr Really? The spec says this:

> If you only care about the day, you can omit the local offset and the time, letting the parser or host application decide both. Good defaults are to use the host machine's local offset and 00:00:00.

It seems like there's plenty of flexibility in those words.

alexeyr:
@BurntSushi 

> letting the parser or host application decide both

reads to me as if they do have to decide both. Maybe add "if desired", "if necessary" or "if they care"?

skystrife:
So what of my concerns about there being ambiguity in how different parsers will handle a date without a time and offset specified? I *really* do not think that it's a good idea for there to be "flexibility" in what the parser returns to me when I have a value of "2015-11-06". That should mean one thing, and one thing only.

One of the things that makes TOML nice is that it is a simple and unambiguous mappping to a hash table across programming languages. Adding dates without times and timezones added ambiguity to the language. This change would remove that ambiguity and still be quite simple.

I don't understand the argument against removing ambiguity by adding a type that is this simple. Do you need more use cases for having a date without a time? Jekyll, Hugo, birthdates, etc. aren't convincing? Is it a mistake that Ruby and Python both have different types for expressing these two concepts?

I fundamentally have a problem with "it's up to the programmer to figure out whether that DateTime was actually a DateTime or just a Date" in application code. I fundamentally have a problem with the spec allowing parsers to make different decisions for how to fill in values when parsing. Adding a simple 3-field type addresses both of these.

skystrife:
> If an application needs a date, then it should interpret the time as a date.

How would you feel about a change that removes the ability to specify dates without times, then? This removes the ambiguity concern and ensures that all parsers do the same thing like before.

BurntSushi:
> I fundamentally have a problem with "it's up to the programmer to figure out whether that DateTime was actually a DateTime or just a Date" in application code.

Why do you need to do this? If the application needs a date, then use it as a date. If the application needs a datetime, then use it as a datetime. I'm not sure I see the problem here. Why do you need to know whether the user entered a date without a time or not?

skystrife:
How about a cron-like app? "Run this thing sometime today" vs "I want you to run this thing at exactly this timestamp". I can easily imagine doing something different for these two cases (first one gives a lot more flexibility).

How about a blog? "This was posted on this day" vs "this was posted at exactly this timestamp". This changes how I would display the posted-at time.

There was a nice coupon example above from @HuwCampbell. "This coupon expires on this day" vs "this coupon is no longer valid after this exact timestamp". It is quite important to tell the difference between a Date and a DateTime here to get the coupon status right across zones.

This example is less important, but if we think about TOML encoders and decoders, I would get a different value written out for a key after decoding a file and re-encoding it if it contains a key that was just a Date and the parser filled in some default values for me when decoding. That's a bit weird.

BurntSushi:
@skynet Thanks for those. That's useful.

@mojombo What do you think? It seems like we either need a new type or remove datetimes without times altogether. It looks like the current state is a bit too complex IMO.

alexeyr:
I think the same choice (possibly with a different result) should also apply to datetimes without offset, especially since there are two different reasonable "default timezone" choices (local and UTC).

BurntSushi:
@alexeyr That would mean having *four* types? No. The parser can make it own choice and specify it. Or even better, most datetime types in languages I'm familiar with will tell you whether they're local or not.

alexeyr:
@BurntSushi That's why I said "possibly with a different result": I think the argument for removing them is much stronger than for removing dates without times.

alexeyr:
@BurntSushi I wasn't suggesting that local and UTC should be different types at all; I meant that parsers being able to choose between them is undesirable.

BurntSushi:
@alexeyr I don't see the problem. The parser should create a datetime object in its language, which should store any timezone information if it exists. What is the problem you're trying to solve by enforcing that all parsers do the same?

alexeyr:
@BurntSushi 

1. Under Objectives we have 

 > TOML aims to be a minimal configuration file format that's easy to read due to obvious semantics. TOML is designed to map unambiguously to a hash table.

 Allowing parsers to choose how to map introduces ambiguity; using noon as default time is unambiguous but non-obvious.

2. In particular, consider the problem of switching between parsers (e.g. to port an application to a different platform, or because a new library for your current platform has significantly better performance) with as little pain as possible.

alexeyr:
For the local time interpretation we can get ambiguity even with a single parser: parsing the same TOML file on different computers or at different times on the same computer can give different results because timezone information is different. It could be non-trivial to detect such a problem if it happened in production.

ChristianSi:
[Above](https://github.com/toml-lang/toml/pull/362#issuecomment-150546775) I proposed doing two things:

* Add a separate date type as per this PR.
* Remove datetimes (timestamps) without timezone.

That move remove all ambiguity about how TOML documents are interpreted by different parsers. I still think it would be the right thing to do.

HuwCampbell:
I don't think there's any good solution apart from ISO 8601 extended format.

It does:
  dates
  local times (lunch is at 1pm)
  times in a zone (the queen eats lunch at 1pm GMT)
  local date times (Halo comes out 9 am November 10, 2015)
  universal date times (I was born on some day at some time specifically)

skystrife:
I'm inclined to side with @ChristianSi here for removing ambiguity from the spec wherever possible, which would entail:

1. Having Date as a separate type (this PR)
2. Removing support for zone-less Datetimes (or explicitly specifying a default like UTC)

Though I care significantly less about the second.

skystrife:
@BurntSushi, @mojombo: Are there any more thoughts on this PR from a maintainer?

jodastephen:
Java SE 8 added a new [date-time package](http://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html) (I was the co-spec lead). The types included in `java.time` represent a very carefully chosen set with clear definitions. I would argue that the TOML spec includes three distinct concepts:

* `LocalDate` - a "local" date without time-zone/offset - 2016-06-30
* `LocalDateTime` - a "local" date and time without time-zone/offset - 2016-06-30T14:30
* `OffsetDateTime` - a date, time and offset - 2016-06-30T14:30+01:00

These three concepts should be considered to be related, but distinct. If a language has representations of the concepts (like Java) then a parser should use them. 

I would argue that the spec should also include the key missing type:

* `LocalTime` - a "local" time without time-zone/offset - 14:30

for which CRON job times are a good example. With those four types, most use cases would be covered.

I'd note that the current spec suggests using 00:00 when the time is missing. It should be noted that this can cause problems on daylight saving cutover in some locales, where local time skips from 00:00 to 01:00, meaning that 00:00 does not happen. Looks like #361 discusses this, but 12:00 is just as much an arbitrary choice.

IMO, the *types* of a local date, local date-time and offset date-time are simply different. The real problem being discussed here is that many programming languages are yet to have the built in support for those types to the same degree that Java now has. Given that fact, TOML has to make a choice:

1. don't support date-times, which would be sad as they are a very common data format
2. only support instants, because this will be supported in all languages (note that an instant loses the offset after parsing)
3. support the current three types, but define mappings to instants (for languages that only have an instant type)
4. support the current three types, but declare the mappings to language-level types to be the choice of the parser

Option 4 is the most flexible and would be my choice. Trying to do option 3 may seem appealing but will have lots of complexity. Option 2 doesn't get you that much, because dates are more common that instants in input files (instants are more common in output files as timestamps).

Given all of the above, this PR is wrong. It is wrong because it only breaks out "local date" in the spec, and not "local date-time".

isiahmeadows:
@jodastephen A great example of that problem is JavaScript, which conflates all three, in a way that working with offset date-times are pretty inconvenient at times (and why Moment.js is so popular).

rsyring:
Coming from Python and having spent time in Date/Time land for a talk I gave a couple years ago, I have to say that I think the best approach is the one with that adds types for date, date/time w/ timezone, and date/time w/out timezone.  That way, there really is no ambiguity from the perspective of TOML and then the languages can map those types to whatever makes sense in the language.

IMO, even though this adds complexity to the spec, it will remove a lot of ambiguity down the road for those using TOML.

mojombo:
In #414 I've proposed some changes that would incorporate many of the well-argued ideas here (special thanks to @jodastephen for the Java reference which I obviously take inspiration from). Please take a look and weigh in.

PR #367, opened 2015-11-06
Allow arrays of arrays of different types even more explicitly.

alexeyr:
There already was an example showing it's legal, so it isn't strictly necessary, but there was one for different kinds of strings as well. Also changes "string types should be considered a single type" to avoid overloading "type".


BurntSushi:
Looks good! Thanks!

PR #368, opened 2015-11-12
Uiri/toml moved to 0.4.0

DavidJFelix:
This project is actively tracking 0.4.x; I haven't tested to ensure that it is actually passing all 0.4.0 tests, but it does claim to be tracking 0.4.0, so the readme might need a check on this one.

Paging @uiri to chime in on this.


uiri:
It is more lenient on datetime parsing (per #297) but should be otherwise conformant to the v0.4.0 spec.

PR #378, opened 2015-12-31
Added grouping to ambiguous alternatives, allow leading newlines inside arrays and spaces after newlines before array elements and newlines and spaces before comments in arrays

joelself:
RFC 4234 Section 3.5 advises the use of grouping notation rather than "bare" alternation when alternatives consist of multiple rules or literals, e.g. "( int float ) / ( bool char )" instead of "int float / bool char".

I might've gotten the grouping wrong, which just serves to illustrate the importance of using grouping notation.

### DIFF
--- a/toml.abnf
+++ b/toml.abnf
@@ -161,7 +161,7 @@ date-time      = full-date "T" full-time
 ws-newline       = *(ws / newline)
 
 array-open  = %x5B ws-newline  ; [
-array-close = ws-newline %x5D  ; ]
+array-close = ws %x5D  ; ]
 
 array = array-open array-values array-close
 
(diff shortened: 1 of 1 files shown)


joelself:
I guess this automatically my new commits to this pull request.

The documentation states that newlines are ignored in arrays and gives examples of arrays with newlines before the first element, however the ABNF doesn't allow for newlines before the first element. I've added a production for multiple whitespaces or newlines, "ws-newline", and changed the "ws" after the array opening brace to "ws-newline".

joelself:
Added another new production to allow whitespace before an array element after a newline. The examples in the README do this, but it isn't allowed by the grammar.

joelself:
Going over your hard example I found I couldn't parse this:
```toml
        multi_line_array = [
            "]",
            # ] Oh yes I did
            ]
```
It turns out that the grammar doesn't allow newlines and whitespaces after the array-sep. I've modified the production again to accommodate newlines and whitespace after the array separator.

PR #380, opened 2016-01-03
Add a hard example test that extensively uses unicode.

joelself:
TOML is supposed to only contain Unicode characters, however all examples and test files exclusively use ASCII chars. This is a version of the existing hard example with all Unicodeable strings (except the comment explaining the rest of the file is incorrect) pseudolocalized with Unicode characters.

A correct TOML parsers has to be able to parse Unicode strings (where allowed) as well as ASCII strings so I believe this test is a valuable addition to verifying if a new parser can correctly parse TOML files.


ChristianSi:
:+1: 

Neat and funny too.

mojombo:
Love it, thanks!

PR #385, opened 2016-01-23
Clarify that literal strings can be table keys

### LINKED ISSUE #354: Allow single-quoted strings as keys
The following constructs don't work:

```toml
[foo.'bar']
# ...
```

```toml
[foo]
'bar' = 42
```

But this does:

```toml
[foo]
bar = 'baz'
```

Since key names look a lot like strings, I think it'd be more consistent if they accepted the full string syntax.

alexcrichton:
This clarifies the definition of "quoted keys" in the section about tables to
indicate that keys can either be basic strings (surrounded by double quotes) or
literal strings (surrounded by single quotes).

Closes #354


mojombo:
Awesome, thanks!

PR #398, opened 2016-03-17
Syntax Highlighting Error on README

markcf:
Hello,

I was reading your readme today and saw a problem with the syntax highlighting. I'm not sure how to fix it well. Is there a way to change how github syntax highlights? I'm not sure. If there is, that'll be much better. 

On the other hand, I figured if I was going to point it out I might include a quick patch to fix it if nobody has a better solution. 

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ exact same rules as either basic strings or literal strings and allow you to use
 a much broader set of key names. Best practice is to use bare keys except when
 absolutely necessary.
 
-```toml
+```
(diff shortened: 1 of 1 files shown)


Alhadis:
GitHub's using [this language package](https://github.com/textmate/toml.tmbundle) to control TOML's syntax highlighting. You can submit a PR or an issue there. :-)

markcf:
That'll be a much better solution, thanks. I've posted an issue there.

PR #438, opened 2016-11-02
Clarify UTF-8 BOM as allowed.

### LINKED ISSUE #437: UTF-8 byte order mark
Let's add a rule whether a BOM at start of file is allowed, required (<-- probably not) or forbidden. I'd prefer allowed (optional in files, required to accept for parsers).

mk-pmb:
Fixes #437.


mojombo:
Fixed by 35fd9216373c866096dbb5ace3f0d4f035b16820.

Issue #436, opened 2016-11-01
Backslashes at end of line


mk-pmb:
TOML faithfully replicates the decades-old multiline string trap:
(Edit: Nope, seems to be a bug in the toml npm package, and it shouldn't have been on the "v0.4.0 compliant" list.)

```text
nodejs -p 'require("toml").parse(require("read-all-stdin-sync")());' <<<'
  msg1 = """hello \
    world"""
  msg2 = """hello \ 
    world"""
  '
{ msg1: 'hello world', msg2: 'hello \\ \n  world' }
```

That really doesn't fit the "O" in TOML.
Could we drop this tradition and boldly adopt a fault-tolerant rule that enables human readers to _visually_ determine what's in a string?

BurntSushi:
Use `'''` instead of `"""`? See the [section on strings](https://github.com/toml-lang/toml#string) for more detail.

> Could we drop this tradition and boldly adopt a fault-tolerant rule that enables human readers to visually determine what's in a string?

I don't understand your objection. Could you say more? Could you also address the downsides of your suggestion? What are the trade offs?

mk-pmb:
With triple apostrophe, I get even more spaces, backslashes and newlines into the text of my paragraph:

```text
noClutter = """Iis igitur est difficilius \
  satis facere, qui se Latina scripta \
  dicunt contemnere."""

hasClutter = """Iis igitur est difficilius \	
  satis facere, qui se Latina scripta \ 
  dicunt contemnere."""

evenMoreClutter1 = '''Iis igitur est difficilius \
  satis facere, qui se Latina scripta \
  dicunt contemnere.'''

evenMoreClutter2 = '''Iis igitur est difficilius \	
  satis facere, qui se Latina scripta \ 
  dicunt contemnere.'''
```

If your browser displays this similar to mine, it's not "obvious" (the O in TOML) where in `hasClutter` the clutter occurs, or without the name, that there was clutter at all.

One way to solve this would be to merge lines if the backslash is the last printing character in a line, trimming not just the whitespace at start of next line but also in the line with the backslash.

The change will only affect files that had whitespace at the end of lines, which I consider a bad style anyway. In cases where that whitespace was accidential, my first-idea method will silently fix the accident. In cases where people rely on whitespace at EOL, I can only recommend to use a less fragile style, e.g. add a custom padding character and strip that at runtime. It may take some more CPU cycles but the files won't break as easily when someone('s editor) "cleans" them up.

[...]

PR #441, opened 2017-01-03
Allow accidental whitespace between backslash and newline.

mojombo:
This change allows (presumably accidental) whitespace to exist in multi-line basic strings between a line-ending backslash and the following newline. See #436 for more context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,13 +207,11 @@ str2 = "Roses are red\nViolets are blue"
 str3 = "Roses are red\r\nViolets are blue"
 ```
 
-For writing long strings without introducing extraneous whitespace, end a line
-with a `\`.  The `\` will be trimmed along with all whitespace (including
-newlines) up to the next non-whitespace character or closing delimiter.
-Accidental whitespace between the backslash and newline is allowed (even though
-it will create an otherwise invalid backslash-space or backslash-tab) and will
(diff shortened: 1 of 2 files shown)


mojombo:
Fixed by 35fd9216373c866096dbb5ace3f0d4f035b16820.

mojombo:
@BurntSushi @ChristianSI Any feedback on this one?

BurntSushi:
LGTM.

ChristianSi:
I'm a bit troubled by the comment "(even though it will create an otherwise invalid backslash-space or backslash-tab)" for that reason that, if there is no accidental whitespace, the backslash followed by a linebreak it not a valid backslash escape sequence either. Hence I would prefer to delete that comment, as I found it needlessly confusing.

Looks good otherwise.

ChristianSi:
Also, I'd propose deleting the word "Accidental" since how is the parser to distinguish between "accidental whitespace" and "non-accidental whitespace"? (The current proposal, when read literally, prohibits the latter.)

There is presumable little reason to add whitespace in that position on purpose, but whether it was added on purpose or by accident is not for the spec writers to judge.

mojombo:
@ChristianSi Ok, try that wording on for size, I think it's much better.

ChristianSi:
@mojombo Excellent!

PR #442, opened 2017-01-03
Fixed datetime links in table of contents

just95:
Added links to new date and time types and removed the old entry.


mojombo:
Awesome, good catch, thanks!

PR #443, opened 2017-01-14
Improve literal strings

ilabdsf:
This change makes it possible to include literal string markers into literal strings by duplicating them. The change is not backwards compatible because I had to forbid multi-line literal strings that span single line. But who thinks single-line multi-line literal strings are obvious anyway? They can be replaced with real multi-line strings simply by inserting a newline at the beginning.

I have thought about escaping ''' line by repeating it too, but for now it is impossible, because these lines may have different newlines and it is not obvious what to do in this case. Otherwise, this change is desirable, as it would allow, for example, to embed any TOML document in a string simply by repeating any line that contains only '''.

Please read third commit message for more details.

ABNF is not updated yet, will do it if feedback is positive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,13 +250,18 @@ regex    = '<\i\c*\s*>'
 regex2   = 'I [dw]on't need \d{2} apples'
 ```
 
-Since there is no escaping, there is no way to write a single quote inside a
-literal string enclosed by single quotes. Luckily, TOML supports a multi-line
-version of literal strings that solves this problem. **Multi-line literal
-strings** are surrounded by three single quotes on each side and allow newlines.
-Like literal strings, there is no escaping whatsoever. A newline immediately
-following the opening delimiter will be trimmed. All other content between the
-delimiters is interpreted as-is without modification.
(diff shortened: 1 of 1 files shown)


ilabdsf:
One issue with ''' escaping is that both lines should have the same line endings. Not sure what to do about it, probably parsers should enforce that line endings are consistent across the TOML document.

dozzie:
I think it's a bad idea to further complicate already complicated string parsing. Literal strings were intended to simplify certain use cases, with basic strings being the fallback when literal string can't do the job.

TheElectronWill:
Just one question: why? 
I agree with @dozzie : literal strings are truely literal and simple, and basic strings are for the other cases.

mojombo:
Thanks for the idea! However, I don't think the benefits of this outweigh the cost. Especially the introduction of the requirement that multiline literal strings must occupy two lines. In any string system there will always be tradeoffs, and certain representations that have to be escaped. I think TOML strikes a good balance as-is.

PR #444, opened 2017-01-18
TOML don't care → TOML parser doesn't care

ilabdsf:


Hrxn:
https://en.wikipedia.org/wiki/Artistic_license

michael-younkin:
I'm pretty sure the grammar error was supposed to be a joke. And even if it wasn't, it's funny, so we should keep it in.

mojombo:
Hah, yes, this derives from the cheeky beginnings of TOML and is true to spirit. =)

PR #452, opened 2017-02-18
Clarify that keys are always strings.

leighmcculloch:
What
===
Add clarification that bare keys are always strings, even if they are
only digits.

Why
===
While it's implied that keys are all the same type and are of the string
type, it's not explicit in the specification. In some libraries using
the specification it's been suggested that bare keys could be
interpreted as other types, e.g. integers, but this is problematic as
most languages cannot effectively handle mixed type keys on maps. We
should be explicit that keys are only strings.

Ref: BurntSushi/toml#154

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,10 +120,12 @@ key = "value"
 
 Keys may be either bare or quoted. **Bare keys** may only contain letters,
 numbers, underscores, and dashes (`A-Za-z0-9_-`). Note that bare keys are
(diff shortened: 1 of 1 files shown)


ChristianSi:
:+1:

leighmcculloch:
@mojombo: What do you think? It's a minor change that brings clarity.

mojombo:
Looks great, thanks!

PR #471, opened 2017-07-10
Mark JavaScript implementations as JavaScript

adamvoss:
As a user
I should be able to search for "JavaScript"
to find JavaScript implementations.

This is more consistent because other entries are languages while Node.js is not a language.  This grew a little complicated because transpiled languages are still useful to a user searching for "JavaScript" so I listed then under JavaScript but also indicated the authored language.


mojombo:
Good point. Thanks!

PR #476, opened 2017-08-13
Show result of example

danny-andrews:
The first thing you see when you read through the README is an example usage of TOML. But without seeing what the result of that example would be in a format one is familiar with, it's not very effective. This fixes that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,53 @@ hosts = [
 ]
 ```
 
+Resultant JSON:
+```json
+{
+  "title": "TOML Example",
+  "owner": {
+    "name": "Tom Preston-Werner",
+    "dob": "1979-05-27T15:32:00.000Z"
+  },
+  "database": {
+    "server": "192.168.1.1",
+    "ports": [
+      8001,
+      8001,
+      8002
+    ],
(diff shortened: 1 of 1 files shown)


jdfergason:
The example that is already there is pretty intuitive.  I think this just adds unnecessary cruft to the introduction of TOML.  TOML stands on it's own; it doesn't need to be related to JSON or YAML and the syntax (for the most part) is intuitive to anyone who programs.

danny-andrews:
I agree that it is fairly intuitive: the resultant JSON isn't surprising by any means. Still, it is a new language, and it is incredibly valuable to have my "intuitions" confirmed by a relation to something I am already familiar with.

If you're uncomfortable with relating it to JSON, then maybe some other more general hashtable format would be suitable.

mojombo:
Hey @danny-andrews, thanks for working up this suggestion. However, I don't think it's necessary. The TOML spec is short enough to be read and understood in a few minutes, so for anyone that's new to TOML and has questions, they should be able to get the answers they need quite quickly. I'd rather not add anything to the spec that doesn't reduce ambiguity. Cheers!

PR #484, opened 2017-11-03
Minor fixes

matklad:
A small clarification to wording, and a more substantial fix to grammar :)


ColinH:
Would it not be easier to define `key` as follows?

`key = unquoted-key / basic-string / literal-string`

matklad:
Fine either way with me :)

PR #490, opened 2017-11-14
Allow multiple newlines and comments in multiline arrays

### LINKED ISSUE #448: ABNF forbids comments in empty arrays
I wouldn't necessarily consider myself an ABNF expert by any means, but I think that a document such as this would be rejected?

```toml
foo = [
    # bar
]
```

Is this a bug or intentional? (it'd be great if it were a bug!)

### LINKED ISSUE #487: Inconsistent comments
Why are the rules for comments so inconsistent and complicated? For example take all the places where empty lines are allowed. In some cases there can be arbitrary many comment lines among the empty lines, in some cases only a single comment is allowed, and in others again there can be no comments.

```
[a]


# Here arbitrary many comments
# can

# be put.


d = [

# This comment is NOT allowed.

1,

# Here only ONE comment is possible.

3
]
```

pradyunsg:
Closes #487
Closes #458
Closes #448

This is a modification to allow an arbitrary number of comments and newline in multiline arrays, instead of a fairly inconsistent format as currently specified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,9 +435,10 @@ arr5 = [ [ 1, 2 ], ["a", "b", "c"] ]
 arr6 = [ 1, 2.0 ] # INVALID
 ```
 
-Arrays can also be multiline. So in addition to ignoring whitespace, arrays also
-ignore newlines, and comments before those newlines, between the brackets.
-Terminating commas (also called trailing commas) are ok before the closing bracket.
+Arrays can also be multiline. Terminating commas (also called trailing commas)
+are ok after the last value of the array. Multiline arrays ignore whitespace
+everywhere between the brackets. There can be an arbitary number of newlines and
+comments before a value and before the closing bracket.
 
 ```toml
 arr7 = [
--- a/README.md
+++ b/README.md
@@ -436,9 +436,8 @@ arr6 = [ 1, 2.0 ] # INVALID
 ```
 
 Arrays can also be multiline. Terminating commas (also called trailing commas)
-are ok after the last value of the array. Multiline arrays ignore whitespace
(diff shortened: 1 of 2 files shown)


ChristianSi:
I would change the ABNF as suggested (provided it works -- didn't check), but leave the README unchanged. The old wording is IMHO good enough, while the new one is confusing and occasionally misleading. ("Multiline arrays ignore whitespace everywhere between the brackets." -- What? Single-line arrays ignore whitespace [i.e., spaces and tabs] too.)

pradyunsg:
> "Multiline arrays ignore whitespace everywhere between the brackets.

Actually, you're right. That sentence is redundant. I'll 🔥 it.

pradyunsg:
@ChristianSi Does this look better?

pradyunsg:
@ChristianSi Does this look better?

ChristianSi:
@pradyunsg Looks much better to me, thanks!

Sadly I'm not in any way "officially" involved with TOML and the maintainers seem to have gone on an extended (read: possible infinite) vacation, so I'm not sure whether that'll get you anywhere...

pradyunsg:
Thanks @ChristianSi!

Indeed. I know that. :)

I just figured it might be a good idea to make it as easy as clicking a button to actually fix some of the things that are needed for a 1.0.0 as per @BurntSushi.

If the maintainers still take forever to come around to this or come around to this in the next minute, I don't mind either way. I do strongly prefer the latter. :P

mojombo:
Hey @pradyunsg, thanks for putting together these ABNF improvements! Merging this one as-is.

PR #491, opened 2017-11-14
Sync ABNF and Specification

### LINKED ISSUE #464: Readme specification of keys is ambiguous
In the readme, it specifies bare keys as follows:

```markdown
Bare keys may only contain letters, numbers, underscores, and dashes `(A-Za-z0-9_-)`
```

This can be confusing, because, for example,

```toml
çe = "value"
```

is not considered valid TOML though I expected it to be. 

If possible, you should either:

1. Avoid specifying letters as `(A-Za-z)` and instead specify letters as "unicode letters and unicode nonspacing marks (to include dependent vowels such as བོ) and numbers as "unicode decimal digists".
2. Specify that letters  and numerals should be valid ASCII and that they may *not* be letters or numerals from utf8. 

Thanks!

### LINKED ISSUE #485: Control character 127
Both the description and the ABNF do not exclude Unicode code point 127, the ASCII "delete" character", in places where all other ASCII control characters are excluded. Probably an oversight that needs to be fixed?

pradyunsg:
Fixes #485 
Fixes #464
Closes #472 
Closes #458
Closes #484

I've made a first pass on the ABNF and README. I think this is close to complete.

Obviously, I'm open to comments and suggestions. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,8 +287,8 @@ int3 = 0
 int4 = -17
 ```
 
-For large numbers, you may use underscores to enhance readability. Each
-underscore must be surrounded by at least one digit.
+For large numbers, you may use underscores between digits to enhance
+readability. Each underscore must be surrounded by at least one digit.
 
 ```toml
 int5 = 1_000
--- a/README.md
+++ b/README.md
@@ -118,12 +118,12 @@ on the same line (though some values can be broken over multiple lines).
 key = "value"
 ```
 
-Keys may be either bare or quoted. **Bare keys** may only contain letters,
-numbers, underscores, and dashes (`A-Za-z0-9_-`). Note that bare keys are
-allowed to be composed of only digits, e.g. `1234`, but are always interpreted
-as strings. **Quoted keys** follow the exact same rules as either basic strings
-or literal strings and allow you to use a much broader set of key names. Best
-practice is to use bare keys except when absolutely necessary.
+Keys may be either bare or quoted. **Bare keys** may only contain ASCII letters,
+ASCII digits, underscores, and dashes (`A-Za-z0-9_-`). Note that bare keys are
+allowed to be composed of only ASCII digits, e.g. `1234`, but are always
+interpreted as strings. **Quoted keys** follow the exact same rules as either
+basic strings or literal strings and allow you to use a much broader set of key
+names. Best practice is to use bare keys except when absolutely necessary.
 
 ```toml
 key = "value"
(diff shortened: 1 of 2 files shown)


pradyunsg:
/ping @BurntSushi 

Pretty sure you got a notification but a mention can't hurt. ;)

pradyunsg:
It might be a good idea to put the ABNF in the README itself. That way, anyone reading the specification can also see the relevant ABNF. Further, the abnf code snippet will make more sense with context. It might mean a script that does sanity checks on the READMe though -- like is the given grammar valid -- but that is no more effort than it is currently to validate the grammar IMO.

Here's a tiny preview of what I'm suggesting. I'm using the `details` tag to hide the ABNF. You can click on it and it'll show. :)

---

Integer
-------

Integers are whole numbers. Positive numbers may be prefixed with a plus sign. Negative numbers are prefixed with a minus sign.

```toml
int1 = +99
int2 = 42
int3 = 0
int4 = -17
```

For large numbers, you may use underscores between digits to enhance readability. Each underscore must be surrounded by at least one digit.

```toml
int5 = 1_000
int6 = 5_349_221
int7 = 1_2_3_4_5     # VALID but discouraged
```

Leading zeros are not allowed. Hex, octal, and binary forms are not allowed. Values such as "infinity" and "not a number" that cannot be expressed as a series of digits are not allowed.

64 bit (signed long) range expected (−9,223,372,036,854,775,808 to 9,223,372,036,854,775,807).

<details>
 <summary>ABNF</summary>

```abnf
integer = [ minus / plus ] int

minus = %x2D                       ; -
plus = %x2B                        ; +

int = DIGIT / digit1-9 1*( DIGIT / underscore DIGIT )
digit1-9 = %x31-39                 ; 1-9
underscore = %x5F                  ; _
```

</details>

mojombo:
These changes all look good, thanks!

> It might be a good idea to put the ABNF in the README itself. That way, anyone reading the specification can also see the relevant ABNF. Further, the abnf code snippet will make more sense with context.

This is an interesting idea, and the [JSON RFC](https://tools.ietf.org/html/rfc7159) relies heavily on inlined ABNF. It's possible that once we have the ABNF nailed down, we should attempt a rewrite of the TOML spec prose to do the same, which would probably reduce a lot of the tortured (and contentious) descriptions of the technical nuances. Let's tackle that in a separate issue or PR.

pradyunsg:
Thanks @mojombo! Great to see you around! ^>^

Lemme know if there's anything I can help with. :)

Issue #389, opened 2016-02-09
Signed zero


jodastephen:
The section on Integers and Floats does not appear to specify whether zero may be signed. ie. whether `+0` or `-0` are allowed, and if so what they mean.

For integers, they could be disallowed or ignored, as +0, 0 and -0 cannot be distinguished.

For floating point numbers some support is probably wise, as floating point numbers typically distinguish between +0 and -0.

adfernandes:
In a similar vein, please reconsider not allowing signed infinities or not-a-number (NaN)s to the spec.

These are very important representations of IEEE floating-point numbers and are often used in configuration files to specify initial values or tolerances.

In fact there are many non-unique NaNs,  usually a default given by LIBC for one particular value, but they are complex and can be signaling or non-signaling.

That is why almost every serialize Asian format for IEEE flooding. Numbers also includes Hex floating-point representations are accepted. There can be a lot of surprising and subtle nuances when converting from a decimal representation to binary representation and vice versa. These conversions are so fraught with non-obvious traps that even LIBC added hex floats years ago!

StefanKarpinski:
Yes to: infinities, signed zeros and a single NaN value. No to different NaN values – this is rarely used for anything meaningful.

adfernandes:
Agreed about the NaNs - apologies if I wasn't clear. A single non-signalling canoncial NaN is fine. That's what `libc` does, after all!

pradyunsg:
> For integers, they could be disallowed or ignored, as +0, 0 and -0 cannot be distinguished.

Made #492 for this.

mojombo:
@StefanKarpinski Would you mind elaborating on the utility of signed zeros for integers? I'm curious what use cases you see.

StefanKarpinski:
To clarify: I don't think signed zeros should be supported for integers at all, although allowing integer zero to be written as `-0`, `0` or `+0` seems reasonable. I was referring to what floating-point values should be supported, which I still think is:

* all finite 64-bit binary IEEE floating-point values **including distinct ±0.0**
* ±Inf
* NaN – but only one of them

[...]

PR #492, opened 2017-11-17
Disallow signed zeros as integers

pradyunsg:
Resolves (partially) #389.

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,7 +280,7 @@ Integer
 -------
 
 Integers are whole numbers. Positive numbers may be prefixed with a plus sign.
-Negative numbers are prefixed with a minus sign.
+Negative numbers are prefixed with a minus sign. Zeros can not be signed.
 
 ```toml
 int1 = +99
--- a/toml.abnf
+++ b/toml.abnf
@@ -72,12 +72,14 @@ array-table-close = ws %x5D.5D  ; ]] Double right quare bracket
 
 ;; Integer
 
-integer = [ minus / plus ] int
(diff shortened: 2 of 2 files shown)


ChristianSi:
I don't like this proposed change. It shouldn't matter whether I write zero as "0", "+0", or "-0".

If forbidding "-0", why not forbidding "+0" as well? This is all unnecessary fiddling. Leave the spec as is (in this regard).

pradyunsg:
Umm, only 0 is valid. -0 and +0 are not. That's what a signed zero is. A zero prefixed with a sign. I'm proposing this.

I thought the language I used was clear. Is it not?

I would like TOML to take a stand in the regard and the proposed one is something I'm more comfortable with than the alternatives. Ideally, floats may have signed 0 based on implementation but yeah integers are better off unsigned. I honestly don't see why anyone would want to use signed zeros in configuration.

StefanKarpinski:
> Umm, only 0 is valid. -0 and +0 are not. That's what a signed zero is.

That’s not what a signed zero is – a signed zero is a distinct zero value whose sign is negative, the syntax is incidental. It’s completely standard in programming languages and data formats to allow + signs on numeric literals and to parse integer `-0` as the same value as `0` and `+0`.

This is not a very controversial position: it is what basically every language and data format does.

pradyunsg:
@StefanKarpinski Let's keep the discussion at #389? :)

mojombo:
I'm closing this in favor of #504 which has a few tweaks I want and doesn't conflict with master.

PR #493, opened 2017-11-17
Clarify certain things about strings

### LINKED ISSUE #473: abnf clarrifications
There are a couple things that were unclear to me comparing the ABNF and the README.

What characters are allowed in literal strings?  The README is not explicit, generally saying "strings without escapes", but the ABNF is explicit with `%x09 / %x20-26 / %x28-10FFFF`.

Can quoted keys use literal strings?  The README says they are basic or literal strings, but the ABNF uses only `quotation-mark`.

### LINKED ISSUE #474: Permit horizontal tabs (ascii 9) in strings
The spec states:

> Basic strings are surrounded by quotation marks. Any Unicode character may be used except those that must be escaped: quotation mark, backslash, and the control characters (U+0000 to U+001F).

which seems to indicate that the horizontal tab character is an illegal control character. Is that right?

`toml.abnf` seems to be split on the matter, allowing `%x09` for `literal-string` and `ml-literal-string` but not for `basic-string` and `ml-basic-string`.

pradyunsg:
Closes #473
Closes #474

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,8 @@ escaped unless their presence would create a premature closing delimiter.
 
 If you're a frequent specifier of Windows paths or regular expressions, then
 having to escape backslashes quickly becomes tedious and error prone. To help,
-TOML supports literal strings where there is no escaping allowed at all.
+TOML supports literal strings which do not allow escaping at all.
+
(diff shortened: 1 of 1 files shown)


mojombo:
These are excellent clarifications. Merging!

PR #494, opened 2017-11-17
Breakup paragraph

pradyunsg:
Make the bold content come at the start of the line. It looks nicer this way.


mojombo:
A fair stylistic point, and I think it'll read fine if we do. We should do it with Multi-line basic strings too, then.

pradyunsg:
I was just reading the spec and thought so too. I'll get to it in a few minutes once I'm properly awake.

PR #497, opened 2017-11-23
Center align the heading

pradyunsg:
It looks nicer this way.


mojombo:
It may look nicer, but I'd rather not have the HTML in the file (it gunks it up). I think it looks fine enough as is.

pradyunsg:
Coolieo. :)

PR #501, opened 2017-11-23
Allow space to separate date and time.

### LINKED ISSUE #424: Allow space to seperate date and time as per RFC3339
With the recent exciting addition of better support for time and date with #414, I think it's time to reconsider the decision to not to allow spaces in date-times  (see #279).

For example, compare the time of the moon landing with/without T:

```
datetime = 1969-07-20T20:18
google_results = 7

datetime = 1969-07-20 20:18
google_results = 5360
```

Another famous (less pleasant) date:

```
datetime = 2001-09-11 10:28
google_results = 87_900

datetime = 2001-09-11T10:28
google_results = 7
```

Seems to me that allowing a space as well would fit the "obvious" part of TOML very well here. The format appears to be several orders of magnitude more popular. In fact, I'd go so far as saying that using T in datetimes is a rather obscure format used only by programmers who have read the Wikipedia page of ISO 8601/RFC3339, as opposed to normal users. While I personally find using a T rather neat, it's not so good for readability.

As for complexity, I believe the addition of allowing a space between date and time `("T" / " ")` would be trivial.

mojombo:
Fixes #424.

RFC 3339 does indeed allow applications to use a space instead of a "T" to delimit date and time. It's certainly easier to read. I'd enjoy thoughts on this. Any downside?


toml.abnf
```diff
@@ -139,6 +139,7 @@ date-time      = offset-date-time / local-date-time / local-date / local-time
 date-fullyear  = 4DIGIT
 date-month     = 2DIGIT  ; 01-12
 date-mday      = 2DIGIT  ; 01-28, 01-29, 01-30, 01-31 based on month/year
+time-delim     = "T" / " "
```
pradyunsg: Maybe use `%x20` instead of `" "`?

mojombo: Yeah, that's probably wise to eliminate any confusion.

pradyunsg:
> Any downside?

None that I can see. I think the readability gain is great + this is sort of the obvious human way to do it. Considering it made it into the RFC, I'd say that there's nothing that can cause it to trip.

pradyunsg:
:shipit:?

PR #504, opened 2017-11-25
Zeros (int and float) cannot have have a sign (+ or -).

mojombo:
Replacement for #492 with a few wording tweaks and extended to Floats as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## HEAD
 
+* Specify that zeros (Integer 0 and float 0.0) cannot have a sign (+ or -).
 * Rename Datetime to Offset Date-Time.
 * Add Local Date-Time.
 * Add Local Date.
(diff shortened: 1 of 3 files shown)


ChristianSi:
This is terribly, utterly illogical. So I can write 1, +1, and -1 (the latter meaning 0 - 1) and likewise for all other integers EXCEPT 0, where the harmless "no-op" + and the negatived form (-0, meaning 0 - 0, and in this case a no-op too) are suddenly forbidden.

Imagine someone finds a "+1" somewhere in their config. Noticing that value is off, they change it to "+0". Now the config is broken and does not longer parse. WTF???

PLEASE DO NOT MERGE until you have come up with a good explanation to that poor config user why that seemingly innocuous change has indeed to be punished with unparseability. The much more logical course of action, of course, would be to abandon this PR and treat these harmless little signs in the same way as they are treated in (probably) every programming language every invented.

And, of course, in the same way as TOML has treated them so far. DON'T BREAK WHAT'S NOT BROKEN.

mojombo:
Closing this per discussion here and in #389. I'll open a new one for IEEE 754 support in a bit.

Issue #409, opened 2016-04-28
Reconsider hex and/or octal integer formats


rmunn:
Issue #53 was closed in June 2014, because the decision at the time was to prefer simplicity of implementation. So because `0xff00ff` or `0o755` were slightly harder to write parsers for than `16711935` or `493`, the choice at the time was not to allow hex or octal numbers in TOML.

However, since that time, issue #263 has been decided the other way. Datetime values are non-trivial to parse, but are highly useful in some scenarios. So the decision was made to keep them in, because they are useful to some real users.

These two decisions are inconsistent. If datetimes are going to be in TOML, the same arguments can be (and have been) made for hex and octal representations of numbers, which are a lot easier to write a parser for than datetimes. Most languages already have a hex parser implementation that TOML parsers could take advantage of. And in any language that doesn't, parsing hex values is not complex. It's a  problem with "Coding 101 homework" levels of difficulty, not "doctoral thesis" levels of difficulty.

And hex and octal values are useful in many scenarios that TOML is intended for, such as config files. Unix permissions use octal values: `0o755` is much easier to mentally translate to `u+rwx, g+rx, o+rx` than `491`. Or was it `493` or `495`? Quick, can you tell which of those three decimal values is the _correct_ conversion of `0o755`? I can't without a calculator, and I'd much rather see `0o755` in config files. Hex, of course, is highly useful when dealing with colors or bit flags. Neither are as common in config files as octal, but if we allow octal there's no good reason not to allow hex.

Therefore, I would ask that #53 be revisited, either be reopening that issue and having the discussion there, or by starting a new discussion here. The reason for closing #53, to keep things simple for TOML implementations, has been abandoned by now, and there's no longer any reason not to allow hex and octal values.

rmunn:
Also, I'll repeat the comment I made on issue #53 last month: if octal values are included, PLEASE don't repeat C's mistake, as so many programming languages have. A leading 0 in an integer should not change its meaning. The format for octal should parallel the format for hex: `0x123` for hex, and `0o123` for octal. (And `0b101` for binary, if binary is allowed.)

Additionally, I would recommend that the only integer format markers allowed be lowercase: `0x`, `0o`, and possibly `0b`. In particular, I suggest that `0O` (digit zero followed by capital letter O) be **forbidden** by the spec. It's too easy to mistake those two characters for each other in many fonts.

[...]

PR #506, opened 2017-11-28
Add non-finite floats and clarify signed zeros

### LINKED ISSUE #389: Signed zero

mojombo:
Closes #389.

This adds support for non-finite float values (`-inf`, `inf`, `+inf`, `nan`) and clarifies that signed zeros (integer and float) are valid and what they mean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## HEAD
 
-* Add non-finite float values (+inf, inf, -inf, nan)
+* Add special float values (inf, nan)
 * Rename Datetime to Offset Date-Time.
 * Add Local Date-Time.
 * Add Local Date.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## HEAD
 
+* Add non-finite float values (+inf, inf, -inf, nan)
 * Rename Datetime to Offset Date-Time.
 * Add Local Date-Time.
 * Add Local Date.
@@ -18,6 +19,7 @@
 * Clarify that literal strings can be table keys.
 * Clarify that at least millisecond precision expected for Date-Time and Time.
 * Clarify that comments are OK in multiline arrays.
+* Clarify that +0, -0, +0.0, and -0.0 are valid and what they mean.
 * TOML has a logo!
 
 ## 0.4.0 / 2015-02-12
--- a/README.md
+++ b/README.md
@@ -305,7 +305,7 @@ identical to an unprefixed zero. Hex, octal, and binary forms are not allowed.
 Float
 -----
 
-Floats should be implemented as 64-bit IEEE 754 values.
+Floats should be implemented as IEEE 754 binary64 values.
 
 A float consists of an integer part (which follows the same rules as integer
 values) followed by a fractional part and/or an exponent part. If both a
@@ -347,7 +347,7 @@ Non-finite float values can also be expressed. They are always lowercase.
 nf1 = +inf # positive infinity
 nf2 = inf  # positive infinity
 nf3 = -inf # negative infinity
-nf4 = nan  # not a number
+nf4 = nan  # not a number; actual sNaN/qNaN value is implementation specific
 ```
 
 Boolean
--- a/README.md
+++ b/README.md
@@ -341,7 +341,7 @@ flt8 = 9_224_617.445_991_228_313
 
 Float values `-0.0` and `+0.0` are valid and should map according to IEEE 754.
 
-Non-finite float values can also be expressed. They are always lowercase.
+Special float values can also be expressed. They are always lowercase.
 
 ```toml
 nf1 = +inf # positive infinity
--- a/README.md
+++ b/README.md
@@ -296,15 +296,16 @@ int6 = 5_349_221
 int7 = 1_2_3_4_5     # VALID but discouraged
 ```
 
-Leading zeros are not allowed. Hex, octal, and binary forms are not allowed.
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -296,16 +296,17 @@ int6 = 5_349_221
 int7 = 1_2_3_4_5     # VALID but discouraged
 ```
 
-Leading zeros are not allowed. Hex, octal, and binary forms are not allowed.
-Values such as "infinity" and "not a number" that cannot be expressed as a
-series of digits are not allowed.
+Leading zeros are not allowed. Integer values `-0` and `+0` are valid and
+identical to an unprefixed zero. Hex, octal, and binary forms are not allowed.
```
pradyunsg: #409?

README.md
```diff
@@ -296,16 +296,17 @@ int6 = 5_349_221
...
 
 64 bit (signed long) range expected (−9,223,372,036,854,775,808 to
 9,223,372,036,854,775,807).
 
 Float
 -----
 
+Floats should be implemented as 64-bit IEEE 754 values.
```
pradyunsg: It should say IEEE 754 binary64 values, since there's a 64-bit decimal64 as well.

README.md
```diff
@@ -338,7 +339,16 @@ underscore must be surrounded by at least one digit.
...
-64-bit (double) precision expected.
+Float values `-0.0` and `+0.0` are valid and should map according to IEEE 754.
+
+Non-finite float values can also be expressed. They are always lowercase.
+
+```toml
+nf1 = +inf # positive infinity
+nf2 = inf  # positive infinity
+nf3 = -inf # negative infinity
+nf4 = nan  # not a number
```
pradyunsg: IEEE 754 has 2 types of NaNs -- sNaN and qNaN.

It might be worthwhile to mention that which one it is would be implementation dependant?

toml.abnf
```diff
@@ -117,14 +117,19 @@ underscore = %x5F                  ; _
 
 ;; Float
 
-float = integer ( frac / ( frac exp ) / exp )
+float =  integer ( frac / ( frac exp ) / exp )
```
pradyunsg: maybe now would be a good time to change this to?

```abnf
float = integer ( exp / frac [ exp ] )
```

toml.abnf
```diff
@@ -117,14 +117,19 @@ underscore = %x5F                  ; _
...
+float =/ non-finite
```
pradyunsg: Maybe call them "named float values"?

README.md
mojombo: Yep, one thing at a time!

README.md
mojombo: Ah, fair enough, will change.

README.md
mojombo: Yeah, and many of each! I'll add a note as you suggest.

toml.abnf
mojombo: Seems reasonable.

toml.abnf
mojombo: From a bit of googling, it seems like "special values" is the preferred nomenclature. How does that strike you?

pradyunsg: Works well. :)

toml.abnf
```diff
@@ -117,14 +117,19 @@ underscore = %x5F                  ; _
 
 ;; Float
 
-float = integer ( frac / ( frac exp ) / exp )
+float = integer ( exp / frac [ exp ] )
+float =/ non-finite
```
pradyunsg: `float-special-values` then?

mojombo: I went with just `special-float`. `values` seems redundant.

pradyunsg: `special-floats` works. :)

mojombo:
@pradyunsg Haha, thanks. =) I'll wait for a few more thoughts from others before final merge.

toml.abnf
```diff
@@ -117,14 +117,19 @@ underscore = %x5F                  ; _
...
+float = integer ( exp / frac [ exp ] )
+float =/ special-float
 
 frac = decimal-point zero-prefixable-int
 decimal-point = %x2E               ; .
 zero-prefixable-int = DIGIT *( DIGIT / underscore DIGIT )
 
 exp = "e" integer
 
+special-float = ( [ minus / plus ] inf ) / nan
```
StefanKarpinski: I wonder if allowing `±` in front of `nan` also would be simpler – keeping the universal rule that numbers can have a sign prefix. Of course `+nan`, `nan` and `-nan` would all mean the same thing, but it could be simpler. Not a big issue, just a thought.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## HEAD
 
+* Add non-finite float values (+inf, inf, -inf, nan)
```
StefanKarpinski: Perhaps only list one of `+inf` and `inf` here – since this says "values" and those are the same value written with different syntaxes.

toml.abnf
mojombo: That's an interesting point. It would keep the consistency across all values, which would be a nice behavior. I suppose if the implementation wanted to, it could carry over the sign to the NaN value, even though it's insignificant, but who knows!

CHANGELOG.md
mojombo: Fair enough, will tighten in up.

PR #507, opened 2017-11-29
Add hex, octal, and binary integer formats

### LINKED ISSUE #409: Reconsider hex and/or octal integer formats

mojombo:
Closes #409.

```
# hexadecimal with prefix `0x`
hex1 = 0xDEADBEEF
hex2 = 0xdeadbeef
hex3 = 0xdead_beef

# octal with prefix `0o`
oct1 = 0o01234567
oct2 = 0o755 # useful for Unix file permissions

# binary with prefix `0b`
bin1 = 0b11010110
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,17 +299,16 @@ int7 = 1_2_3_4_5     # VALID but discouraged
 Leading zeros are not allowed. Integer values `-0` and `+0` are valid and
 identical to an unprefixed zero.
 
-Integer values may also be expressed in hexadecimal, octal, or binary. Hex
-values are case insensitive. Leading zeros are allowed. Underscores are allowed
-between digits (but not between the prefix and the value).
+Non-negative integer values may also be expressed in hexadecimal, octal, or
+binary. In these formats, leading zeros are allowed (after the prefix). Hex
+values are case insensitive. Underscores are allowed between digits (but not
+between the prefix and the value).
 
 ```toml
 # hexadecimal with prefix `0x`
 hex1 = 0xDEADBEEF
 hex2 = 0xdeadbeef
 hex3 = 0xdead_beef
-hex4 = +0xDEADBEEF # same as above
-hex5 = -0xDEADBEEF # negative of above
 
 # octal with prefix `0o`
 oct1 = 0o01234567
--- a/README.md
+++ b/README.md
@@ -297,7 +297,27 @@ int7 = 1_2_3_4_5     # VALID but discouraged
 ```
 
 Leading zeros are not allowed. Integer values `-0` and `+0` are valid and
-identical to an unprefixed zero. Hex, octal, and binary forms are not allowed.
+identical to an unprefixed zero.
+
+Integer values may also be expressed in hexadecimal, octal, or binary. Hex
+values are case insensitive. Leading zeros are allowed. Underscores are allowed
+between digits (but not between the prefix and the value).
+
+```toml
+# hexadecimal with prefix `0x`
+hex1 = 0xDEADBEEF
+hex2 = 0xdeadbeef
+hex3 = 0xdead_beef
+hex4 = +0xDEADBEEF # same as above
+hex5 = -0xDEADBEEF # negative of above
+
+# octal with prefix `0o`
+oct1 = 0o01234567
+oct2 = 0o755 # useful for Unix file permissions
+
+# binary with prefix `0b`
+bin1 = 0b11010110
+```
 
 64 bit (signed long) range expected (−9,223,372,036,854,775,808 to
 9,223,372,036,854,775,807).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## HEAD
 
+* Add hex, octal, and binary integer formats.
 * Add special float values (inf, nan)
 * Rename Datetime to Offset Date-Time.
 * Add Local Date-Time.
--- a/toml.abnf
+++ b/toml.abnf
@@ -106,25 +106,35 @@ ml-literal-char = %x09 / %x20-10FFFF
 
 ;; Integer
 
-integer = [ minus / plus ] int
+integer = [ minus / plus ] ( dec-int / hex-int / oct-int / bin-int )
 
 minus = %x2D                       ; -
 plus = %x2B                        ; +
-
-int = DIGIT / digit1-9 1*( DIGIT / underscore DIGIT )
-digit1-9 = %x31-39                 ; 1-9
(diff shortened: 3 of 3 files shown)


mojombo:
> You've additionally defined DIGIT and ALPHA at the end. You might want to define HEXDIG at the end.

Of the ABNF? It already is.

pradyunsg:
Ah right. I should wake up properly.

README.md
```diff
@@ -297,7 +297,27 @@ int7 = 1_2_3_4_5     # VALID but discouraged
 ```
 
 Leading zeros are not allowed. Integer values `-0` and `+0` are valid and
-identical to an unprefixed zero. Hex, octal, and binary forms are not allowed.
+identical to an unprefixed zero.
+
+Integer values may also be expressed in hexadecimal, octal, or binary. Hex
+values are case insensitive. Leading zeros are allowed. Underscores are allowed
```
rmunn: Might want to clarify that the "Leading zeros are allowed" sentence here applies only to prefixed hex, octal, and/or binary literals. Otherwise there appears to be a contradiction in the spec between "Leading zeros are not allowed" in the previous paragraph and "Leading zeros are allowed" in this paragraph.

README.md
```diff
@@ -297,7 +297,27 @@ int7 = 1_2_3_4_5     # VALID but discouraged
...
+between digits (but not between the prefix and the value).
+
+```toml
+# hexadecimal with prefix `0x`
+hex1 = 0xDEADBEEF
+hex2 = 0xdeadbeef
+hex3 = 0xdead_beef
+hex4 = +0xDEADBEEF # same as above
+hex5 = -0xDEADBEEF # negative of above
```
rmunn: What are the semantics of negative hex literals if the literal is large enough to have its 63 bit set? E.g., since integer literals are considered (a couple lines further down in the spec) to be signed longs (signed 64-bit integers), what are the semantics of `0x8000_0000_0000_0000` or `0xffff_ffff_ffff_ffff`? And what about the semantics of `-0x8000_0000_0000_0000` and `-0xffff_ffff_ffff_ffff`?

Most people will probably implement this spec such that `0xffff_ffff_ffff_ffff` becomes -1, and thus `-0xffff_ffff_ffff_ffff` will end up producing the value 1. And then both `0x8000_0000_0000_0000` and `-0x8000_0000_0000_0000` will end up producing the value -9,223,372,036,854,775,808 (which is the corner case of 64-bit int negation via two's complement), whereas `0x8000_0000_0000_0001` will produce -9,223,372,036,854,775,807 and thus `-0x8000_0000_0000_0001` will produce +9,223,372,036,854,775,807.

I think it might be wise to document those specific corner cases, either a few examples in the spec or with a few test cases to ensure that all implementations handle them identically.

franklinyu: @rmunn How about letting negative number follows similar "loosen-later" policy? What't the use case of negative hexadecimal or binary?

rmunn: That's one of the things I've been wondering, and my original proposal was to disallow negative hex numbers since their semantics aren't totally clear in edge cases. (And languages vary about whether they're allowed: for example, in F#, `0x8000000000000000L` (the L is required in F# and means "this is a `long` literal) is -9,223,372,036,854,775,808 (and is of type "signed long"), and `-0x8000000000000000L` is *also* -9,223,372,036,854,775,808 since that's the one bit pattern that is the two's complement of itself. Java appears to follow the same semantics as F#, at least if http://www.javarepl.com/ is to be trusted. However, in C#, `0x8000000000000000` is *positive* 9,223,372,036,854,775,808 (and is of type "unsigned long"), and `-0x8000000000000000` produces `error CS0023: The '-' operator cannot be applied to operand of type 'ulong'`.

So since there's not one consistent way to handle negative hex numbers among major languages, my original proposal was to forbid them in TOML. However, @mojombo is the one whose name is going on the language, so it's ultimately his call. Since he wants to allow negative hex literals, I won't argue with that decision, and I'll just point out corner cases that need to be covered, at least in the tests if not in the spec.

mojombo: All fair points. I originally allowed it for consistency sake, but you're right, I don't see any compelling use cases, and it opens up the can of worms you've stated. I just pushed a commit that revises this proposal to disallow signed hex/oct/bin values.

README.md
mojombo: Clarified in the commit I just pushed.

rmunn:
With the change to disallowing signed hex/oct/bin literals, this LGTM.

I'll just note my comment at https://github.com/toml-lang/toml/issues/409#issuecomment-347796827 where I mentioned that C# 7.2 has just changed the rules to allow underscores between hex prefixes and the first digit (and similarly for oct and bin). I.e., `0x_dead_beef` is now valid in C# 7.2, where it wasn't before. I talk about this a bit in that comment, but the TL;DR is that I think we should leave the rules exactly as they are in the current spec for now, and if general language consensus shifts towards allowing `0x_dead_beef` in other languages besides C#, then we can allow that since it would be backwards-compatible. (Previously-invalid literals would become valid, but all previously-valid literals would remain valid).

PR #512, opened 2018-01-08
typo: [fruit.physical] should be double-bracketed

xftroxgpx:
Right? Or am I missing something obvious? :)
Cheers!


xftroxgpx:
actually, I'm wrong :)
(according to the mapped JSON that follows)

PR #519, opened 2018-02-05
Remove 127 as a control character in ML strings

### LINKED ISSUE #518: Control character 127 inconsistent
#485 fixed it so that 0x7f (127) is properly considered a "control character".  However, only the definition of basic string was updated.  Both the spec and the ABNF still allow 0x7f in multiline basic strings, literal strings, and multiline literal strings.  Was this intentional?

pradyunsg:
Fixes #518


ehuss:
Should this also apply to the literal strings in the ABNF?  They forbid other control characters.

PR #520, opened 2018-02-05
Update inline table example with dotted keys

DraagrenKirneh:
Add example for dotted key in inline table for consistency.


eksortso:
This section is immediately followed by an equivalent standard-table version. We need to have `animal.type.name` referenced in that.

DraagrenKirneh:
Thx, I have updated the readme

PR #531, opened 2018-04-07
Fix the exponent part of ABNF

bicycle1885:
The specs clearly state that an exponent part may start with an "e" or "E" character but the ABNF disallows "E". This change resolves the conflict between them.


pradyunsg:
Hey @bicycle1885! Literal Strings in ABNF are case insensitive. See page 4 on https://tools.ietf.org/html/rfc5234.

bicycle1885:
Oops! Thank you for pointing it out.

pradyunsg:
You're welcome.

Have a good day/night. :)

PR #505, opened 2017-11-25
Dotted keys

pradyunsg:
Closes #499

This will probably conflict with #500 since this also adds a "Keys" section.

PR #536, opened 2018-05-11
Remove incorrect/redundant statements

pradyunsg:
#505 wasn't super comprehensive with that.


pradyunsg:
Merging since I expect this to be non-controversial anyway.

PR #542, opened 2018-05-24
Add version pragma

### LINKED ISSUE #522: Version pragma
Back in my Erlang days, when I was working on [BERT-RPC](http://bert-rpc.org/), I gave a talk about it at the Erlang Conference in Stockholm. Joe Armstrong was there, and after the speech, during questions, he spoke up and said "It appears you've made the number one biggest mistake in protocol design: not having a way to specify the protocol version number in use." Well, in my defense, almost nobody adopted BERT-RPC and therefore a version number was irrelevant. SO THERE, JOE.

But TOML has already enjoyed more success than my esoteric binary format, and I expect that versions beyond 1.0 will exist. In that future universe, TOML documents of several disparate versions will be laying around, and I must ask the question: do we need an official way to specify what version a TOML document intends to parse it?

I could, indeed, argue that no, it's not necessary, because a TOML document expects to be configuration, and a configuration file will be consumed by an application, and that application will specify and know what version of TOML it is parsing. However, I could then argue that applications also evolve over time, and perhaps a program would like to accept either TOML v1 or v2 or v3, etc, in which case it would rather rely on the document specifying a version, rather than having to use some feature detection heuristics, or coming up with its own application-specific way to document the version in use.

So perhaps we solve this preemptively in v1.0.0 so that future versions may rest easy knowing that a solution exists for when the time comes that v1.1 or v2.0 is released. Maybe it looks like this:

```toml
# TOML v1.0
```

A few questions for consideration:

1. Optional or required?
2. Top of file, bottom, anywhere?
3. What does the `v` specification mean exactly? Does `1.0` mean any `1.0.x` version? Do we need to use a `~` operator and get fancy like npm dependencies?

I have a few thoughts on these matters that I will elaborate, but I'm sitting on a 777 about to blast off and will be sans-internet for the next 11 hours. In the meantime, I'd love any thoughts on the matter or prior art that you've seen solve this problem elegantly.

pradyunsg:
Closes #522

### DIFF
--- a/README.md
+++ b/README.md
@@ -801,9 +801,12 @@ the document is known to be compatible with. It takes the form:
 # TOML v1.0
 ```
 
-The version pragma should occur at the start of the document before any keys are specified. The version should be the major and minor version numbers.
+The version pragma should occur at the start of the document before any keys
+are specified. The version should be the major and minor version numbers.
 
-The expectation is that an implementation capable of parsing a version pragma only accepts such a document if it has a matching major version and equal or greater minor version.
+The expectation is that an implementation capable of parsing a version pragma
+only accepts such a document if it has a matching major version and equal or
+greater minor version.
 
 Filename Extension
 ------------------
--- a/README.md
+++ b/README.md
@@ -791,6 +791,17 @@ points = [ { x = 1, y = 2, z = 3 },
            { x = 2, y = 4, z = 8 } ]
 ```
 
+Version Pragma
+--------------
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -791,6 +791,23 @@ points = [ { x = 1, y = 2, z = 3 },
...
+the document is known to be compatible with. It takes the form:
+
+```
+# TOML v1.0
+```
+
+The version pragma should occur at the start of the document before any keys
+are specified. The version should be the major and minor version numbers.
+
+The expectation is that an implementation capable of parsing a version pragma
```
pradyunsg: This could probably use some rewording.

PR #543, opened 2018-05-31
Correct restriction on integer length in ebnf definition

cleishm:
The integer rule in the ebnf contains `digit1-9 1*( DIGIT / underscore DIGIT )`. This should probably be `digit1-9 *( DIGIT / underscore DIGIT )` (removing the length 1 restriction).

### DIFF
--- a/toml.abnf
+++ b/toml.abnf
@@ -122,7 +122,7 @@ oct-prefix = %x30.6f               ; 0o
 bin-prefix = %x30.62               ; 0b
 
 dec-int = [ minus / plus ] unsigned-dec-int
-unsigned-dec-int = DIGIT / digit1-9 1*( DIGIT / underscore DIGIT )
+unsigned-dec-int = DIGIT / digit1-9 *( DIGIT / underscore DIGIT )
 
 hex-int = hex-prefix HEXDIG *( HEXDIG / underscore HEXDIG )
 oct-int = oct-prefix digit0-7 *( digit0-7 / underscore digit0-7 )


lmna:
Without length 1 restriction it becomes ambiguous with left-hand alternative:
<pre>+unsigned-dec-int = <b>DIGIT</b> / digit1-9 1*( DIGIT / underscore DIGIT )</pre>

lmna:
@cleishm 

https://tools.ietf.org/html/rfc5234#section-3.6

```
3.6.  Variable Repetition:  *Rule

   The operator "*" preceding an element indicates repetition.  The full
   form is:

         <a>*<b>element

   where <a> and <b> are optional decimal values, indicating at least
   <a> and at most <b> occurrences of the element.

   Default values are 0 and infinity so that *<element> allows any
   number, including zero; 1*<element> requires at least one;
   3*3<element> allows exactly 3; and 1*2<element> allows one or two.
```

cleishm:
@lmna Ok! So it’s already correct. My mistake and thanks for the clarification.

cleishm:
@lmna BTW - thanks for the documentation reference. I was looking for a decent description of ebnf. My mistake was assuming `1*` meant "1 repetition _only_". It's much clearer now :)

pradyunsg:
The syntax of that file is ABNF. http://www.ietf.org/rfc/rfc5234.txt is the spec, as linked above.

A good comparison between ABNF/EBNF is at http://matt.might.net/articles/grammars-bnf-ebnf/.

PR #556, opened 2018-08-20
Fix exp ABNF definition and time-delim comment

swissr:
- acc. to README, 'E' and 'e' are allowed
- acc. to RFC (and the actual abnf definition) only 'T' is allowed as time-delim


ehuss:
I believe that ABNF is case-insensitive, so this shouldn't be necessary.  Also, RFC 3339 explicitly says it can be lowercase T.

eksortso:
Strings in ABNF are case-insensitive by default. So these changes aren't necessary.

pradyunsg:
As everyone else has helpfully stated, these changes are not needed.

The ABNF specification clearly states that quoted strings are case insensitive. For cases where case sensitivity matters (like `true`), we've used the %dd format.

Thanks for the PR though. :)

swissr:
Sorry about the noise. Didn't catch that string literals in ABNF aren't case sensitive (iirc in EBNF/BNF case matters) and also got confused because in one place the definition used lower case ("e") and in other places capital case ("T" and "Z"). Did check RFC 3339 but missed _'NOTE: Per [ABNF] and ISO8601, the "T" and "Z" characters in this syntax may alternatively be lower case "t" or "z" respectively'_ (has been mentioned above). All fine, thanks for the friendly comments.

PR #557, opened 2018-08-20
Simplify inline-table ABNF definition

swissr:
Not 100 % sure but I don't see a reason not to use optional directly in the 'inline-table' definition.

### DIFF
--- a/toml.abnf
+++ b/toml.abnf
@@ -212,14 +212,13 @@ std-table-close = ws %x5D     ; ] Right square bracket
 
 ;; Inline Table
 
-inline-table = inline-table-open inline-table-keyvals inline-table-close
+inline-table = inline-table-open [ inline-table-keyvals ] inline-table-close
 
 inline-table-open  = %x7B ws     ; {
 inline-table-close = ws %x7D     ; }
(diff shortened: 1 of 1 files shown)


toml.abnf
```diff
@@ -212,14 +212,13 @@ std-table-close = ws %x5D     ; ] Right square bracket
 
 ;; Inline Table
 
-inline-table = inline-table-open inline-table-keyvals inline-table-close
+inline-table = inline-table-open [ inline-table-keyvals ] inline-table-close
```
bitwalker: This change means that at least one key/value pair is required - the old way of expressing it allowed for "empty" tables.

lmna: @bitwalker New way allows tables that do not have `inline-table-keyvals` at all. Is it not the same as "empty tables"?

bitwalker: Ah, yep, you're right, I was looking at the old definitions of `inline-table-*` and forgot about the `[]` around `inline-table-keyvals` up here.

pradyunsg:
Thanks for this PR @swissr! This is definitely an improvement over how it is currently. :)

Issue #453, opened 2017-03-07
Global keys defined below array of tables


bartekn:
Consider the following toml:
```toml
[[song]]
name = "Thunder Road"
duration = "4m49s"

[[song]]
name = "Stairway to Heaven"
duration = "8m03s"

another = "value"
```
The specification isn't clear if `another` should be a part of `song` or a global value. [tomlv](https://github.com/BurntSushi/toml/tree/master/cmd/tomlv) from Validators section in the README file returns the following result:
```
song               ArrayHash
    song.name      String
    song.duration  String
song               ArrayHash
    song.name      String
    song.duration  String
    song.another   String
```
I think the specification should be updated with either:
* Allowing setting global values after array of tables (after a blank line) or
* Explicitly specifying that every `key = value` line (even after a blank line) will be considered a key the preceding array.

BurntSushi:
What part of this is unclear?

> Tables (also known as hash tables or dictionaries) are collections of key/value pairs. They appear in square brackets on a line by themselves. You can tell them apart from arrays because arrays are only ever values.
>
> ```
> [table]
>```
>
> Under that, and **until the next table or EOF are the key/values of that table.** Key/value pairs within tables are not guaranteed to be in any specific order.

bartekn:
OK, I missed that because I was checking [Array of Tables](https://github.com/toml-lang/toml#array-of-tables) section. I think that copying this sentence there can be helpful for someone in a future. Thanks!

xpe:
I agree completely with @bartekn -- this confusion could be easily remedied with some documentation updates to clarify the behavior. I specifically mean documentation for arrays of tables -- not just tables.

PR #558, opened 2018-09-06
Clarify what happens after an array of tables.

xpe:
This PR is an outgrowth of issue #453. The current specification says nothing about what happens below an array of tables. This results in ambiguity; this PR adds text to clarify.

### DIFF
--- a/versions/en/toml-v0.5.0.md
+++ b/versions/en/toml-v0.5.0.md
@@ -701,6 +701,32 @@ In JSON land, that would give you the following structure.
 }
 ```
 
+Newlines do **not** end the scope of an array of tables. Key/value pairs belong
+to the most recent double bracketed name. For example:
+
+```toml
+[[song]]
+name = "Thunder Road"
+duration = "4m49s"
+
+[[song]]
+name = "Stairway to Heaven"
+duration = "8m03s"
+
+another = "value"
+```
+
+The above parses to this JSON (note the position of the `"another"` key):
+
+```json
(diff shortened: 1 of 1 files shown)


pradyunsg:
I would do this slightly differently, modifying the existing example itself (add a new line before `color = "gray"`) and adding a sentence in the same vein to the paragraph above as the existing one in the Table section:

> Under that, and until the next table or EOF are the key/values of that table.

xpe:
The current documentation needs clarification in this regard, so I welcome any changes that improve it. 

@pradyunsg Your suggested changes generally sound ok to me, but I can't envision exactly what you have in mind. Would you please go ahead and make and commit the suggestions that you recommend? (In my experience, when I try to wordsmith for someone else, the results are not as effective as if the other person were to just write what they want to see.)

pradyunsg:
Sure! I'll just push them here when I'm near a PC.

Please feel free to ping me if I forget to do so since at least the next two weeks are super busy for me.

marcos-abreu:
@pradyunsg - you probably forgot about this PR.

I found it when facing this problem myself (even after reading the current docs). I think this is really helpful

marco-m:
@pradyunsg ping :-)

pradyunsg:
I did forget about this completely, yes. :)

pradyunsg:
Thanks @xpe for the initial PR and @marcos-abreu + @marco-m for the pings. :)

PR #561, opened 2018-09-20
Clarify that float integer parts are decimal digits only

### LINKED ISSUE #560: Are hexadecimal / octal / binary integers allowed in floats?
From the doc:

> A float consists of an integer part (which follows the same rules as integer values) followed by a fractional part and/or an exponent part.

This seems to imply that something like `0x56.78` is a valid float, equal to `86.78`. This is unusual enough that it should be explicitly documented if this is intended (and if it isn't, then the documentation should be updated to not seem to say that).

(FWIW, in my opinion this should be explicitly disallowed, because inventing a new mixed-radix format that conflicts with the standard hex-float format seems bad.)

pradyunsg:
Fixes #560

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,7 +415,7 @@ flt7 = 6.626e-34
 A fractional part is a decimal point followed by one or more digits.
 
 An exponent part is an E (upper or lower case) followed by an integer part
-(which follows the same rules as integer values).
+(which follows the same rules as decimal integer values).
 
 Similar to integers, you may use underscores to enhance readability. Each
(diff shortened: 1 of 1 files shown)


percontation:
Line 481 (which says how the exponent part follows the same rules as an integer) could also benefit from inserting the word "decimal" :)

pradyunsg:
> Line 418 (which says how the exponent part follows the same rules as an integer) could also benefit from inserting the word "decimal" :)

Nice catch. :)

PR #564, opened 2018-10-02
Add `MIME Type` to table of contents

y-yagi:
Since all the other contents have links, I thought that it would be good to add `MIME Type`. Thanks!


pradyunsg:
Thanks @y-yagi! 🎉

y-yagi:
Thanks you :heart:

PR #587, opened 2019-01-17
Repair space between words

LongTengDao:
No change for content.
There was one space lacked between words.


pradyunsg:
Hey @LongTengDao!

Could you go through the rest of the document one and confirm if there are any more such minor changes? :)

LongTengDao:
@pradyunsg Sorry for being entangled in the details, your words are also what I thought. The changes in previous PR is collected for a few weeks, and after going through some times, I was happy to submit it. But sadly, just after I submit it, I found this one. After hesitation, I create this PR. After another times going through, I think it should be prefect.

PR #590, opened 2019-01-24
Support standalone escape in multiline strings

bd82:
An escape was not allowed in a multi line string
except as part of an escape sequence or a "line ending backslash"

So the following would be illegal according to this grammar:
```
myPattern = """\d{2}"""
```
was this the intent?

I suggest removing the specific syntax for the "line ending backslash"
because it is already included in the defined grammar (all backslash, ws and NL are  already allowed inside a multi-line string).
One could consider "line ending backslash" as the semantic behavior  that should be implemented in a compiler
rather than the syntactic definition that a parser is responsible for reading.

### DIFF
--- a/toml.abnf
+++ b/toml.abnf
@@ -84,9 +84,9 @@ ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
 
 ml-basic-string-delim = 3quotation-mark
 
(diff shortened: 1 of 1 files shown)


bd82:
Right I realized the backslash is only allowed as a line ending backslash 😄

Issue #566, opened 2018-10-06
Unclear how comments in strings are handled


jorisvr:
It is not clear from the specification how comments inside quoted strings are handled.
For example:

    a = """ # is this a comment ?
    # and this ? """

On the one hand, the **String** section states that "_Any unicode character may be used except ..._"
However, the **Comments** section states unconditionally that "_A hash symbol marks the rest of the line as a comment._"

It is clear from the ABNF definition that # inside a quoted string is just string content, but the ABNF is currently not authoritative.

pradyunsg:
I think this can be considered a common knowledge thing that a # in quotes won't start a comment.

Will be happy to get a PR clarifying this if more people feel that it's somehow ambiguous.

ChristianSi:
I'd say it's self-evident and no further clarification is needed. Otherwise we'd also have to explain that '[' does NOT open a table head if it happens to occur within a quoted string, that '=' within a quoted string does not separate a key from its value, and so on.

pradyunsg:
I think it's clear what to do here.

PR #598, opened 2019-02-26
Be explicit about the use of # inside strings

rougeth:
A hash symbol should **not** mark the rest of the line as a comment if it is inside a string.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,11 +102,12 @@ Spec
 Comment
 -------
 
-A hash symbol marks the rest of the line as a comment.
+A hash symbol marks the rest of the line as a comment, except when inside a string.
 
 ```toml
 # This is a full-line comment
-key = "value" # This is a comment at the end of a line
+key1 = "value" # This is a comment at the end of a line
+key2 = "#this is not a comment"
 ```
 
 Key/Value Pair


TheElectronWill:
See issue #566

rougeth:
Hey @TheElectronWill, thanks for the reference. I disagree with the conclusion on that issue, don't think it compares to tables or `'='` inside Strings. Docs says that `'#'` marks the rest of the line as a comment, so `key = "# should be a valid comment"`.

My point here is: it should be explicit when the rules described doesn't apply.
Sorry for not checking that issue before open the PR.

TheElectronWill:
I like the way you've explained it in a simple yet clear manner. Moreover, you're not the first one to ask a clarification about this topic so I think it's worth it.

eksortso:
Leveling a contradictory specification in short order. Nice work! I'm actually a little jealous.

Anything that makes the README more authoritative, in both fact and tone, is worth accepting. The ABNF is a good supplement but it's not as authoritative on a human level as the README is.

pradyunsg:
Thanks @rougeth for the PR and everyone else for the discussion! :)

PR #602, opened 2019-03-10
Unify terminology translation of inline table, restore terminology translation about number

LongTengDao:
1. Make the spec 0.5 use the same translation with spec 0.4 and change log (#600 ), `inline table` should not be translated to `table in line`-like Chinese.
2. Use English directly, because `signed long` (which was translated to `sign as long integer type`-like thing) and `binary64` (which was translated to `64 bit double precision`-like Chinese, but `binary64` together equals to `double precision`) is computer terminology.


pradyunsg:
Lemme know if this is ready.

chenrui333:
@pradyunsg yeah, it is ready and good!

PR #614, opened 2019-04-28
Clarify that leading `+` is not allowed for x/o/b

LongTengDao:
According to `.abnf` and reason. (Although I don't understand why hex/oct/bin could not used for negative.)


pradyunsg:
Thanks @LongTengDao!

PR #619, opened 2019-05-11
Do not allow surrogate code points in ABNF

sevmeyer:
This is more a suggestion than a request.

I noticed that the current ABNF grammar allows [UTF-16 surrogate code points](https://unicode.org/glossary/#surrogate_code_point) (0xD800 - 0xDFFF) in strings and comments. This seems inconsistent, because it is invalid to insert these code points via escape sequences (e.g. \uD800).

This commit excludes the surrogate range. Because this affects five separate rules, I added a reusable rule `non-ascii`, to reduce redundancy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* None.
+* Disallow UTF-16 surrogate code points in grammar (U+D800 - U+DFFF).
 
 ## 0.5.0 / 2018-07-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-* Disallow UTF-16 surrogate code points in grammar (U+D800 - U+DFFF).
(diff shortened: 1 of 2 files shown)


pradyunsg:
Thanks for finding this and filing this PR (and for your patience!) @sevmeyer.

I'm split on whether we should include a note on this in the specification. We should however add a note to the changelog. Would you be willing to update your PR to do that?

sevmeyer:
I have added an entry to CHANGELOG.md, if that is alright.

I don't think the specification needs any changes.
It already excludes the surrogate range with the statement:

> A TOML file must be a valid UTF-8 encoded Unicode document.

According to Unicode 12.0, D92:

> Because surrogate code points are not Unicode scalar values, any UTF-8 byte sequence that would otherwise map to code points U+D800..U+DFFF is ill-formed.

pradyunsg:
Thanks for filing this @sevmeyer, for your patience and for the super quick response!

I tweaked the wording a bit before merging. :)

PR #621, opened 2019-05-17
Add editorconfig suggestion

LongTengDao:


pradyunsg:
I don't think this needs to be documented in the language standard. This is general good-practice for software development and belongs in a different place.

pradyunsg:
Thanks for the suggestion @LongTengDao!

PR #624, opened 2019-05-31
close #616

### LINKED ISSUE #616: Add an example about dotted numbers (3.14) as a dotted key
> Add a small note to the dotted keys section that `1.0 = true` gets parsed into `{"1": {"0": true"}}`.

_Originally posted by @pradyunsg in https://github.com/toml-lang/toml/issues/592#issuecomment-489573215_

yjqiang:
Add an example about dotted numbers as a dotted key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,8 @@ name = "Orange"
 physical.color = "orange"
 physical.shape = "round"
 site."google.com" = true
+3.141592 = "pi"
+"3.141592" = "pi"
 ```
 
 In JSON land, that would give you the following structure:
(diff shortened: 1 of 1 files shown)


pradyunsg:
Thanks for filing this PR @yjqiang!

Instead of adding to the existing code snippet, could you add a separate example, with the following language (or equivalent language):

> Since bare keys are allowed to compose of only ASCII integers, it is possible to write dotted keys that look like floats but are 2-part dotted keys. Don't do this unless you have a good reason to (you probably don't).

I'd also suggest sticking with the same structure of example with resulting JSON for that example.

yjqiang:
@pradyunsg So you think it is not obvious enough? Maybe add new one named "Pitfall in TOML" in https://github.com/toml-lang/toml#table-of-contents?

pradyunsg:
> So you think it is not obvious enough?

While this PR as it stands does cover the required functionality, I also want us to call it out as not-a-good-idea.

I filed #625 for this; I'd appreciate if you could 👀 at that and lemme know what you think!  :)

PR #625, opened 2019-06-03
Add cautionary note regarding floats as dotted keys

### LINKED ISSUE #616: Add an example about dotted numbers (3.14) as a dotted key

pradyunsg:
Closes #616 
Closes #624

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,20 @@ name = "Tom"
 name = "Pradyun"
 ```
 
+Since bare keys are allowed to compose of only ASCII integers, it is possible
+to write dotted keys that look like floats but are 2-part dotted keys. Don't do
+this unless you have a good reason to (you probably don't).
+
+```toml
+3.14159 = "pi"
+```
+
+The above TOML maps to the following JSON.
+
+```json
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -201,6 +201,20 @@ name = "Tom"
 name = "Pradyun"
 ```
 
+Since bare keys are allowed to compose of only ASCII integers, it is possible
+to write dotted keys that look like floats but are 2-part dotted keys. Don't do
+this unless you have a good reason to (you probably don't).
+
+```toml
+3.14159 = "pi"
```
yjqiang: How about add `"3.141592" = "pi"`?  By contrast, we can make it clearer.

pradyunsg: I wanted to make it clear what value came from what key. :)

yjqiang: 👌

pradyunsg:
Thanks for taking a look @yjqiang! :D

PR #626, opened 2019-06-09
Update example for inline tables to not use dotted key

luonghuutinh:
TOML type.name = "pug" should be  type = { name = "pug"} in JSON


pradyunsg:
This is intended to demonstrate that dotted keys can be used inside inline tables.

luonghuutinh:
I got it. That is my mistake when thinking an inline table is JSON syntax.

Issue #571, opened 2018-10-06
Unclear whether raw tab is allowed inside multi-line basic string


jorisvr:
It is unclear from the specification whether raw tab characters are allowed inside multi-line basic strings.

On the one hand, the section **Multi-line basic strings** states that "_All other whitespace and newline characters remain intact._". This suggests that _all_ whitespace is allowed inside the string. And whitespace has already been defined as tabs and spaces.

On the other hand, the same section states that "_Any Unicode character may be used except those that must be escaped ... (U+0000 to U+001F, U+007F)_".

So which is it ? Allowed because whitespace, or forbidden because control character ?

The ABNF definition forbids raw tab characters inside multi-line basic strings, but the ABNF is currently not authoritative.

pradyunsg:
Nice catch @jorisvr! :)

I think we should clarify that not "all whitespace" is allowed unescaped.

LongTengDao:
**Basic strings** is designed to input chars which can't easy input, like control chars, use escape mark ( these marks created in computer epoch ).
**Literal strings** is designed to input the escape mark self, or content completely has no relationship with coding on computer.

So there is no difference between **basic strings** and **literal strings** about treating `Tab` -- `Tab` is a human visible charactor, like `Space` or `ABC...`. `CR` and `LF` is same to `TAB` on this, but used for file lines split, it's special only so. So the control-chars I think even should not be treated special. Because it's no reason and useless -- Unicode defined so mush control-type chars.

I think the spec should yaoqiu the editors default treat the whitespace chars in `.toml` file like:
```
# http://editorconfig.org
[*.toml]
trim_trailing_whitespace = false
```
to avoid multi string written in config file be auto changed accidentally.

`TOML` is a config format, not a transfer format, I think it could ignore many popular limit, to be as simple as possible.

eksortso:
Even though it's fairly easy to type a tab, it seems that these whitespace characters were overlooked in the definition of multi-line basic strings, and in fact needlessly forbidden in single-line basic strings.

Let's let tabs be free in all strings. The escape code `\t` is still alright in both types of basic strings.

pradyunsg:
> Let's let tabs be free in all strings.

Done. :)

PR #627, opened 2019-06-10
Re #571, allow raw tabs in basic and MLB strings

eksortso:
In response to #571, these changes permit the horizontal tab character to be used in both basic strings and multi-line basic strings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,9 +314,10 @@ str3 = """\
 ```
 
 Any Unicode character may be used except those that must be escaped: backslash
-and the control characters other than tab (U+0000 to U+0008, U+000A to U+001F,
-U+007F). Quotation marks need not be escaped unless their presence would create
-a premature closing delimiter.
+and the control characters other than tab, line feed, and carriage return
+(U+0000 to U+0008, U+000B, U+000C, U+000E to U+001F, U+007F). Quotation marks
+need not be escaped unless their presence would create a premature closing
+delimiter.
 
 If you're a frequent specifier of Windows paths or regular expressions, then
 having to escape backslashes quickly becomes tedious and error prone. To help,
--- a/README.md
+++ b/README.md
@@ -239,7 +239,7 @@ multi-line literal. All strings must contain only valid UTF-8 characters.
 
 **Basic strings** are surrounded by quotation marks. Any Unicode character may
 be used except those that must be escaped: quotation mark, backslash, and the
-control characters (U+0000 to U+001F, U+007F).
+control characters other than tab (U+0000 to U+0008, U+000A to U+001F, U+007F).
 
 ```toml
 str = "I'm a string. \"You can quote me\". Name\tJos\u00E9\nLocation\tSF."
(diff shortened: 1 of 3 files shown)


eksortso:
This may take some investigation in case some parsers are doing this already, but perhaps this feature could be deferred until after the release of TOML 1.0?

pradyunsg:
Yea, I don't see this as being super critical for 1.0 tbh. It's probably a good idea but be super conservative logic applies here.

pradyunsg:
BTW, thanks for the ping on this; I'd completely forgotten about this PR.

ChristianSi:
I oppose the delay. The problem stated in  #571 is that the TOML spec is currently contradictory. The written spec says:

> All other whitespace and newline characters remain intact.

Which, according to the common understanding of whitespace, allows tabs. But another sentence in the spec forbids tabs, and the ABNF forbids them as well. TOML 1.0 **should** not be ambiguous, hence this PR must be merged (or the ambiguity must be resolved in another way) before TOML 1.0 can be released.

README.md
```diff
@@ -314,8 +314,9 @@ str3 = """\
 ```
 
 Any Unicode character may be used except those that must be escaped: backslash
-and the control characters (U+0000 to U+001F, U+007F). Quotation marks need not
-be escaped unless their presence would create a premature closing delimiter.
+and the control characters other than tab (U+0000 to U+0008, U+000A to U+001F,
+U+007F). Quotation marks need not be escaped unless their presence would create
```
ChristianSi: This still forbids line breaks, while the whole point of multi-line strings is to allow them. This is not a problem of the PR, but of the original spec. The correct wording, I believe, would be:

> ... the control characters other than tab, line feed, and carriage return (U+0000 to U+0008, U+000B,  U+000C,
> U+000E to U+001F, U+007F)....

ChristianSi:
Also I just noted that the written spec currently seems to forbid line breaks within multi-line basic strings (they are within the forbidden range), which is of course silly. I added an inline comment to the PR on how this can be resolved.

This makes it all the more urgent to fix this (properly) before 1.0 is released.

toml.abnf
```diff
@@ -64,7 +64,7 @@ basic-string = quotation-mark *basic-char quotation-mark
 quotation-mark = %x22            ; "
 
 basic-char = basic-unescaped / escaped
-basic-unescaped = %x20-21 / %x23-5B / %x5D-7E / non-ascii
+basic-unescaped = %x09 / %x20-21 / %x23-5B / %x5D-7E / non-ascii
```
pradyunsg:
```suggestion
basic-unescaped = wschar / %x21 / %x23-5B / %x5D-7E / non-ascii
```

toml.abnf
```diff
@@ -86,7 +86,7 @@ ml-basic-string-delim = 3quotation-mark
 
 ml-basic-body = *( ml-basic-char / newline / ( escape ws newline ) )
 ml-basic-char = ml-basic-unescaped / escaped
-ml-basic-unescaped = %x20-5B / %x5D-7E / non-ascii
+ml-basic-unescaped = %x09 / %x20-5B / %x5D-7E / non-ascii
```
pradyunsg:
```suggestion
ml-basic-unescaped = wschar / %x21-5B / %x5D-7E / non-ascii
```

pradyunsg:
@ChristianSi consider me convinced.

toml.abnf
ChristianSi: Looks good to me.

README.md
eksortso: Good catch, @ChristianSi. I'm adding this now.

I'm still pretty green with these tools. Commit [9e46984](https://github.com/toml-lang/toml/pull/627/commits/9e469844938e64ed178732f8c4c74ca04c4ef2ed) addresses this. I used your suggestion verbatim. Can you verify it?

eksortso:
> I oppose the delay. The problem stated in #571 is that the TOML spec is currently contradictory.

Thanks for reminding me. I'd forgotten that!

PR #634, opened 2019-06-21
Typo in README

jiegillet:
`0x0D0A` is the Malayalam Symbol `ഊ` while `0x0D` is `\r`, so I'm calling it a typo.


pradyunsg:
Ah nice catch -- that's intended to mean `\n` or `\r\n`.

jiegillet:
Mmh, I'm not sure now, I changed it to `0x0D 0x0A` but maybe just `\n` or `\r\n` like you suggested is better? Or both?

pradyunsg:
Thanks @jiegillet! ✨

Issue #638, opened 2019-06-29
Unclear whether table [a] can be defined after [a.b] has been defined


mmakaay:
When implementing my parser, the BurntSushi parser taught me that the following is valid:
```
[a.b]
some="key"
[a]
lala="lalala"
```
The part of the spec related to this is:
```
You don't need to specify all the super-tables if you don't want to.
TOML knows how to do it for you.

# [x] you
# [x.y] don't
# [x.y.z] need these
[x.y.z.w] # for this to work
...
Like keys, you cannot define any table more than once. Doing so is invalid.
```
How I interpreted this, was that `[x.y.z.w]` would be equivalent to 
```
[x]
[x.y]
[x.y.z]
[x.y.z.w]
```
and that therefore, in the following definition, the definition of `[x]` would be invalid, since under the hood that table was already created:
```
[x.y.z.w]
[x]
```

I updated my code to allow this definition now, by keeping track whether or not a table was created implicitly or explicitly. Table `[x]` as created by `[x.y.z.w]` is implicit and I upgrade it to an explicit table based on the next definition of `[x]`.

Is this indeed the intended behavior of TOML?
If yes, then the specification could do with a clarification on that. 
Two directions that I see for this:

1) use the implicit / explicit terminology. E.g. "You don't need to specify all the super-tables if you don't want to. TOML will implicitly create the intermediate tables for you. [..] Like keys, you cannot define any explicitly defined table more than once. Doing so is invalid."

2) focus on key being used for defining the table. E.g. "Like keys, you cannot define a table for a given key more than once. Doing so is invalid."

Adding an example would of course explain things beyond any doubt. It could be a little extension of the existing explanatory example:
```
# [x] you
# [x.y] don't
# [x.y.z] need these
[x.y.z.w] # for this to work

[x] # you can still define this one afterwards
```

ChristianSi:
Yes, that's indeed the intended behavior. An "implicit table" is still not, technically, "defined" (at least not by you), hence explicitly defining it later does not violate the "you cannot define any table more than once" rule.

It might be useful to add the sentence such as "Tables may be defined in any order" to the spec, followed by an example such as the one you gave:

````toml
[a.b]
some="key"
[a]
lala="lalala"
````

mmakaay:
Yes, that would have made it crystal clear to me.

pradyunsg:
Happy to add a small clarification for this.

mmakaay:
A pull request is created.
[...]

PR #641, opened 2019-07-03
Clarify the use of [a] after [a.b]

mmakaay:
It wasn't fully clear from the specs whether or not it is allowed to
create a table [a] explicitly when earlier on in the TOML document the
table [a.b] was created, since that expression creates [a] implicitly.

See: https://github.com/toml-lang/toml/issues/638

### DIFF
--- a/README.md
+++ b/README.md
@@ -626,6 +626,8 @@ how to do it for you.
 # [x.y] don't
 # [x.y.z] need these
 [x.y.z.w] # for this to work
+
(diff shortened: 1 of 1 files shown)


pradyunsg:
I'm going to go ahead and merge this as is; I don't think we need to add any more clarification text here since the fact that you can't define something again is implicit/understood from other language in the specification + seems pretty intuitive to me. :)

pradyunsg:
Thanks @mmakaay! :D

Issue #637, opened 2019-06-27
Unclear whether multiple key/value pairs can be on a single line


mmakaay:
When testing my implementation of TOML 0.5.0 against the BurntSushi test set, one test that failed whas the test where there are multiple key/value pairs on a single line. My parser accepted this and during development I did interpret the spec as such consciously.

The spec says "The key, equals sign, and value must be on the same line"
That tells me that this group of three elements must not have inner newlines.
It doesn't say that after a completed key/value pair, there cannot be another pair.

So, based on the spec, I implemented this as valid:
```
key1="value1" key2="value2"
```

If you look at the description of a [table] element, then there's the following text in the spec: "They appear in square brackets on a line by themselves. ". Because of that clear wording, I derived that the same thing didn't apply to key/value pairs.

Since I only seem to find tests that reject this syntax, and examples show key/value pairs on their own line, I went for it and updated my parser to expect an end of line after the value assignment. 

Should be spec be updated to be more clear on this?

pradyunsg:
Thanks for filing this!

It's not valid. You can only have one key-value pair in a line (unless in an inline table).

> Should be spec be updated to be more clear on this?

Do you have a concrete suggestion on what language we should add?

eksortso:
May I suggest the following, or something like it? This example would explain the concept well, after the first few examples in the spec show the key/value pairs on separate lines.

---
Key/value pairs are separated with one or more newlines. Running pairs together on the same line is not permitted. (See [Inline Table](https://github.com/toml-lang/toml#inline-table) for exceptions.)
```toml
first = "Tom" last = "Preston-Werner"  # INVALID
```

ChristianSi:
It's true that this isn't yet explicitly mentioned in the spec -- something that should clearly be corrected.

mmakaay:
My suggestion would be to use the same kind of wording as used to explain that a [table] must be on its own on a line. So maybe a little extension of the wording would do: "The key, equals sign, and value must be on the same line by themselves".

I think eksortso's suggestion is better than mine would have been. The example makes things crystal clear.

pradyunsg:
@eksortso's suggestion LGTM. Happy to merge a PR, in case someone gets to it before me. :)

mmakaay:
A pull request is created.
I tried to keep the changes to the spec as minimal as possible.

mmakaay:
The pull request was merged, so I'm closing the issue.

PR #642, opened 2019-07-03
Clarify that multiple key/vals on 1 line are invalid

mmakaay:
It wasn't fully clear from the specs whether or not it is allowed
to have multiple Key/Value pairs together on 1 line of a TOML doc.

See: https://github.com/toml-lang/toml/issues/637

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,13 @@ Datetime, Array, or Inline Table. Unspecified values are invalid.
 key = # INVALID
 ```
 
+There must be a newline after a key/value pair.
+(See [Inline Table](#user-content-inline-table) for exceptions.)
+
+```
+first = "Tom" last = "Preston-Werner" # INVALID
+```
+
 Keys
 ----
 


pradyunsg:
You've included changes from #641 in this. :)

mmakaay:
Yeah, I accidentally branched from the #641 branch. Too much sleep in my eyes I guess ;-)
But I did push another commit with a squashed fix for it.
Thanks for the merge.

pradyunsg:
Thanks for the PR @mmakaay! :)

ChristianSi:
Is it legal to have a value right at the end of the file, with no trailing newline?

Such things are rare under Linux, were editors tend to auto-append a trailing newline, but quite common under Windows. The merged pull requests forbids them, but that's probably unintentional?

It would be better to write something like "There must be a newline (or the end of the file) after a key/value pair."

mmakaay:
Acorrding to the ABNF production rules, the file can end with a value that has no newline after it:
```ABNF
toml = expression *( newline expression )

expression =  ws [ comment ]
expression =/ ws keyval ws [ comment ]
expression =/ ws table ws [ comment ]
```
Newlines are between expressions. 
The fact that a TOML file can have a trailing newline is defined by the fact that `ws [ comment ]` is a valid expression, also allowing for an empty expression prefixed by a newline.

So yes, I'd say that stating the end of the file as a possible terminator would be correct.

Issue #446, opened 2017-02-03
Possible positional restrictions on table definitions


alexcrichton:
Hello! I'm currently in the process of rewriting a TOML parser and have come
across a curious aspect of TOML. I believe that a document such as this is
valid:

```toml
[a]
key = 1

[unrelated-table]
key = 1

[a.b]
key = 1
```

That is, you can define a table across many portions of a document. There's
certainly nothing wrong with this per-se, but it means that a parser decoding
this document is forced to keep all intermediate tables in-memory or in some
side data structure. For example if you'd like to decode the table `a` from
this document you're forced to parse `unrelated-table` and store it in memory
elsewhere to come back to later (if requested).

This is definitely a *super* niche concern, of course! TOML is intended for
configuration, which should never be something like gigabytes large. On the
other hand though this does make writing a parser a little bit more difficult
in some situations, and to me at least I'd subjectively find such a definition
relatively wonky.

I wonder, what would others think about adding a clause such as this to the
spec?

> Although one table may be defined across multiple sections (if its keys are
> themselves tables) all sections that define a table must be contiguous. For
> example this is invalid:
>
> ```toml
> [a]
> key = 1
>
> [unrelated-table]
> key = 1
>
> [a.b]
> key = 1
> ```
>
> Similarly, this is also invalid
>
> ```toml
> [a.sub-table]
> key = 1
>
> [a.unrelated-table]
> key = 1
>
> [a.sub-table.b]
> key = 1
> ```

This does make the otherwise very simple spec somewhat more complicated, and also somewhat more arbitrarily. If this were *only* about parser performance I don't think I'd bother with the suggestion, but subjectively interpreting this as odd prompted me to file this report!

BurntSushi:
I think probably the biggest concern I'd have with this is that it's a breaking change. It could make an otherwise valid TOML file invalid. :-(

alexcrichton:
Ah yes good point, I forgot to mention that! 

I wasn't quite sure whether the 0.4 -> 1.0 transition (assuming that happens) is just a clarification as opposed to a breaking release. If no breaking changes are being made then I think the ship has sailed!

lmna:
I believe that suggested change would make TOML more obvious and friendly to a human reader, and more consistent as well.

[...]

PR #646, opened 2019-08-06
Add advice to not define dotted keys out-of-order

### LINKED ISSUE #632: Add advice to not define dotted keys out-of-order
This should be after the example of ASCII-float keys in "Keys" section.

LongTengDao:
Closes #632

TOML 1.0 +1

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,10 +241,28 @@ a.b.c = 2
 Defining dotted keys out-of-order is valid but discouraged:
 
 ```
-o.1.1 = 1.1
-o.2.1 = 2.1
-o.1.2 = 1.2
-o.2.2 = 2.2
+# VALID BUT DISCOURAGED:
+
+a.type = ''
+b.type = ''
+
+a.name = ''
+b.name = ''
+
+a.data = ''
+b.data = ''
+```
+
+```
+# RECOMMENDED:
+
+a.type = ''
+a.name = ''
+a.data = ''
+
+b.type = ''
+b.name = ''
+b.data = ''
 ```
 
 String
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -238,6 +238,15 @@ a.b = 1
 a.b.c = 2
 ```
 
+Defining dotted keys out-of-order is valid but discouraged:
+
+```
+o.1.1 = 1.1
+o.2.1 = 2.1
+o.1.2 = 1.2
+o.2.2 = 2.2
```
workingjubilee: This IS totally valid `.toml`, but is a little odd because it goes against the admonishment to not do things that look like "float keys" (multiple keys by integers). Then again it's an example for "what not to do, unless you're really sure", so... appropriate?

LongTengDao:
Updated after thinking. Thanks for asking.

rwitzel:
:+1: Nice.

README.md
```diff
@@ -238,6 +238,33 @@ a.b = 1
 a.b.c = 2
 ```
 
+Defining dotted keys out-of-order is valid currenctly, but discouraged:
```
ChristianSi: Maybe just "Defining dotted keys out-of-order is discouraged:" – That's it's valid is also pointed out at the top of the example, and "currenctly" (actually there's a typo in there) suggests that it will become invalid in the future – which might nor might not happen.

README.md
```diff
@@ -663,6 +690,22 @@ b = 1
 c = 2
 ```
 
+And defining table out-of-order is valid currenctly but discouraged too:
```
ChristianSi: I would again delete "valid currenctly but ", see above; "table" should be "tables".

ChristianSi:
I made some review comments, looks good otherwise :+1:

README.md
LongTengDao: I wrote "currently" because:
https://github.com/toml-lang/toml/issues/446#issuecomment-345363045
https://github.com/toml-lang/toml/issues/446#issuecomment-391778942

But I forgot this example, which is reasonable:
https://github.com/toml-lang/toml/issues/446#issuecomment-391861244

So I removed the "currently".

pradyunsg:
Thanks for filing this @LongTengDao and for reviewing this @workingjubilee and @ChristianSi! ^>^

PR #647, opened 2019-08-14
Clarify: no dotted keys or subtables into inline tables

### LINKED ISSUE #630: Clarify inline tables are immutable (and dotted keys can't "inject" into them)
You can't "inject" into inline tables with dotted keys. This isn't clear in the specification and we should clarify.

eksortso:
This intends to resolve #630, i.e. it clarifies that inline tables are immutable once defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -701,7 +701,7 @@ type.name = "pug"
 
 ```
 
-Once inline tables are defined, their values are fully set, and new elements and
+Once inline tables are defined, their values are fully set, and new keys and
 subtables cannot be added to them.
 
 ```toml
--- a/README.md
+++ b/README.md
@@ -701,6 +701,15 @@ type.name = "pug"
 
 ```
 
+Once inline tables are defined, their values are fully set, and new elements and
+subtables cannot be added to them.
+
+```toml
+a = {}
+#a.b = 1  # INVALID
+#[a.b]  # INVALID
+```
+
 Array of Tables
 ---------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Clarify that inline tables cannot have keys or subtables injected into them.
 * Clarify in ABNF that UTF-16 surrogate code points (U+D800 - U+DFFF) are not
   allowed in strings or comments.
 * Allow raw tab characters in basic strings and multi-line basic strings.


README.md
```diff
@@ -701,6 +701,15 @@ type.name = "pug"
 
 ```
 
+Once inline tables are defined, their values are fully set, and new elements and
```
workingjubilee: "Element" only occurs in reference to Arrays in TOML, but I don't think you intended to reference Arrays, so please use "keys" instead of "elements"?

eksortso: I get so sick of saying "Key/Value Pair" inside of tables. That must be why I said "element" instead. Since this is a clarification update and doesn't slog down the wording (in fact, it tightens it up a bit), I'll gladly make the change that you suggest.

I could support a lawyerly rewrite of the spec so that all terms are rigidly defined. I don't think it was originally written that way, though, and I'm a little afraid that the strident tone of the original might get lost in such an effort. But TOML is catching eyes because it's well-defined. So this is a step in the right direction.

ChristianSi:
A problem I see here is that the text doesn't address the reverse case. It the following allowed?

```toml
a.b = 1  # INVALID
a = {c = 2}
```

Obviously, it isn't, but the text doesn't make this clear. Indeed, the wording "Once inline tables..." might suggest that dotted keys can "inject" valued into inline tables as long as the inline table itself is defined *later* in the document.

eksortso:
That's easy to resolve. How about if we add this?

----
If a table is defined with dotted-key expressions, it cannot be redefined with
an inline table.

```toml
a.b = 1       # The table `a` is defined here.
#a = {c = 2}  # INVALID: This table is defined with dotted keys.
a.c = 2       # This is okay
```

ChristianSi:
@eksortso That reads better, however, the other case should be mentioned too:

```toml
#[a.b]  # INVALID, since a is an immutable inline table (see below)
a = {}
```

workingjubilee:
Perhaps something like "inline tables by definition must contain all their key/value pairs inline, including sub-tables, or else they would not be inline tables, and so, are invalid" would be to-the-point.

mmakaay:
> That's easy to resolve. How about if we add this?
> 
> If a table is defined with dotted-key expressions, it cannot be redefined with
> an inline table.
> 
> ```toml
> a.b = 1       # The table `a` is defined here.
> #a = {c = 2}  # INVALID: This table is defined with dotted keys.
> a.c = 2       # This is okay
> ```

That's not really a good distinction IMO, since the `a.b = 1` could also be defined without dotted keys, while still the `a = {c = 2}` case is invalid.

```
a = {b = 1}
a = {c = 2}  # INVALID
```

The point is that you cannot create a table more than once, regardless how you define it in the TOML document. In the modified example that I provide above, by rewriting `a.b = 1` this is especially clear, since now you see two assignment for `a = ...`.

I have struggled with this a bit too, when implementing my TOML parser, and what I ended up with was that internally I keep track of implicit and explicit tables. An implicit table is a table which is only used as a "path" in a dotted key syntax. For example:

```
a.b.c = 1234
```

Here, I have to create a table "a" and a table "b". Table "a" is flagged implicit, table "b" is explicit.
The rule that I follow, is that I can create an explicit table when there's not yet an explicit table at a given path. For example:

```
a.x = { very = "good" }  # a.x did not exist at all, creating it is OK
a = { good = "as well" } # a was implicit up to now, creating it is OK and makes it explicit
a.b.field = "another key value pair" # standard table, so adding a key/value pair is OK
a.b = { hmmm = "do not do that" } # a.b already exists as an explicit table, INVALID 
```

I'm not saying that "implicit" versus "explicit" should be wording to be used in the spec. I'm just trying to help getting clear what constructions are allowed and what are not allowed.

pradyunsg:
Thanks for getting the ball rolling on this @eksortso, and for your inputs here @ChristianSi and @workingjubilee!

I've posted what I had locally at #652 from a few weeks back. If no one has any concerns with that, let's merge that. If there are concerns, please voice them here and I'll close the other PR -- let's avoid splitting the discussion. :)

eksortso:
> Perhaps something like "inline tables by definition must contain all their key/value pairs inline, including sub-tables, or else they would not be inline tables, and so, are invalid" would be to-the-point.

Sorry, but I honestly don't know how to react to that.

eksortso:
You got it down pat with #652, @pradyunsg, thank you. I'm closing this PR.

workingjubilee:
😅 Sorry, was an unfortunately half-baked thought.

PR #649, opened 2019-08-16
Clarify how dotted keys interact with ordinary table syntax

### LINKED ISSUE #631: Clarify that a table defined by dotted keys can not be overridden

workingjubilee:
This is intended to close #631, or at least to hopefully further spur the discussion towards a final conclusion. It includes an example of how dotkey syntax allows some exceedingly flexible definitions of tables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Clarify how dotted keys interact with ordinary table syntax.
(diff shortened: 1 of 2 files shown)


ChristianSi:
I **strongly** object to this. "You cannot define any table more than once" has been a part of the spec since a long time. In this example, the table "a" is effectively defined twice and the TOML parser is supposedly forced to merge both definitions. This is not what TOML was meant to be.

workingjubilee:
‾\_(ツ)_/‾

PR #652, opened 2019-08-19
Immutable Inline tables

### LINKED ISSUE #630: Clarify inline tables are immutable (and dotted keys can't "inject" into them)

pradyunsg:
Closes #630 
(maybe?) Supercedes and closes #647.

---

This was a draft I had locally that I just updated to use "cannot be added to them" borrowed from #647.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Clarify that inline tables are immutable.
 * Clarify in ABNF that UTF-16 surrogate code points (U+D800 - U+DFFF) are not
   allowed in strings or comments.
 * Allow raw tab characters in basic strings and multi-line basic strings.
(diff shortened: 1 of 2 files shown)


eksortso:
Definitely supercedes and closes #647. A much more engaging example, too. I'd suggest some tidy-up changes, but all tidy-up changes can wait.

pradyunsg:
Merged since @eksortso has closed #647 and OK'ed this PR. :)

pradyunsg:
> I'd suggest some tidy-up changes, but all tidy-up changes can wait.

Oh btw, please do suggest these. Even if they can wait, it'll be helpful to note them down when you think of them. That way, you can forget about it. :)

PR #654, opened 2019-08-20
Describe dot-key implicit table definition.

### LINKED ISSUE #631: Clarify that a table defined by dotted keys can not be overridden

workingjubilee:
Adds additional notes to the description of dotted keys, to hint where the reader should look for the full explanation about how dot-keys implying tables, which itself also arrives with this change, and lives in the Tables section. Intended to resolve #631, except for real this time. Honest!

As I was thinking through it, I thought of a particularly notable edge case (to be clear, it's omnipresent as a concern due to the spec, this variant just stood out to me), which led to the particular example I chose and the included caution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Clarify that dot-keys define tables and when they stop accepting new keys.
 * Clarify that inline tables are immutable.
 * Clarify in ABNF that UTF-16 surrogate code points (U+D800 - U+DFFF) are not
   allowed in strings or comments.
--- a/README.md
+++ b/README.md
@@ -174,8 +174,9 @@ discouraged).
 '' = 'blank'     # VALID but discouraged
 ```
 
-**Dotted keys** are a sequence of bare or quoted keys joined with a dot. This
-allows for grouping similar properties together:
+**Dotted keys** are sequences of bare or quoted keys joined with a dot, which
+imply a table for each key except the final key. This allows for grouping
+similar properties together:
 
 ```toml
 name = "Orange"
@@ -670,7 +671,7 @@ Empty tables are allowed and simply have no key/value pairs within them.
 
 Like keys, you cannot define any table more than once. Doing so is invalid.
 
-```
+```toml
 # DO NOT DO THIS
 
 [a]
@@ -680,7 +681,7 @@ b = 1
 c = 2
 ```
 
-```
+```toml
 # DO NOT DO THIS EITHER
 
 [a]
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -706,6 +707,40 @@ Similarly, defining tables out-of-order is discouraged as well.
 [b]
 ```
 
+Likewise, a table implied by a dot-key chain finishes being defined at the
+same moment as the table it is inside. After that, it cannot be redefined.
+
+```toml
+[public-domain]
+authors.'Mary Shelley'."Frankenstein; or, the Modern Prometheus" = true
```
pradyunsg: Hmm... These examples feel *very* involved and exhibit multiple concepts in addition to what the example is for (unicode, case sensitivity and quoted strings vs implicit tables from dotted keys). I'd prefer that we simplify these to only show the interactions of implicit tables.

I'd suggest using something like `subject.math.class-size` or something similarly trivial, as the thing to design the examples around, instead of `authors.'紫式部'."The Tale of Genji"`.

An involved example might also make things more complicated for translators though I don't recall if any translation has modified the example text.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Clarify that dot-keys define tables and when they stop accepting new keys.
```
pradyunsg: Clarify that dotted keys implicitly define tables.
Clarify behavior of tables defined implicitly by dotted keys.

README.md
```diff
@@ -670,7 +671,7 @@ Empty tables are allowed and simply have no key/value pairs within them.
 
 Like keys, you cannot define any table more than once. Doing so is invalid.
 
-```
+```toml
```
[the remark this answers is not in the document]
workingjubilee: Oh OK.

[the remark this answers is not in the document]
pradyunsg: Yea, it's helpful.

However, GitHub/linguist occasionally need an update for that to work properly and that needs to wait for a new release at times.

If we're going to do this, let's do it in a separate PR. :)

README.md
workingjubilee: OK, I can do a second pass on trimming things down.

re: translation ease, I actually have worked on localization projects before and proper names are actually a boon, precisely because there is no conceptual space in the target language "for" that thing. Rather, it is a proper name, so it should not be semantically modified. Instead, if you want to make things more convenient, you can offer a phonetic remapping into the target language, maybe switch around a name order, and call it "done", without having to make any too-hard decisions like "how do I translate these two differing conceptions of what 'size' even is?" If they're of actual people, then probably someone has already done the work and created a canonical example, and you can end with a simple copy-and-paste job (translating names of non-real people is more complex because Reasons).

This doesn't mean there aren't other good reasons to not go with the examples I contrived! Far from it. I just wanted to explain that detail, as long as we're talking about i18n.

pradyunsg: Here's my shot at this example:

```toml
[fruit]
apple.color = "red"
apple.taste.sweet = true

# [fruit.apple]  # INVALID
# [fruit.apple.taste]  # INVALID

[fruit.apple.texture]  # you can add sub-tables to implicitly defined tables
smooth = true
```

pradyunsg: Hmm... Lemme make an equivalent PR, since I think I can budget the time now and procrastinate on studying.

pradyunsg:
@workingjubilee Could you look at #655 and provide feedback on the same? :)

ChristianSi:
@workingjubilee Truth be told, @pradyunsg had already said 5 days ago "Lemme make an equivalent PR", so you had ample warning.

pradyunsg:
Sorry. I understand that you didn't like that my position seemed to change from "request changes" to "I'll do this instead".

Part of the point of me making the PR and asking you for feedback immediately, was to avoid duplicated effort -- the thought was to show the example that I had in mind and get your inputs on it, in a better manner than a PR comment, since well, I figured that's not the most effective way to discuss this. I waited around a couple of days since that's, basically, the communication cadence on this project when I'm actually working on this.

I'm still not sure how that wasn't optimal, so I think you had different [expectations][1] here. If you could clarify on what your expectations were, that'd help avoid this in the future.

[1]: https://snarky.ca/setting-expectations-for-open-source-participation/

Anyway, I don't have much emotional energy lately, so apologies if anything above isn't worded in a polished manner.

pradyunsg:
> Sorry to bother!

Please don't be. :)

> I was definitely more plain confused than having a particular request, really, and trying to "be constructive" led to me not just making a post that was a series of question marks, on day of.

Please do feel free to ask questions, especially if anything I do seems to be out of place. I'm more than happy to be called out on my mistakes or be asked to justify the stuff I'm doing. :)

PR #655, opened 2019-08-21
Clarify behavior of tables defined implicitly by dotted keys

### LINKED ISSUE #631: Clarify that a table defined by dotted keys can not be overridden

pradyunsg:
Closes #631 
(maybe?) Supersedes and closes #654

### DIFF
--- a/README.md
+++ b/README.md
@@ -706,6 +706,21 @@ Similarly, defining tables out-of-order is discouraged as well.
 [b]
 ```
 
+Dotted keys imply the creation of tables. Once created, the `[table]` form can
+only be used to define sub-tables within such tables.
+
+```toml
+[fruit]
+apple.color = "red"
+apple.taste.sweet = true
+
+# [fruit.apple]  # INVALID
+# [fruit.apple.taste]  # INVALID
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -706,6 +706,21 @@ Similarly, defining tables out-of-order is discouraged as well.
 [b]
 ```
 
+Dotted keys imply the creation of tables. Once created, the `[table]` form can
+only be used to define sub-tables within such tables.
```
ChristianSi: These two sentence are correct, but their terseness makes them hard to understand. I would be somewhat more verbose, maybe as follows:

> By using dotted keys, you implicitly define everything to the left of each dot as a table. Since tables cannot be defined more than once, redefining such tables using a `[table]` header is not allowed. Using dotted keys to redefine tables already defined in `[table]` form is likewise not allowed. The `[table]` form can, however, be used to define sub-tables within tables defined in dotted-key form.

ChristianSi:
The example looks good, but I would prefer a more verbose explanation, see my inline comment.

pradyunsg:
Merging since I prefer this to #654. If anyone wants to provide inputs on this, we can discuss/address that in a follow up PR!

Issue #356, opened 2015-10-06
Are leading zeros allowed in the exponent part of a float?


avakar:
An issue have arisen in my parser: https://github.com/avakar/pytoml/issues/9, and I find specs ambiguous about this. In particular, it says

> An exponent part is an E (upper or lower case) followed by an integer part (which may be prefixed with a plus or minus sign).

Does the phrase _integer part_ mean it's an integer as specified in the Integer section of the specs and thus disallows leading zeros? In other words, is the following a valid TOML file?

```
maximum_error = 4.85e-06
```

Note the leading zero in the exponent.

ziotom78:
I am the initial reporter of the bug https://github.com/avakar/pytoml/issues/9. I would advocate the correctness of using leading zeroes, as this is common practice in a number of languages. (It allows nicely aligned numbers.)

Here are a few examples:

Python 2.7/3.4:
`````python
print(4.86e-6)
# Prints "4.86e-06"
`````

Ruby 2.1:
`````ruby
puts(4.86e-6)
# Prints 4.86e-06
`````

mojombo:
Can someone provide a concrete example where allowing leading zeros is useful?

ziotom78:
Here is my case. @avakar's TOML parser refused leading zeros because it failed to parse some of the TOML files that a Python script of mine was producing automatically. It turned out that such files were the ones where a parameter turned so small that scientific notation was used for it. (As I said above, Python's `print` automatically puts a leading zero to the exponent if this has just one digit.)

mojombo:
@ziotom78 Oh, I see, I didn't realize that's what your example was pointing out. That's quite curious that Ruby and Python do that. Do you know what the rationale is behind that behavior?

Hrxn:
I'd call that bad design...

Reminds me of that old thing, the C strcmp function and the implications it had for sorting...

ziotom78:
I imagine that this way of formatting numbers might allow nicer alignment when printing numbers in columns, though I am not really sure. Python says nothing about this: https://docs.python.org/2/library/string.html#formatspec. (See this question on StackOverflow for some more interesting information: http://stackoverflow.com/questions/9910972/python-number-of-digits-in-exponent.)

Interesting enough, there are cases where this behaviour is intentional and documented, see how .NET does floating-point formatting (https://msdn.microsoft.com/en-us/library/dwhawy9k.aspx#EFormatString):

[...]

PR #656, opened 2019-08-21
Permit leading zeroes in exponent parts of floats

eksortso:
Per #356, this PR allows for leading zeroes on the integer parts of exponents in floats.


ChristianSi:
Looks good to me.

mojombo:
Looks good. Thanks for the PR!

Issue #596, opened 2019-02-15
Are LocalDate, LocalTime, LocalDateTime and DateTime different types?


quadrupleslap:
**toml-rs** treats them as the same type (for the purpose of forbidding heterogeneous arrays), but that doesn't sound like the correct behavior.

quadrupleslap:
Also, why are they called Local* when their interpretation is actually undefined? Doesn't Naive* make more sense?

eksortso:
Good question! My first take was that, for the sake of making arrays heterogenous, all Datetime subtypes are the same. (Historically, there was only one Datetime type in TOML.)

Changing all the subtypes into distinct types makes sense to me. But my reasoning has more to do with type translations between TOML and programming languages, and less to do with preserving homogeneity in arrays.

ChristianSi:
I'd say it's pretty obvious that they are different types (and hence cannot be mixed as arrays), since they are listed as different types in the spec.

If Offset Date-Time and Local Date-Time are mixed in an array that's probably a mistake – either one should be used consistently. The only potential use case I see where mixing date types *might* be useful is a combination of Local Date-Time and Local Date with an (app-specific) convention such as "Local Date represents that day at (local) midnight" – or possibly Offset Date-Time and Local Date-Time with "Local Date represents that day at UTC midnight". But that's all a bit fuzzy and probably better avoided.

From a user's viewpoint, it would be more useful if Integers and Floats could be mixed (with ints optionally autoconverted to floats by parsers in strongly typed languages), so that series such as `[0, 0.5, 1, 1.5]` could be written in a natural fashion. But that's a different matter.

quadrupleslap:
@ChristianSi you're right, it looks like most parsers already treat the date types as separate types, and the spec does, too.

quadrupleslap:
Closing this issue because "it's pretty obvious that they are different types". Thanks @ChristianSi.

eksortso:
Just curious, @quadrupleslap: do you have a real-life requirement of homogenous arrays of date/time values?

Asking because of discussion surrounding arrays recently.

Issue #553, opened 2018-08-02
Mixed-type arrays still legal, even though the intent appears to have been to forbid them


rmunn:
Creating a new issue for this because the history of this discussion is spread among multiple old (and now-closed) issues, and it's not at all clear which one should be used to re-open the discussion.

The TOML spec currently (2018-08-02) says that data types may not be mixed inside an array, but if the array contains nested sub-arrays then mixing data types is allowed. The example given is:

```toml
arr5 = [ [ 1, 2 ], ["a", "b", "c"] ]
```

In https://github.com/toml-lang/toml/issues/28#issuecomment-14034416, the comment was made that strongly-typed languages like Haskell are going to have a very difficult time mapping `arr5` to a native data structure, because of the mixing of types in the inner arrays. That is, `arr5` is an array that contains two different types: "array of ints" and "array of strings", and in Haskell those two data types are completely different and can't be combined in the same collection: arr5 is neither an "array of arrays of ints" nor an "array of arrays of strings" and is thus invalid.

After some discussion, @mojombo [decided](https://github.com/toml-lang/toml/issues/28#issuecomment-14276709) to forbid mixing data types in arrays, and moved the `[ [ 1, 2 ], ["a", "b", "c"] ]` example to the "things that are not allowed" section of the spec. He created https://github.com/toml-lang/toml/pull/154 for this, which also included adding a syntax for tuples. The discussion on #154 went on for some time, with a lot of back-and-forth between @dahu (who wanted to allow heterogenous arrays) and @BurntSushi (who was in favor of keeping arrays homogenous and forbidding arrays like `[ [ 1, 2 ], ["a", "b", "c"] ]`).

In the end, #154 ended up being closed in favor of https://github.com/toml-lang/toml/pull/235 (motivation found in #219). But what was lost in the move from #154 to #219 was the decision about homogenous arrays. #154 updated the spec to forbid `[ [ 1, 2 ], ["a", "b", "c"] ]`, but #219 did not.

So despite the fact that both @mojombo and @BurntSushi appear to have been in favor of forbidding `[ [ 1, 2 ], ["a", "b", "c"] ]`, the spec still allows that mixing today. This has caused some confusion (see https://github.com/toml-lang/toml/issues/28#issuecomment-244587821 for example), which is why I think the discussion needs to be re-opened.

**Is `[ [ 1, 2 ], ["a", "b", "c"] ]` legal syntax or not?**

[...]

PR #663, opened 2019-09-03
Change ABNF array definition to permit single value types only

eksortso:
This modifies `toml.abnf` to enforce the current specification, which only considers homogeneous arrays to be valid. Per the `README`, strings are the same type regardless of their syntax, and arrays are the same type even if the values they contain differ in type between them. Empty arrays are still valid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ key = "value"
 ```
 
 Values must have one of the following types.
+
 - [String](#user-content-string)
 - [Integer](#user-content-integer)
 - [Float](#user-content-float)
--- a/README.md
+++ b/README.md
@@ -123,8 +123,19 @@ on the same line (though some values can be broken over multiple lines).
 key = "value"
 ```
 
-Values must be of the following types: String, Integer, Float, Boolean,
-Datetime, Array, or Inline Table. Unspecified values are invalid.
+Values must be one of the following types.
+- [String](#user-content-string)
+- [Integer](#user-content-integer)
+- [Float](#user-content-float)
+- [Boolean](#user-content-boolean)
+- [Offset Date-Time](#user-content-offset-date-time)
+- [Local Date-Time](#user-content-local-date-time)
+- [Local Date](#user-content-local-date)
+- [Local Time](#user-content-local-time)
+- [Array](#user-content-array)
+- [Inline Table](#user-content-inline-table)
+
+Unspecified values are invalid.
 
 ```toml
 key = # INVALID
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ on the same line (though some values can be broken over multiple lines).
 key = "value"
 ```
 
-Values must be one of the following types.
+Values must have one of the following types.
 - [String](#user-content-string)
 - [Integer](#user-content-integer)
 - [Float](#user-content-float)
@@ -590,9 +590,13 @@ Array
 -----
 
 Arrays are square brackets with values inside. Whitespace is ignored. Elements
-are separated by commas. Data types may not be mixed (different ways to define
-strings should be considered the same type, and so should arrays with different
-element types).
+are separated by commas.
+
+All values that are allowed in [key/value pairs](#user-content-keyvalue-pair)
+are also allowed in arrays, but all values within an array must have the same
+data type. Different ways to define strings are considered the same type, and so
+are arrays with different element types, and inline tables regardless of their
+contents.
 
 ```toml
 arr1 = [ 1, 2, 3 ]
--- a/toml.abnf
+++ b/toml.abnf
@@ -197,7 +197,10 @@ array-values =  array-strings
 array-values =/ array-booleans
 array-values =/ array-arrays
 array-values =/ array-inline-tables
-array-values =/ array-date-times
+array-values =/ array-offset-dts
+array-values =/ array-local-dts
+array-values =/ array-local-dates
+array-values =/ array-local-times
 array-values =/ array-floats
 array-values =/ array-integers
 
@@ -213,8 +216,17 @@ array-arrays =/ ws-comment-newline array ws [ array-sep ]
 array-inline-tables =  ws-comment-newline inline-table ws array-sep array-inline-tables
 array-inline-tables =/ ws-comment-newline inline-table ws [ array-sep ]
 
-array-date-times =  ws-comment-newline date-time ws array-sep array-date-times
-array-date-times =/ ws-comment-newline date-time ws [ array-sep ]
+array-offset-dts =  ws-comment-newline offset-date-time ws array-sep array-offset-dts
+array-offset-dts =/ ws-comment-newline offset-date-time ws [ array-sep ]
+
+array-local-dts =  ws-comment-newline local-date-time ws array-sep array-local-dts
+array-local-dts =/ ws-comment-newline local-date-time ws [ array-sep ]
+
+array-local-dates =  ws-comment-newline local-date ws array-sep array-local-dates
+array-local-dates =/ ws-comment-newline local-date ws [ array-sep ]
+
+array-local-times =  ws-comment-newline local-time ws array-sep array-local-times
+array-local-times =/ ws-comment-newline local-time ws [ array-sep ]
 
 array-floats =  ws-comment-newline float ws array-sep array-floats
 array-floats =/ ws-comment-newline float ws [ array-sep ]
--- a/toml.abnf
+++ b/toml.abnf
@@ -193,8 +193,34 @@ array = array-open [ array-values ] ws-comment-newline array-close
 array-open =  %x5B ; [
 array-close = %x5D ; ]
 
-array-values =  ws-comment-newline val ws array-sep array-values
-array-values =/ ws-comment-newline val ws [ array-sep ]
+array-values =  array-strings
+array-values =/ array-booleans
+array-values =/ array-arrays
+array-values =/ array-inline-tables
+array-values =/ array-date-times
+array-values =/ array-floats
+array-values =/ array-integers
+
+array-strings =  ws-comment-newline string ws array-sep array-strings
+array-strings =/ ws-comment-newline string ws [ array-sep ]
+
+array-booleans =  ws-comment-newline boolean ws array-sep array-booleans
+array-booleans =/ ws-comment-newline boolean ws [ array-sep ]
+
+array-arrays =  ws-comment-newline array ws array-sep array-arrays
+array-arrays =/ ws-comment-newline array ws [ array-sep ]
+
+array-inline-tables =  ws-comment-newline inline-table ws array-sep array-inline-tables
+array-inline-tables =/ ws-comment-newline inline-table ws [ array-sep ]
+
+array-date-times =  ws-comment-newline date-time ws array-sep array-date-times
+array-date-times =/ ws-comment-newline date-time ws [ array-sep ]
+
+array-floats =  ws-comment-newline float ws array-sep array-floats
+array-floats =/ ws-comment-newline float ws [ array-sep ]
+
+array-integers =  ws-comment-newline integer ws array-sep array-integers
+array-integers =/ ws-comment-newline integer ws [ array-sep ]
 
 array-sep = %x2C  ; , Comma
 
--- a/toml.abnf
+++ b/toml.abnf
@@ -188,7 +188,7 @@ local-time = partial-time
 
 ;; Array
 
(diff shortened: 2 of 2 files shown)


ehuss:
I don't think this is necessary, and it hinders the readability.  It is customary for grammars to not include semantic restrictions like this.  Typically languages will define arrays as something like ``` `[` _expr_, ... `]` ``` and leave the type restrictions to the language definition.  The ABNF is not intended to be used without reading the spec.

mmakaay:
ABNF can be used for many things, one of which could be document validation. What's needed here is a verdict on what the intention for the ABNF is, since this can be anything from basic tokenization until full document validation.

workingjubilee:
The intention as specified in PR #236 is as a tool for parsers to reference, and this has been alluded to elsewhere, re: testing document generation or validation, as a reference implementation essentially. It is in fact intended to have a complete match.

ChristianSi:
I see one bug in the PR: the spec clearly defined four different date/time types (Offset Date-Time, Local Date-Time, Local Date, Local Time), but the PR erroneously treats them as one, allowing them to be mixed (in contrast to the written spec).

Otherwise I repeat my plea (https://github.com/toml-lang/toml/issues/553#issuecomment-445187862) to abolishing array typing altogether, but I admit that that's a different issue.

eksortso:
I'll argue that on the contrary, all four date-time types are grouped together as a single type, "Datetime":
> Values must be of the following types: String, Integer, Float, Boolean, Datetime, Array, or Inline Table. Unspecified values are invalid.

If you can make a strong appeal to enough of the core team, especially @mojombo, that heterogeneous arrays are the way to go, then a counter-proposal would be most welcomed. You have the arguments for it already laid out, and so I say, go for it.

ChristianSi:
@eksortso You're right that the written spec say that but that too is clearly a bug. The word "Datetime" is mentioned just that single time in the spec, never again. Instead, four separate date and time-related types are defined farther down. The grouping exists, so far, only in your imagination -- though it might be a way to resolve the contradiction.

As for making a case for heterogeneous arrays, we'll see about that. I admit that I don't even have particularly strong feelings regarding that matter, though I would certainly prefer that mixing integers and floats were allowed (to prevent confusing users with nonintuitive and illogical error messages).

eksortso:
@ChristianSi I have to admit, I don't see where the spec explicitly states that each of the date/time formats constitute their own separate types. Could you provide some clarification, either here or in reference to an issue, old or new?

workingjubilee:
This is one of the cases where the spec is vague but I think the intended takeaway is clear: there is one Datetime, and it can be said different ways, and mean different things. This is probably hell on implementation, of course, since we really have something like
```
date
datetime
datetimetz
```
so I am not at all against revision or clarification, even if it constitutes an adjustment to implementation, because I suspect that any reconciliation that has been done, on the parser level, between these entities, would have already been made, and so any sensible change we would make will merely formalize what is already out there.

workingjubilee:
That said, this thread is about ABNF, TOML, and the handling of **array** types. Continuing to beg the maintainers for the abolishment of array typing is probably more on topic than discussing the nuances of Datetime's type and what that means. ^_~

ChristianSi:
@eksortso:

> I have to admit, I don't see where the spec explicitly states that each of the date/time formats constitute their own separate types.

Though the spec is somewhat vague when it comes to talking about types, I'd say that the TOC is already a strong hint:

> - ...
> - String
> - Integer
> - Float
> - Boolean
> - Offset Date-Time
> - Local Date-Time
> - Local Date
> - Local Time
> - Array
> - Table
> - Inline Table
> - Array of Tables
> - ...

Each of the entries defines a type, though the "Table" and "Array of Tables" types are special -- we might call them structural types instead of value types since they form the basic structure of a TOML document and cannot be used as values (in key/value pair or arrays), while all other types can.

The "Array of Tables" section explicitly starts with: "The last *type* that has not yet been expressed..." (emphasis added).

When we look into the past, we notice that, a long time ago, TOML did indeed have just a single Datetime type. This was changed by PR #414 ("Specify OffsetDateTime, LocalDateTime, LocalDate, and LocalTime types") which eliminated that type and introduced the four different types we have today. Note that the issue title talks about "types". And the PR was a spin-off of issue #362 ("Treat Dates and Datetimes as separate types"). Separate types, indeed.

That PR was followed by short time later by PR #442 ("Fixed datetime links in table of contents") which eliminated the now-orphaned "Datetime" link from the TOC, replacing it with links to the four different types we have now. But nobody, except you, had noticed the other orphaned "Datetime" mention which has survived, until this day, in the "Key/Value Pair" section. But that was clearly an accident of history, not a deliberate decision.

I rest my case.

eksortso:
Well, @ChristianSi. With that degree of prosecutorial zeal, you could have made heterogeneous arrays the standard by now. There's still time for that though.

But more importantly to you, I concede. All of the date/time formats are separate types, and this will be made explicit.

Since the array syntax only addresses value types, I made certain that each type indicated in `val` in `toml.abnf` got its due. This does suggest that the best place to enumerate all of the valid types is where value types are first made explicit.

Updates are coming soon.

By the way, let's mention #596. I already agreed with you once before, and I only recently remembered.

ChristianSi:
@eksortso: That looks good know, but since we're now also cleaning up the written spec I'd suggest one further revision. With your latest change, the written spec is up-to-date regarding the value types allowed in key-value pairs, but it is not very explicit about what's allowed in an array.

> Arrays are square brackets with values inside.

Supposedly, that refers to the values inside a key/value pair, but that should be made explicit. Hence I'd suggest the replace with sentence

> Data types may not be mixed (different ways to define strings should be considered the same type, and so should arrays with different element types).

with something like

> All values that are allowed in [key/value pairs](#user-content-keyvalue-pair) are also allowed in arrays, but all values within an array must have the same data type. Different ways to define strings are considered the same type, and so are arrays with different element types, and inline tables regardless of their contents.

ChristianSi:
Also, to remove the conceptual oddity (arrays are brackets??), I'd suggest to replace

> Arrays are square brackets with values inside.

with

> Arrays are sequences of comma-separated values enclosed in square brackets.

README.md
```diff
@@ -123,8 +123,19 @@ on the same line (though some values can be broken over multiple lines).
 key = "value"
 ```
 
-Values must be of the following types: String, Integer, Float, Boolean,
-Datetime, Array, or Inline Table. Unspecified values are invalid.
+Values must be one of the following types.
```
ChristianSi: Grammar: likely this should be "Values must be of one of the following types." or "Values must have one of the following types."

eksortso: I'll go with "have one".

eksortso:
> Arrays are sequences of comma-separated values enclosed in square brackets.

Not going to use this. It means exactly the same thing as "square brackets with values inside", with additional aspects spelled out in the following sentences.

eksortso:
> All values that are allowed in [key/value pairs](#user-content-keyvalue-pair) are also allowed in arrays, but all values within an array must have the same data type. Different ways to define strings are considered the same type, and so are arrays with different element types, and inline tables regardless of their contents.

I like this, and the link back to the section containing the value type list is a good addition. It's going in.

README.md
```diff
@@ -123,8 +123,19 @@ on the same line (though some values can be broken over multiple lines).
 key = "value"
 ```
 
-Values must be of the following types: String, Integer, Float, Boolean,
-Datetime, Array, or Inline Table. Unspecified values are invalid.
+Values must have one of the following types.
+- [String](#user-content-string)
```
pradyunsg: nit: empty line before list.

README.md
```diff
@@ -579,9 +590,13 @@ Array
 -----
 
 Arrays are square brackets with values inside. Whitespace is ignored. Elements
-are separated by commas. Data types may not be mixed (different ways to define
-strings should be considered the same type, and so should arrays with different
-element types).
+are separated by commas.
+
+All values that are allowed in [key/value pairs](#user-content-keyvalue-pair)
```
pradyunsg: That's means... all values are allowed?

I'd prefer that we rephrase this paragraph as something along the lines of:

> \<prev-paragraph\> All elements within an array must have the same data type.
>
> All string values are considered the same type. All arrays (regardless of types of elements in them) are considered the same type. All inline tables (regardless of the key-value pairs in them) are considered the same type.

pradyunsg:
Thanks @eksortso, @ChristianSi and @workingjubilee for the discussion on this PR. Much appreciated! :)

ChristianSi:
@eksortso:

> With that degree of prosecutorial zeal, you could have made heterogeneous arrays the standard by now.

I don't know about that, but here is my attempt: #665. Support (or even disagreement) welcome.

README.md
eksortso: Picked. Adding the empty line.

workingjubilee:
I'll work up an amicus brief. 💖

README.md
ChristianSi: @pradyunsg "All values that are allowed in key/value pairs are also allowed in arrays..." – That (partial) sentence or something similar is essential because we have to explain which types are allowed in arrays (String, Integer, Float, Boolean, Offset Date-Time, Local Date-Time, Local Date, Local Time, Array, Inline Table) and which are not (Table, Array of Tables), and currently we don't.

It might be possible to rephrase it a bit, say: "An array collects values of any of the types that are allowed in key/value pairs..." But we must be explicit about this, and a backreference to the key/value pairs it much better than having to list all (currently ten) allowed types again.

eksortso: Optionally, we could move the statements about string, array, and inline table types up to the proposed Value Types section. We'd simplify it, though, if we did that.

----

All string values are considered the same type. All arrays are considered the same type, regardless of content. All inline tables are considered the same type, regardless of content.

eksortso: Having said all that, let me propose the following synthesis of both of your suggestions. We'd also include a backreference from "value type".

---

Arrays are square brackets with values inside. Whitespace is ignored. Elements
are separated by commas.

All elements within an array must have the same value type.

_(Examples go here.)_

eksortso:
> I don't know about that, but here is my attempt: #665. Support (or even disagreement) welcome.

I'll support it. But, I'm going to offer a spin that may work for both single-type and heterogeneous varieties.

eksortso:
> I'll work up an amicus brief.

Definitely looking forward to this!

eksortso:
@pradyunsg wrote:
> I _think_ the following ABNF is equivalent and more readable:

Definitely more readable. Equivalent, too. Consider it done.

toml.abnf
```diff
@@ -188,13 +188,51 @@ local-time = partial-time
...
+array-values =/ array-arrays
+array-values =/ array-inline-tables
+array-values =/ array-offset-dts
+array-values =/ array-local-dts
+array-values =/ array-local-dates
+array-values =/ array-local-times
+array-values =/ array-floats
+array-values =/ array-integers
+
+array-strings =  string ws array-sep ws-comment-newline array-strings
```
pradyunsg:
```abnf
array-value-sep = ws array-sep ws-comment-newline
array-value-ending = ws [ array-sep ]

array-strings =  string array-value-sep array-strings
array-strings =/ string array-value-ending
```

README.md
pradyunsg: > Arrays are square brackets with values inside. Whitespace is ignored. Elements
> are separated by commas.
> 
> All elements within an array must have the same value type.

LGTM.

ChristianSi:
If #676 is merged, the proposed changes to the abnf would be obsolete. However, the changes to the README (update the list of allowed value types which is currently outdated) still remain relevant and are worth rescuing – maybe fork them off into a different PR?

pradyunsg:
Well, I just merged in #676.

I second the suggestion that 4a72cb0 (and friends) would be good to have. I leave it to @eksortso's judgement on whether to open a new PR or repurpose this one. :)

pradyunsg:
Actually, on second though, I'll go ahead and close this PR -- the other changes aren't exactly related to the original intent of the PR and it'll be "cleaner" to start fresh vs continue work in the this PR.

eksortso:
@pradyunsg That's alright, mostly. Just as long as it moved one way or another, I'm glad, and I took up the cause to get things moving. But I sincerely ran out of time to work on this proposal.

I'll make time to possibly reintroduce some of the other concepts that made their way here. So the branch won't be deleted too soon.

pradyunsg:
No worries. I think I can pick it up. Thanks for getting things moving here! :)

PR #664, opened 2019-09-03
Explain ordering of subtables under arrays of tables

### LINKED ISSUE #658: Explain ordering of subtables under arrays of tables
The spec says: "You can create nested arrays of tables as well. Just use the same double bracket syntax on sub-tables. Each double-bracketed sub-table will belong to the most recently defined table element above it."

This does not, however, explicitly mention that case that the subtable is a regular table rather than a table array -- though clearly the spec expects this case to be valid as well. This too should be mentioned and a suitable example added, e.g.

```toml
[[fruits]]
name = "pomegranate"

[fruits.tastes]
forbidden = true
sweet = true
spicy = false

[[fruits]]  # et cetera
name = "peach"
```

The spec also should make it clear that defining subtables (whether regular tables or arrays) *before* the first entry of a table array is not allowed, since it's not clear to which element of the (so far empty) table array they should be attached.

```
# INVALID
[fruits.tastes]  # this works only if "fruits" is a regular table
forbidden = true
sweet = true
spicy = false

[[fruits]]  # parser should throw an ERROR upon discovering that it's a table array
name = "pomegranate"

[[fruits]]  # et cetera
name = "peach"
```

Another issue about which the spec might be clarified are attempts to mix table arrays and plain tables. The spec already says: "Attempting to define a normal table with the same name as an already established array must produce an error at parse time", giving this example:

```toml
[[fruit.variety]]
  name = "red delicious"

# This table conflicts with the previous table
[fruit.variety]
  name = "granny smith"
```

While some might consider it obvious, it would be useful to add a "or vice versa" to make it clear that the reverse order is forbidden too:

```toml
[fruit.variety]
  name = "granny smith"

# This table conflicts with the previous table
[[fruit.variety]]
  name = "red delicious"
```

This issue is a spin-off of #651 – thanks @workingjubilee for bringing this to my attention.

@pradyunsg: Maybe add it to the critical path for TOML 1.0?

ChristianSi:
* Explicitly mention that table arrays can have regular subtables.
* Document that children of table arrays must not be defined above their parents (with example).
* Document that attempts to redefine a normal table as an array must produce a parse-time error (with example).

Closes #658.

### DIFF
--- a/README.md
+++ b/README.md
@@ -819,18 +819,17 @@ In JSON land, that would give you the following structure.
 
 You can create nested arrays of tables as well. Just use the same double bracket
 syntax on sub-tables. Each double-bracketed sub-table will belong to the most
-recently defined table element above it. Normal sub-tables (not arrays) likewise
-belong to the most recently defined table element above them.
+recently defined table element above it.
 
 ```toml
 [[fruit]]
   name = "apple"
 
-  [fruit.physical]  # subtable
+  [fruit.physical]
     color = "red"
     shape = "round"
 
-  [[fruit.variety]]  # nested array of tables
+  [[fruit.variety]]
     name = "red delicious"
 
   [[fruit.variety]]
--- a/README.md
+++ b/README.md
@@ -906,7 +906,7 @@ as an array must likewise produce a parse-time error.
   [[fruit.variety]]
     name = "red delicious"
 
-  # Not allowed: This table conflicts with the previous array of tables
+  # INVALID: This table conflicts with the previous array of tables
   [fruit.variety]
     name = "granny smith"
 
@@ -914,7 +914,7 @@ as an array must likewise produce a parse-time error.
     color = "red"
     shape = "round"
 
-  # Not allowed: This array of tables conflicts with the previous table
+  # INVALID: This array of tables conflicts with the previous table
   [[fruit.physical]]
     color = "green"
 ```
--- a/README.md
+++ b/README.md
@@ -819,17 +819,18 @@ In JSON land, that would give you the following structure.
 
 You can create nested arrays of tables as well. Just use the same double bracket
 syntax on sub-tables. Each double-bracketed sub-table will belong to the most
-recently defined table element above it.
+recently defined table element above it. Normal sub-tables (not arrays) likewise
+belong to the most recently defined table element above them.
 
 ```toml
 [[fruit]]
   name = "apple"
 
-  [fruit.physical]
+  [fruit.physical]  # subtable
     color = "red"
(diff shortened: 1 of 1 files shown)


ChristianSi:
@pradyunsg Can you find time to look at this?

README.md
```diff
@@ -819,17 +819,18 @@ In JSON land, that would give you the following structure.
 
 You can create nested arrays of tables as well. Just use the same double bracket
 syntax on sub-tables. Each double-bracketed sub-table will belong to the most
-recently defined table element above it.
+recently defined table element above it. Normal sub-tables (not arrays) likewise
```
pradyunsg: This is self-evident from the example JSON below, so I don't think we should call this out explicitly. None the less, if you think we should, let's tackle that in a separate PR?

Same for the `# subtable` and `# nested` comments below.

README.md
```diff
@@ -868,18 +869,34 @@ The above TOML maps to the following JSON.
...
+[[fruit]]  # parser must throw an error upon discovering that "fruit" is
+           # an array rather than a table
+  name = "apple"
+```
+
 Attempting to append to a statically defined array, even if that array is empty
 or of compatible type, must produce an error at parse time.
 
-```toml
+```
```
pradyunsg: 👍

README.md
```diff
@@ -889,9 +906,17 @@ array must produce an error at parse time.
   [[fruit.variety]]
     name = "red delicious"
 
-  # This table conflicts with the previous table
+  # Not allowed: This table conflicts with the previous array of tables
```
pradyunsg:
```suggestion
  # INVALID: This table conflicts with the previous array of tables
```

README.md
```diff
@@ -889,9 +906,17 @@ array must produce an error at parse time.
...
   [fruit.variety]
     name = "granny smith"
+
+  [fruit.physical]
+    color = "red"
+    shape = "round"
+
+  # Not allowed: This array of tables conflicts with the previous table
```
pradyunsg:
```suggestion
  # INVALID: This array of tables conflicts with the previous table
```

ChristianSi:
@pradyunsg Thanks for the review. I've made the requested changes.

ChristianSi:
Following @pradyunsg's suggestion, that explanation that table arrays can have regular subtables has moved into a separate PR: #666.

ChristianSi:
@pradyunsg What's the status? I've modified the PR as suggested.

pradyunsg:
Thanks! I've been AWOL from TOML for a while now for RL reasons.

PR #666, opened 2019-09-11
Explicitly mention that arrays of tables can contain subtables

ChristianSi:
This is a spin-off of #664. It makes it explicit in the written spec that arrays of tables can contain regular (non-array) subtables which belong to the most recently defined table element above them.

@pradyunsg considered this unnecessary, arguing that "this is self-evident from the example JSON below". However, I would object that the examples should *not* introduce additional features, they should only *exemplify* what's already mentioned in the text.

Additionally, this particular example is currently a bit hard to understand since the text above it only talks about "double-bracketed sub-tables" (nested arrays of tables), making it surprising to also find a single-bracketed sub-table there (in fact, I initially overlooked it totally, instead just seeing what I had been told I would see).

This PR brings the text above the example in agreement with the example itself, following the "Explicit is better than implicit" principle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -819,17 +819,18 @@ In JSON land, that would give you the following structure.
 
 You can create nested arrays of tables as well. Just use the same double bracket
 syntax on sub-tables. Each double-bracketed sub-table will belong to the most
-recently defined table element above it.
+recently defined table element above it. Normal sub-tables (not arrays) likewise
+belong to the most recently defined table element above them.
 
 ```toml
 [[fruit]]
   name = "apple"
 
-  [fruit.physical]
+  [fruit.physical]  # subtable
     color = "red"
     shape = "round"
 
-  [[fruit.variety]]
+  [[fruit.variety]]  # nested array of tables
     name = "red delicious"
 
   [[fruit.variety]]


ChristianSi:
Ping @pradyunsg (and other interested parties): What do you think?

README.md
```diff
@@ -819,17 +819,18 @@ In JSON land, that would give you the following structure.
 
 You can create nested arrays of tables as well. Just use the same double bracket
 syntax on sub-tables. Each double-bracketed sub-table will belong to the most
-recently defined table element above it.
+recently defined table element above it. Normal sub-tables (not arrays) likewise
+belong to the most recently defined table element above them.
```
workingjubilee: I realize it was here previously but "above it" is redundant.

README.md
```diff
@@ -819,17 +819,18 @@ In JSON land, that would give you the following structure.
...
 [[fruit]]
   name = "apple"
 
-  [fruit.physical]
+  [fruit.physical]  # subtable
     color = "red"
     shape = "round"
 
-  [[fruit.variety]]
+  [[fruit.variety]]  # nested array of tables
```
workingjubilee: Comment annotations here look good to me.

README.md
pradyunsg: Seconded.

ChristianSi:
@pradyunsg I'm a bit unsure about the "Seconded" comments you left. Do you want me to delete the words "above it" and "above them", as @workingjubilee suggested? I can do that.

ChristianSi:
Ping?

pradyunsg:
> Do you want me to delete the words "above it" and "above them", as @workingjubilee suggested?

Yep. Sorry that I wasn't clear -- that was probably written in a hurry but that's a bad excuse to not communicate clearly.

pradyunsg:
Actually, since that's sorta independent of the change in this PR, I'm happy to merge this in as-is.

If someone would like to file a follow-up PR to make that change, that'd be great! ^>^

PR #673, opened 2019-09-26
Make the ABNF syntax tree, more semantic for ml-basic-string

### LINKED ISSUE #669: More meaningful ml-basic-body's newline escaping
## Context

The ABNF spec, while still valid on that part, could benefit from some improvement:

The README.md shows:
```toml
str2 = """
The quick brown \


  fox jumps over \
    the lazy dog."""
```

And in ABNF:
```ABNF
ml-basic-body = *( ml-basic-char / newline / ( escape ws newline ) )
```

This works:  After the first `\`, there might be multiple whitespaces until a newline is reached, then the next newlines match the standalone `newline` token.

## Problem

But this does not correspond to the semantics: the `newline` token is for inserting an actual newline token into the buffer, while `escape ws newline` is for skipping newlines or whitespaces.

## Proposed solution

I suggest either adding a `ignored-ws-nl` ABNF token:
```ABNF
ml-basic-body = *( ml-basic-char / newline / escaped-ws-nl )
escaped-ws-nl = escape 1*( wschar / newline )
```

Or changing the current token:
```ABNF
ml-basic-body = *( ml-basic-char / newline / ( escape 1*( wschar / newline ) )
```

[EDIT]: it needs at least one whitespace *or* newline

Both seems compatible with the proposed solutions for #640.

I am going for the extra token in my implementation.  I can provide a patch if wanted.

Cheers!

josuah:
this:
```
string = """ escaped \

     end"""
```

Was parsed as a single escaped newline, then as a newline part of the
string itself.

While it is fitting the role of rejecting invalid input and accepting
the valid one, it did not fit the role of parsing the content into
tokens.

To help the implementation, a multiline whitespace/newline token is
introduced: escaped-ws-nl

This fixes #669.

### DIFF
--- a/toml.abnf
+++ b/toml.abnf
@@ -84,9 +84,10 @@ ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
 
 ml-basic-string-delim = 3quotation-mark
 
-ml-basic-body = *( ml-basic-char / newline / ( escape ws newline ) )
+ml-basic-body = *( ml-basic-char / newline / ml-basic-escaped-nl )
 ml-basic-char = ml-basic-unescaped / escaped
 ml-basic-unescaped = wschar / %x21-5B / %x5D-7E / non-ascii
+ml-basic-escaped-nl = escape ws newline *( wschar / newline )
 
 ;; Literal String
 


ChristianSi:
Looks good to me.

toml.abnf
```diff
@@ -84,9 +84,10 @@ ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
 
 ml-basic-string-delim = 3quotation-mark
 
-ml-basic-body = *( ml-basic-char / newline / ( escape ws newline ) )
-ml-basic-char = ml-basic-unescaped / escaped
-ml-basic-unescaped = wschar / %x21-5B / %x5D-7E / non-ascii
+ml-basic-body = ws *( ml-basic-char ws / newline / ml-basic-escaped-ws-nl )
+ml-basic-char = ( ml-basic-unescaped / escaped ) ws
+ml-basic-unescaped = %x21-5B / %x5D-7E / non-ascii
+ml-basic-escaped-ws-nl = escape newline *( wschar / newline )
```
ehuss: Doesn't there need to be whitespace between `escape` and `newline`?

toml.abnf
```diff
@@ -84,9 +84,10 @@ ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
 
 ml-basic-string-delim = 3quotation-mark
 
-ml-basic-body = *( ml-basic-char / newline / ( escape ws newline ) )
-ml-basic-char = ml-basic-unescaped / escaped
-ml-basic-unescaped = wschar / %x21-5B / %x5D-7E / non-ascii
+ml-basic-body = ws *( ml-basic-char ws / newline / ml-basic-escaped-ws-nl )
+ml-basic-char = ( ml-basic-unescaped / escaped ) ws
```
ehuss: I'm not sure I see how removing whitespace from `ml-basic-unescaped` is trying to do.  It seems to make it a bit more complex, but I don't see why.  Can you say more about what this changes and why?

Also, this seems to duplicate `ws` at the end here, since it is also included in `ml-basic-body`.

josuah: First off, the README.md is not matching the ABNF:

> When the last non-whitespace character on a line is a \, it will be trimmed along with all whitespace (including newlines) up to the next non-whitespace character or closing delimiter.

So `( escape ws newline )` shoud me something more like `( escape ws newline 1*( wschar / newline ) )`.

So for two successive tokens:
```
( escape ws newline 1*( wschar / newline ) ) ( escape ws newline 1*( wschar / newline ) ) [...]
```

So we can end-up with:

```
( escape ws newline 1*( wschar / newline ) ) ( ml-basic-char ) [...]
```

Or:
```
escape ws newline 1*( wschar / newline ) ( ml-basic-unescaped / escaped ) [...]
```

Or possibly:
```
escape ws newline 1*( wschar / newline ) ml-basic-unescaped [...]
```

Or possibly:
```
escape ws newline 1*wschar ml-basic-unescaped [...]
```

Or:
```
escape ws newline 1*wschar ( wschar / %x21-5B / %x5D-7E / non-ascii ) [...]
```


Or possibly:
```
escape ws newline 1*wschar wschar [...]
```

And then we see that the first ` 1*wschar` is part of  the "escape" token, and the second `wschar` is part of ml-basic-body.

So face to this:

```
escape wschar newline wschar wschar wschar wschar wschar [...]
```

Where do we draw a line between the `wschar` which is part of escape, and the `wschar` which is part of the normal text character (which can also be a space).

Not only there might be this bug in the proposed version, but it might also be present on mine, which is still wrong!

josuah: I redid that commit to fix it partially and simplify the `ws` mess I had:

```
ml-basic-body = *( ml-basic-char / wschar / newline / ml-basic-escaped-ws-nl )
ml-basic-char = ( ml-basic-unescaped / escaped )
ml-basic-unescaped = %x21-5B / %x5D-7E / non-ascii
ml-basic-nonblank = %x21-7E / non-ascii
ml-basic-escaped-ws-nl = escape newline *( wschar / newline ) ml-basic-nonblank
```

This now matches much more closely the sentence from the `README.md` (yay!):

> When the last non-whitespace character on a line is a `\`, it
> will be trimmed along with all whitespace (including newlines) up to the next
> non-whitespace character **or closing delimiter**.

But it is still wrong as as you can see in bold, this won't work for a \ whitespaces at the end of a basic multiline string.

But to have that in the ABNF definition, we will need to duplicate quite a large proportion, because if we include the closing delimiter in the `ml-basic-escaped-ws-nl`, then we have to *not* include it in the `ml-basic-string` delimiter for not having it duplicated.

So we would need this instead:

```
;; Multiline Basic String

ml-basic-string =  ml-basic-string-delim ml-basic-body ml-basic-string-delim
ml-basic-string =/ ml-basic-string-delim ml-basic-body escape newline *( wschar / newline ) ml-basic-string-delim

ml-basic-string-delim = 3quotation-mark

ml-basic-body = *( ml-basic-char / wschar / newline / ml-basic-escaped-ws-nl )
ml-basic-char = ( ml-basic-unescaped / escaped )
ml-basic-unescaped = %x21-5B / %x5D-7E / non-ascii
ml-basic-escaped-ws-nl = escape newline *( wschar / newline ) ml-basic-nonblank
ml-basic-nonblank = %x21-7E / non-ascii
```

toml.abnf
```diff
@@ -84,9 +84,11 @@ ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
...
 ml-basic-string-delim = 3quotation-mark
 
-ml-basic-body = *( ml-basic-char / newline / ( escape ws newline ) )
-ml-basic-char = ml-basic-unescaped / escaped
-ml-basic-unescaped = wschar / %x21-5B / %x5D-7E / non-ascii
+ml-basic-body = *( ml-basic-char / wschar / newline / ml-basic-escaped-ws-nl )
+ml-basic-char = ( ml-basic-unescaped / escaped )
+ml-basic-unescaped = %x21-5B / %x5D-7E / non-ascii
+ml-basic-nonblank = %x21-7E / non-ascii
+ml-basic-escaped-ws-nl = escape newline *( wschar / newline ) ml-basic-nonblank
```
ChristianSi: I don't understand what's `ml-basic-nonblank` token is doing at the end of this rule. Supposedly, that non-blank character should NOT be escaped, but rather be part of the normal string content? Also, it's possible that the "line ending backslash" + subsequent ignored whitespace is immediately followed by a "closing delimiter", but the  `ml-basic-nonblank` token prevents that.

My impression is that the `ml-basic-nonblank` token could, and should, be removed to get the correct behavior. That's true, at least, if ABNF rules are "greedy", matching as many `wschar / newline` as the end of the `ml-basic-escaped-ws-nl` as they possibly can – which in my understanding is the case. ('The [ABNF] repetition operator is "greedy", meaning that it always matches the longest possible string. It never backtracks to see if a shorter match would succeed', says [the Internet](http://www.coasttocoastresearch.com/overview).)

ChristianSi: Also, I'd propose to rename the rule to `ml-basic-escaped-nl`. After all, the escaped thing is always the newline, not any other whitespace (though other whitespace might follow the escaped newline).

josuah: The fact that ABNF operator `*` is greedy solves everything!

That is not specified in ABNF RFC but RFC stands for "Request For Comments", not "Rules Forever Constant".  What matters is that the message reach parser implementers.

I see reason to have the `ml-basic-escaped-nl` at the end, and we can now have semantics back.

I amended the suggested changes to the commit.

josuah: To my greatest surprise, instaparse is non-greedy!  (noticed that while checking).

https://github.com/Engelberg/instaparse#regular-expressions-a-word-of-warning

Do we still consider ABNF's "*" to be greedy?

josuah: It is not either greedy or non-greedy, it shows all alternative forms.

With the original version, we have only one possible form (because it only takes the first space).

What a storm of irony: I created the problem I was trying to solve!

toml.abnf
```diff
@@ -84,9 +84,10 @@ ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
 
 ml-basic-string-delim = 3quotation-mark
 
-ml-basic-body = *( ml-basic-char / newline / ( escape ws newline ) )
-ml-basic-char = ml-basic-unescaped / escaped
+ml-basic-body = *( ml-basic-char / wschar / newline / ml-basic-escaped-nl )
```
ChristianSi: The `wschar` token seems unnecessary here, since it's already part of `ml-basic-char`.

toml.abnf
```diff
@@ -84,9 +84,10 @@ ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
...
+ml-basic-char = ( ml-basic-unescaped / escaped )
```
ChristianSi: The parentheses surrounding the role are unnecessary and should be dropped (reverting to the one rule).

toml.abnf
```diff
@@ -84,9 +84,10 @@ ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
...
 ml-basic-unescaped = wschar / %x21-5B / %x5D-7E / non-ascii
+ml-basic-escaped-nl = escape newline *( wschar / newline )
```
ehuss: This seems to remove the whitespace between the escape and the newline?

ChristianSi: @ehuss You're right. The rule should start with `escape ws newline` rather than `escape newline`, since the written spec clearly says that whitespace is allowed here: "When the last non-whitespace character on a line is a `\` ...".

josuah: And `escape ws newline` (that was here at the beginning) along with the extra `*( wschar / newline )` added seems to be fitting that sentence.  Both under the new token matching all input to skip.

The changeset is more minimal now.

toml.abnf
josuah: Indeed, this was also not present in the original input.

ChristianSi:
@josuah Soory, but that's still not correct. The written spec clearly says that whitespace (spaces + tabs) is *allowed* between the backslash and the (first) newline.

toml.abnf
eksortso: My apologies, @josuah, for pushing you the wrong way. I misread the rule in the spec and gave you bad advice. I'm sorry.

pradyunsg:
FYI -- my OSS time has been fairly limited lately, this is on my radar.

josuah:
I wonder how I missed this.  I will try to save you more time than I cost to you.

1. I added `ws` between `escape` and `newline` for `ml-basic-escaped-nl`
2. I removed `wschar` that I added earlier as it is already included in `ml-basic-unescaped`

Now the changeset is only adding a new token and appending `*( wschar / newline )` after `escape ws newline`

toml.abnf
josuah: @eksortso I am to blame as much for not checking the clear and unambiguous sentence "last non-whitespace character on a line is a `\`".

Should we rephrase it like so it is more foolproof (1) ?

```diff
-When the last non-whitespace character on a line is a \,
+If a \ is followed by (optional) whitespaces and then a newline,
```

The original sentence is still fine if you stick to decomposing its accurate meaning and do not read what you want to see.

(1): https://1.bp.blogspot.com/-HucRsU8S-OM/XMxkwBN4uoI/AAAAAAAACdE/IdF0VaUqXsI8xt_7OjAurjlp_47pAyEoACPcBGAYYCw/s1600/MultilineString.png

ChristianSi:
The proposed changes to the ABNF look good to me now.

However, why are there also changes to the Chinese README in there? Those seem entirely irrelevant to this PR.

pradyunsg:
Ah, right. This change to the ABNF doesn't change anything about the language that's accepted -- it's changing the "tree" that comes out of the parsing:

```
:ml-basic-char
	:ml-basic-unescaped
		:wschar 
:escape "\"
:ws " "
:newline "\n"
:newline "\n"
:ml-basic-char
	:ml-basic-unescaped
		:wschar 
```

With this PR:

```
:ml-basic-char
	:ml-basic-unescaped
		:wschar 
:ml-basic-escaped-nl
	:escape "\"
	:ws " "
	:newline "\n"
:newline "\n"
:ml-basic-char
	:ml-basic-unescaped
		:wschar 
```

PR #680, opened 2019-11-07
Clarify when array of tables ends

pradyunsg:
Closes #558

### DIFF
--- a/README.md
+++ b/README.md
@@ -794,7 +794,8 @@ The last type that has not yet been expressed is an array of tables. These can
 be expressed by using a table name in double brackets. Each table with the same
 double bracketed name will be an element in the array. The tables are inserted
 in the order encountered. A double bracketed table without any key/value pairs
-will be considered an empty table.
+will be considered an empty table. Under that, and until the next table or EOF
+are the key/values of that table.
 
 ```toml
 [[products]]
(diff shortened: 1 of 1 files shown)


ChristianSi:
The text now reads a bit strangely:

> A double bracketed table without any key/value pairs will be considered an empty table. Under that, and until the next table or EOF are the key/values of that table.

So under that "empty table" are the key/values of the (now suddenly non-empty) table? That doesn't make much sense.

Hence I'd rather suggest to move the newly inserted sentence a few sentences up:

> The last type that has not yet been expressed is an array of tables. These can be expressed by using a table name in double brackets. *Under that, and until the next table or EOF are the key/values of that table.* Each table with the same double bracketed name will be an element in the array *of tables.* The tables are inserted in the order encountered. A double bracketed table without any key/value pairs will be considered an empty table.

(Italics marking proposed changes.)

More generally: Before you merge a PR of your own, maybe give others 1–2 days to review it?

pradyunsg:
/me nods.

All that you're saying sounds reasonable to me! ^.^

ChristianSi:
@pradyunsg What now? Should I push a PR performing the suggested sentence switch? Or do you want to do it?

pradyunsg:
Please do feel free to file a PR -- I don't think I'll get to this before end-of-day today. :)

ChristianSi:
See  #684 – please review and merge.

PR #690, opened 2019-12-18
Clarify that trailing commas are not allowed in inline tables

### LINKED ISSUE #569: Unclear whether trailing commas are allowed in inline tables
It is unclear from the specification whether trailing commas are allowed inside inline tables.
For example:

    a = { b = 1, c = 2 , } # is this allowed ?

The section **Inline Table** says nothing about trailing commas, implying that they are not allowed.
However, a casual reader of the specification could reasonably assume that the same rules apply to trailing commas in arrays as in inline tables.

The ABNF definition forbids trailing commas in inline tables, but the ABNF is currently not authoritative.
I think the section **Inline Table** should explicitly forbid trailing commas to avoid possible confusion.

pradyunsg:
Closes #569

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Leading zeroes in exponent parts of floats are permitted.
 * Clarify behavior of tables defined implicitly by dotted keys.
 * Clarify that inline tables are immutable.
+* Clarify that trailing commas are not allowed in inline tables.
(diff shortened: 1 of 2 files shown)


ChristianSi:
Looks good to me.

mojombo:
@pradyunsg This looks good to me as-is.
@eksortso I'm not sure what your suggested edit clarifies.

PR #691, opened 2019-12-18
Clarify that control characters are not permitted in comments

### LINKED ISSUE #567: Unclear whether control characters are allowed in comments
It is not clear from the specification whether control characters (U+0000 .. U+001F and U+007F) are allowed in comments.

The section **Comment** only states that "_A hash symbol marks the rest of the line as a comment._".
This implies that any non-newline characters are allowed after the hash symbol.

Discussion about control characters only appears in the section **String**, which should have no implication on comments.

The ABNF definition forbids control characters (except tab) in comments, but the ABNF is not authoritative.

I think the section **Comment** should explicitly state which characters are allowed in comments.
Note that it would be unlogical to allow e.g. form-feed inside comments, because form-feed is traditionally a stronger separator than newline.

pradyunsg:
Fixes #567

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Leading zeroes in exponent parts of floats are permitted.
+* Clarify that control characters are not permitted in comments.
 * Clarify behavior of tables defined implicitly by dotted keys.
 * Clarify that inline tables are immutable.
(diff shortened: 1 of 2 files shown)


ChristianSi:
LGTM.

mojombo:
Looks good to me as well. :shipit:

PR #692, opened 2019-12-22
Add Pradyun, et al.

mojombo:
@pradyunsg If you're ok with it, I'd like to add your name to the byline of the spec. This also acknowledges that many others have given time and effort in the creation of today's TOML spec.


ChristianSi:
Good idea!

pradyunsg:
Yep, works with me. :)

Issue #701, opened 2020-02-05
Suggestions to enhance TOML documentation and specification


Suhoy95:
Hello,

I have prepared Russian translation for TOML  (#700).  I got several ambiguation. Here are my notes, which I think should be corrected.

Notes which request a better explanations / example:

- [x]  1. In Spec there are defined Whitespace and Newline, but in the multi line Strings uses term "Whitespace characters", which includes both terms. 
- [x]  6. In the data-time sections a lot of duplications which could be express in better way (I suppose) (related PR #696)
- [x]  7. Some ambiguation: you text that "A TOML file must be a valid UTF-8 encoded Unicode document",
but class Whitespace characters is defined very poor. What if there are 
other [UTF-8 whitespace characters](http://emptycharacter.com/) before a Key?

Work-in-Progress:

- [x]  3. Some paragraphs contains more than one idea, and it is better to split them into several to improve readability. For example, when multi-line Strings are defined at the same time there is enumeration of its properties, which may be separated and have dedicated link anchor. (WIP #705)
- [x]  5. In Integer definition as I suppose, it says "64 bit (signed long) range expected ...", but it is not clear defined that implementation should use signed long encoding. (WIP #705, alse there is discussion in #688)

Closed notes: 

- [X]  2. It is useful to use more heading levels with `#`, `##`, `###` and etc, It makes table of content more structured and easy to refer to exact part if it require. (Note: also it would be useful to `pdf` rendering with pandoc as Example) (this topic are touched in #702, and moved to #703 and #704 )
- [X]  4. Table should be renamed to Hash Table or Dictionary. This word makes wrong association with CSV and Spreadsheets (Which I initially expected). (With context "Table" as name is OK. [comment](https://github.com/toml-lang/toml/issues/701#issuecomment-585024355))
- [X]  5.1 You try to define (or recommend) hardware limitations of numbers (like 64 bit or IEEE binary64. But as mentions a lot of times it is really hardware specific. Maybe you should describe more clear a template of relationships between toml and hardware, and provide "recommended" filled template which appropriate for the most cases (i.e. x64). (Discussion in #538)
- [X]  8. Do you have any recommendations to keep TOML document compatible with INI format, or Shell file (with variables only)? It would be good practice to prepare system to transfer from INI to TOML format. (Clearly described in [comment](https://github.com/toml-lang/toml/issues/701#issuecomment-584969917))
- [X]  9. This example really unexpected (from README.md) (~~WIP: #702~~, now it is #706):
[...]

PR #702, opened 2020-02-10
Explain nested nature of dotted key notation

Suhoy95:
Issue #701 point 9

Signed-off-by: Ilya Sukhoplyuev suhoy@gramend.ru
(Gramend R&D and Innopolis University, SNE master)

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,13 +242,23 @@ The above TOML maps to the following JSON.
 As long as a key hasn't been directly defined, you may still write to it and
 to names within it.
 
+Following JSON analogy, you should note that dotted keys creates nested Hash 
+Tables (see Table section). So you may define nested structures in a flat way 
+as long as all types is compatible.
+
+
 ```
+# THIS IS VALID: 
+#    fruit is Table with apple and orange, 
+#    apple is Table, and orange is Integer - there is no conflicts
 fruit.apple.smooth = true
 fruit.orange = 2
 ```
 
 ```
 # THIS IS INVALID
+# fruit is Table with apple,
+# but apple CAN NOT be an Integer and a Table at the same time
 fruit.apple = 1
 fruit.apple.smooth = true
 ```


Suhoy95:
I guess I have got the idea of the example

README.md
```diff
@@ -242,13 +242,23 @@ The above TOML maps to the following JSON.
 As long as a key hasn't been directly defined, you may still write to it and
 to names within it.
```
Suhoy95: I am not sure that we should keep this paragraph. It seems redundant

ChristianSi:
I agree that the current remarks on dotted keys might be confusing, but I'm not quite convinced that this PR is the best way to resolve the potential confusion.

A deeper problem in the spec is that the "Keys" section comes quite early in the document, while the "Table" section is defined much later. But in order to understand what a key is, you need at least an implicit understanding of what a table is – and, of course, you cannot understand tables without at least a vague understanding of key/value pairs (which are defined even earlier than keys).

The "Key/Value Pair" section starts with: "The primary building block of a TOML document is the key/value pair."

I'm not sure that's correct. Arguably tables (and arrays of tables) are the highest, the primary structure in a TOML document, key/value pairs are second, and keys as wells as values are third.

So key/value pairs are conceptually in the middle, which makes it a bit strange that the spec starts with them. A more principled approach would be to proceed either top down or bottom up. Assuming a top-down approach, the sections in the spec should be reordered as follows:

1. Preliminaries (unchanged):

    Example
    Spec
    Comment

2. Block-level structures and their immediate constituents:

    Table
    Array of Tables
    Key/Value Pair
    Keys (better rename to "Key" for consistency with other headings)

3. Value types:

    String
    Integer
    Float
    Boolean
    Offset Date-Time
    Local Date-Time
    Local Date
    Local Time
    Array
    Inline Table

4. Concluding sections (unchanged):

    Filename Extension
    MIME Type
    Comparison with Other Formats
    Get Involved
    Wiki

If we do it like this, the definitions and explanations related to dotted keys could also be completely contained in the "Keys" section (or maybe add a new "Dotted Key" section?), while currently they are spread over the "Keys" and "Table" section, with some redundancy between the two. And since Tables are already defined at that point, we could make the remark:

> Dotted keys define everything to the left of each dot as a table.

immediately when introducing dotted keys. (Currently that remark is in the Table section, near the end of the document, while the initial explanation of dotted keys says much more vaguely: "This allows for grouping similar properties together.")

The sketched re-organsation of the spec would be a somewhat bigger change, but I think it's a more principled solution than introducing even more redundancy in the treatment of dotted keys.

Suhoy95:
@ChristianSi , Yeah I think the similar way.

But I am prefer more nested structure with 1-6 header, it can make it more verbose and intuitive:

```md
0. Preamble
1. Example
2. Spec
    2.0 Basic types
        2.0.0 Boolean
        2.0.1 Integer
        2.0.2 Float
        2.0.3 String (`# it is cool that String comes before Keys`)
        2.0.4 Date and Time
            2.0.4.0 Offset Date-Time
            2.0.4.1 Local Date-Time
            2.0.4.2 Local Date
            2.0.4.3 Local Time
    2.1 Table
        2.1.1 Key/Value Pair (`# because it is part of the Table type`)
        2.1.2 Keys syntax (better rename to "Key" for consistency with other headings)
        2.1.3 Inline Table (special syntax)
    2.2 Array
        2.2.0 Array of Tables (special syntax)
3. Environment related points (`# I do not know how to name it better`)
    3.0 Filename Extension
    3.1 MIME Type
4. Comparison with Other Formats
5. Community
    5.0 Get Involved
    5.1 Wiki
```

As example, I am following the bottom to top, but we can add hint\remarks to point about Table type basis and facilitate understanding. 


And yes, current solution is more like short hack to fix it **right now**, but if we do not hurry, we can refactor this PR when free time will be found

Suhoy95:
Hey, here is dirty proof-of-concept: [toml.pdf](https://github.com/toml-lang/toml/files/4190821/toml.pdf)

ChristianSi:
Suhoy95 Yeah, something like that would be fine with me. Not sure how realistic it is that such a big re-organization will happen before 1.0 is released, though.

Also note that the proposal to re-arrange the TOML spec should be made and discussed in a separate (new) issue, as it's not really about the "nested nature of dotted key notation" anymore.

Suhoy95:
Now, I guess it is duplicated of #706. #706 looks more accurate

PR #703, opened 2020-02-13
Reorganization of TOML specification

Suhoy95:
The Proof-Of-Concept of reorganization which was emerged in #702.

Example: [pdf](https://github.com/toml-lang/toml/files/4190821/toml.pdf)

Signed-off-by: Ilya Sukhoplyuev suhoy@gramend.ru
(Gramend R&D and Innopolis University, SNE master)

PR #705, opened 2020-02-13
[WIP] Split paragraphs, make definition of Integer as Float

Suhoy95:
Issue #701, Point 3, 5

Signed-off-by: Ilya Sukhoplyuev suhoy@gramend.ru
(Gramend R&D and Innopolis University, SNE master)

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,9 +118,12 @@ Key/Value Pair
 
 The primary building block of a TOML document is the key/value pair.
 
-Keys are on the left of the equals sign and values are on the right. Whitespace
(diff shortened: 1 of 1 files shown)


Suhoy95:
I have began to feeling responsible as soon as I mention it. It is more as an example.  The #703 will change the game catastrophically. But this is not mean that we should ignore these little inaccuracies now and in future

pradyunsg:
Closing since this is mostly copy-editing or making a change that's otherwise covered by #736.

PR #706, opened 2020-02-14
Clarify example showing how to add new keys

eksortso:
Per a conversation regarding point 9 of #701, this PR annotates two examples with comments to show how you can write to unused names, and to show why reusing a name as a table when it originally wasn't is invalid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,13 +243,21 @@ As long as a key hasn't been directly defined, you may still write to it and
 to names within it.
 
 ```
+# This makes the key "fruit" into a table.
 fruit.apple.smooth = true
+
+# So then you can add to the table "fruit" like so:
 fruit.orange = 2
 ```
 
 ```
-# THIS IS INVALID
+# THE FOLLOWING IS INVALID
+
+# This defines the value of fruit.apple to be an integer.
 fruit.apple = 1
+
+# But then this treats fruit.apple like it's a table.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -243,13 +243,21 @@ As long as a key hasn't been directly defined, you may still write to it and
 to names within it.
 
 ```
+# This makes the key "fruit" into a table.
```
ChristianSi: Maybe, to be more precise:

> This makes the keys fruit and fruit.apple into tables.

Or even:

> This makes the keys fruit and fruit.apple into tables (one nested inside the other).

eksortso: Thanks, but for this one example, I want focus to be on the key `fruit`. True that `fruit.apple` is also made to be a table. But I don't want to discuss how to define tables as much as I want to show how keys' value types remain the same.

pradyunsg:
Would it make sense to instead show the JSON-equivalent here?

ChristianSi:
@pradyunsg I don't think so.

eksortso:
@pradyunsg I'm a little confused. How would the JSON equivalent be useful here?

pradyunsg:
@eksortso I Sorry for the confusion, I might've been a little too tired that day. :)

pradyunsg:
And right now, as well, apparently.

Issue #622, opened 2019-05-17
Is it good to add short escaping alias \s for space (\u0020)?


LongTengDao:
To avoid spaces at line end in multi-line string being auto trimmed accidentally, and simplify `\u0020` writting.

eksortso:
I'm a little fond of this idea. It makes spaces explicit. But it's definitely new syntax.

And mentally, the notion conflicts with the regular-expression character class "`\s`", which is used to match any sort of whitespace, and not just `\u0020`. Different use cases for these, of course.

Is there a formulation of this bit of syntax that doesn't conflict with past usages anywhere?

LongTengDao:
Or define a line end wrapper character which not output?

```
a = """
There are\s
two ways \# even comment
to write space at line end."""
```

```json
{ "a": "There are \ntwo ways \nto write space at line end." }
```

pradyunsg:
Marking as post-1.0 since I don't really have the bandwidth to think about the edge cases here and this isn't super important to get to 1.0.

LongTengDao:
> Marking as post-1.0 since I don't really have the bandwidth to think about the edge cases here and this isn't super important to get to 1.0.

Sure, focus on getting 1.0 things done first, best wishes!

ChristianSi:
Even if this is post-1.0, it was supposedly meant to stay open until then?

eksortso:
@LongTengDao, since this issue is already tagged post-1.0, could you please reopen it? We can effectively ignore it until after 1.0 is released, as I think the issue still bears some examination. You could even argue to remove that tag if you think it's merited.

pradyunsg:
I can reopen it now but if it's closed again by OP, I won't be able to reopen.

mmakaay:
Quite confusing if I use regexps in my toml docs (which I do), and having to escape \s as \\s. For other escaped chars it's kind of natural to do so since they are generally used well known escapes. The \s to represent a space would be rather unique to TOML (as far as my knowledge goes), which might cause people to trip over it.

```
my.regexp = "something\\sfeels\sfishy"
```

ChristianSi:
@mmakaay: For regexes, literal strings might, in any case, be a better fit.

LongTengDao:
@ChristianSi @eksortso @pradyunsg Sorry, I see.

@mmakaay `"\s"` is also invalid now in TOML, you need always write `"\\s"` or `'\s'`.

But the different meaning of `\s` in regExp is a problem indeed.

[...]

PR #709, opened 2020-02-21
Include an \xHH escape code sequence

eksortso:
Per a conversation in #622, this PR adds a two-character `\x` escape code sequence for the first 256 Unicode code points. This provides a short explicit sequence for, among other code points: ASCII characters, certain control characters, and whitespace characters such as tab (`\x09`) and space (`\x20`) that need visible representation without resorting to longer `\u` and `\U` codes. Such sequences are commonly found in languages that also support the `\u` and `\U` codes for Unicode.


ChristianSi:
Looks good, I'm all for merging!

pradyunsg:
Hmm... this seems to have a slightly messed up history here.

eksortso:
I hope I didn't mess anything up. I was trying to bring it up to date. It's still for after v1.0.0, so if a new PR is in order, I can replace this one when the time is right.

Ekkarat30:
Excellent to get the writing access.

eksortso:
Since the history on this PR got messed up, I am replacing it with a new one, #796. The changes to the text (from when it was branched) should be identical to the changes done here.

Issue #725, opened 2020-04-12
why multiline basic string can ends with " inside now, while multiline literal string still can't ends with '?


LongTengDao:
I saw this is valid in v1.0.0-rc.1 now, which is different from v0.5.0:

```toml
str = """she said: "...""""
```

But what's the rule, how should I write the parser, and remember it while writing toml config file? Is this valid too?

```toml
str = """..."""""
```

even:

```toml
str = """...""""""
```

?

If they are all valid, then why this is invalid?

```toml
str = '''15: ''''''''''''''''''
```

I saw "You can write 1 or 2 single quotes anywhere within a multi-line literal string, but sequences of three or more single quotes are not permitted." in the spec.

Does this mean that, both multiline basic/literal can include ending quote marks, only if they are shorter than three? If so, I don't think current doc said it clear, neither explains nor examples.

PR is easy, but need to make sure at first.

marzer:
It means that the first time the sequence `'''` is seen inside a multi-line literal string it terminates the string, so for practical purposes you can't have three of them consecutively unless you intend to end the string.

You're right to point out that it's not the same as regular ML strings, though. I have no idea why a distinction is made between the two on this point. @pradyunsg, any input here?

eksortso:
Since I wrote the original PR, I'll do my best to explain. If we do need better examples than the ones in the README, then go ahead and post them.

Sorry @marzer, but triple quotes don't immediately end a string if there are quotes at the end of the string. I know that was the original plan. But it's slightly different now.

Sequences of one or two `'`s are allowed within multiline literal strings. What that means, by extension, is that a single set `'''` is not necessarily the end of the string, because 1 or 2 quotes are allowed at the end. So a series of 3, 4, or 5 quotes terminates a MLL string that ends with 0, 1, or 2 quotes, respectively. And actually, in MLBs, you could even have 6 in a row, if the first one is escaped:
```toml
this-in-quotes-plus-two-more = """"THIS\""""""
#                                 ^^^^^-^^^
```

@LongTengDao, let me point out the strings in your examples. I'll keep highlighting turned off, because as of now, GitHub's TOML highlighting is out of date.

```
# Example 1
str = """she said: "...""""
#        ^^^^^^^^^^^^^^^
# Four quote marks at the end;
# the first one is part of the string,
# and the last three terminate the string.

# Example 2 is VALID.
str = """..."""""
#        ^^^^^
Same rule: 5 quotes, 2 on the inside.

[...]

Issue #723, opened 2020-04-12
maybe mistake


LongTengDao:
https://github.com/toml-lang/toml/blame/master/README.md#L18

Maybe `non-` should be removed? Or it's correct in English expression?

---

BTW:
1. https://github.com/toml-lang/toml/blame/master/README.md#L795 has two spaces after period.
2. https://github.com/toml-lang/toml/blame/master/README.md#L794 the space after `key/value` is `\xA0`, not `\x20`.
3. https://github.com/toml-lang/toml/blame/master/README.md#L945 ` or of compatible type` should be removed, because now mixed type array is ok.

Since it's rc version now, fix them is not too late.

marzer:
The "non" is correct here; it's saying that the will only be changes between the RC and full release if something critical comes up.

For the other minor text formatting things, you could submit a pull request...

LongTengDao:
@marzer Does "critical change" mean "important change" and "break change" here?
But the whole sentence seems tried to tell people version 1.0.0-rc.1 is very stable, and won't have "important" nor "break" change with version 1.0.0 in the future?

pradyunsg:
+1 to all that @marzer said. :)

That language basically keeps the option to make a change open. I think it also clearly communicates that we don't plan to make any changes to the to the language's design without a very strong reason to do so, between now and 1.0.0.

LongTengDao:
I see, so this sentence means that 1.0.0 won't have big changes, except the change (problem) is too big, which we have to change. Thank you!

Issue #722, opened 2020-04-11
since there are v0.3.1 to fix incorrect example, should there also be v0.5.1 for same reason?


LongTengDao:
1. Wrong float example https://github.com/toml-lang/toml/blame/master/versions/en/toml-v0.5.0.md#L417 which is fixed in v1.0.0-rc.1.
2. And missed menu `MIME Type` item between line 42 and 43.

eksortso:
Go ahead and submit a PR to update the table of contents. We don't need a v0.5.1 for this. Worst case scenario, another release candidate could be released before the final v1.0.0. But for small corrections, another rc won't be necessary.

LongTengDao:
@eksortso I know what you mean, but I'm not sure if docs kind of spec could modify after release. If can, then why there was a v0.3.1, which was only to fix date example.

pradyunsg:
I don't think there's much value in releasing patch releases for the older version, when we're moving toward a major release. Plus, this isn't significant and the timeline of 0.3.0 vs 0.3.1 would be very different from 0.5.0 vs 0.5.1 if we make the release now.

Feel free to file a PR for adding the "Table of Contents" amendment. Let the mistake in the float example stay-as-is; it can stay that way forever.

So, to answer the question in the title with fewer words: no. But we can fix the "Table of Contents" mistake via a PR to 0.5.0.

PR #727, opened 2020-04-13
fix some pretty small problem

LongTengDao:
from #723 

1. https://github.com/toml-lang/toml/blame/master/README.md#L795 has two spaces after period.
2. https://github.com/toml-lang/toml/blame/master/README.md#L794 the space after key/value is \xA0, not \x20.
3. https://github.com/toml-lang/toml/blame/master/README.md#L945 or of compatible type should be removed, because now mixed type array is ok.

I found these when I was translating docs word by word. Since it's rc version now, fixing them do more good than harm.

---

for #725 

Make it clear in example, that multiline literal string can also end with (single) quote less than three, like multiline basic string do now.

---

#722

### DIFF
--- a/README.md
+++ b/README.md
@@ -791,8 +791,8 @@ same form as key/value pairs in standard tables. All value types are allowed,
 including inline tables.
 
 Inline tables are intended to appear on a single line. A terminating comma (also
-called trailing comma) is not permitted after the last key/value pair in an
-inline table.  No newlines are allowed between the curly braces unless they are
+called trailing comma) is not permitted after the last key/value pair in an
+inline table. No newlines are allowed between the curly braces unless they are
 valid within a value. Even so, it is strongly discouraged to break an inline
 table onto multiples lines. If you find yourself gripped with this desire, it
 means you should be using standard tables.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -432,8 +432,8 @@ quot15 = '''Here are fifteen quotation marks: """""""""""""""'''
 # apos15 = '''Here are fifteen apostrophes: ''''''''''''''''''  # INVALID
 apos15 = "Here are fifteen apostrophes: '''''''''''''''"
 
-# 'That's still pointless', she said.
-str = ''''That's still pointless', she said.'''
+# 'That,' she said, 'is still pointless.'
+str = ''''That,' she said, 'is still pointless.''''
```
pradyunsg, line 436: This is for #725?

LongTengDao: Oh, yes, it's 725, 727 is this PR...

LongTengDao: I have updated the hash number in PR explain. (Translating 1.0.0 and updating my parser has worn me out and I'm a bit dazed now... But still so excited to see 1.0.0 come)

ChristianSi:
I cannot really see what's that whitespace change in L794, but looks okay in general.

LongTengDao:
> I cannot really see what's that whitespace change in L794, but looks okay in general.

Thanks for review! A quick way to comapre is to copy the space and compare that with `"\x20"` and `"\xA0"` in browser console. I found that because my editor show characters non-ASCII in 1em width, not 1en width.

PR #732, opened 2020-05-02
Polish up the README

ChristianSi:
* Break lines that are too long
* Clarify that EOF is allowed after key/value pair
* Fix grammar oddity
* Better explain what the "expected" integer range means
* Add a few hyphens
* Turn internal crossreference into hyperlink
* Avoid duplicate "expressed"

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,8 +322,8 @@ Any Unicode character may be escaped with the `\uXXXX` or `\UXXXXXXXX` forms.
 The escape codes must be valid Unicode [scalar
 values](http://unicode.org/glossary/#unicode_scalar_value).
 
-All other escape sequences not listed above are reserved; if they are used used,
-TOML should produce an error.
+All other escape sequences not listed above are reserved; if they are used, TOML
+should produce an error.
 
 Sometimes you need to express passages of text (e.g. translation files) or would
 like to break up a very long string into multiple lines. TOML makes this easy.
--- a/README.md
+++ b/README.md
@@ -488,8 +488,8 @@ oct2 = 0o755 # useful for Unix file permissions
 bin1 = 0b11010110
 ```
 
-TOML implementations should be able to losslessly handle at least all integers
-in the 64 bit (signed long) range (−9,223,372,036,854,775,808 to
+TOML implementations must be able to losslessly handle at least all integers in
+the 64 bit (signed long) range (−9,223,372,036,854,775,808 to
 9,223,372,036,854,775,807).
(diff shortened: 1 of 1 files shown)


ChristianSi:
@eksortso: You're misunderstanding about the "should" vs. "must". That was merely a clean-up commit within the PR itself. The actual change is from

> 64 bit (signed long) range expected

(slightly ambiguous and already a source of confusion in the past) to the much clearer

> TOML implementations must be able to losslessly handle at least all integers in the 64 bit (signed long) range

I don't think that's a content change at all, it merely makes the meaning of the non-RFC'd "expected" clearer. (The old wording meant: "all users may **expect** that a 64 bit range is supported" == "All implementations **must** fulfill this requirement so as not to violate this expectation".)

eksortso:
But expectation of a condition makes no claim on what to do when the condition doesn't hold up. I think you're reading a MUST into a weak specification that could be a SHOULD at the very least.

ChristianSi:
@eksortso I see that differently, I've always read "expected" in this spec as meaning a requirement rather than a mere recommendation. But I guess that'll be for others to decide. @mojombo might mean how that "expected" was originally intended, but I don't know if he's around.

Meanwhile you could strengthen your case by finding a TOML implementation that only supports 32-bit integers, silently converting all bigger numbers into ... well, whatever. I sincerely hope you won't find any! But if that's the case than at least we aren't breaking anything by making it clear that this is an requirement.

README.md
```diff
@@ -486,7 +488,8 @@ oct2 = 0o755 # useful for Unix file permissions
 bin1 = 0b11010110
 ```
 
-64 bit (signed long) range expected (−9,223,372,036,854,775,808 to
+TOML implementations must be able to losslessly handle at least all integers in
+the 64 bit (signed long) range (−9,223,372,036,854,775,808 to
 9,223,372,036,854,775,807).
```
pradyunsg, line 493: Would you be willing to drop this change, for now? We'll tackle this in a follow up, and I'd prefer to not hold up the entire PR over a single point of disagreement. :)

ChristianSi: Sure, I've dropped that change for now.

ChristianSi:
I've also pushed a small formatting change: hyphen lines under sections headers are now always as long as the header itself (in a few cases they were longer or shorter).

README.md
pradyunsg: Yay! Thank you! ^>^

PR #735, opened 2020-05-07
Explicitly state that keys are strings

### LINKED ISSUE #733: Clarification: same bare and quotted keys sematically equivalent?

```toml
a = 4
"a" = 5
```

eksortso:
Explicitly states that keys are actually strings used as names for values within tables. Includes some grammatical corrections and modifies examples for the sake of clarity and consistency. Closes #733.

Please judge the aesthetics of these changes along with their technical precision. If you can find ways to streamline the text used here, then by all means recommend them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,8 @@ are not permitted in comments.
 Key/Value Pair
 --------------
 
-The primary building block of a TOML document is the key/value pair.
+The primary building block of a TOML document is the key/value pair. A key is
+simply a string in a table, and a key/value pair defines that string's value.
 
 Keys are on the left of the equals sign and values are on the right. Whitespace
 is ignored around key names and values. The key, equals sign, and value must be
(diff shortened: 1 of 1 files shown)


eksortso:
I'll go ahead and close this PR. I may bring up some of the points elsewhere.

ChristianSi:
Pity! I think it would be good to have at least a short clarification of this issue.

eksortso:
Well, @ChristianSi, I don't mind, really. It was tough making the new statements flow with the existing text. That's because the README is really for TOML's users more than it is for technical writers, and the string-iness of TOML's keys is a finer technical detail than most users would care to be bothered with, which is why the issue wasn't raised earlier.

Now that [the compliance project is starting up](https://github.com/toml-lang/compliance/pull/1), we'll have a good and strict way to enforce this sort of technical detail.

PR #737, opened 2020-05-10
Unify the expression between versions

LongTengDao:


pradyunsg:
@LongTengDao Look like I missed this in my triage of the issue tracker prior to deploying toml.io -- could you please file a PR against https://github.com/toml-lang/toml.io for this?

PR #739, opened 2020-05-15
Replace "expected" by more precise terms

### LINKED ISSUE #736: Clarify what "expected" means
The spec uses the word "expected" several times, and, as the discussion in #732 shows, the meaning of this word is not quite clear. I would propose to make it clear that these expectations are actually requirements, and reword accordingly.

Specifically, in the definition of the Integer type, change

> 64 bit (signed long) range expected

to

> TOML implementations must be able to losslessly handle at least all
> integers in the 64 bit (signed long) range

In the sections on time-related types, change

> The precision of fractional seconds is implementation-specific, but at least
> millisecond precision is expected.

to

> The precision of fractional seconds is implementation-specific, but TOML
> implementations must support at least millisecond precision.

(This change needs to be done three times, in the sections on Offset Date-Time, Local Date-Time, and Local Time).

The first wording has been moved out of #732 because the issue has turned out to be controversial. So let's discuss it here!

ChristianSi:
* Replace the vague term "expected" with the RFC 2119-compatible terms "should", "must", "required"
* Clarify that an error must be thrown if an integer cannot be represented losslessly

Fixes #736.

### DIFF
--- a/README.md
+++ b/README.md
@@ -488,8 +488,10 @@ oct2 = 0o755 # useful for Unix file permissions
 bin1 = 0b11010110
 ```
 
-64 bit (signed long) range expected (−9,223,372,036,854,775,808 to
-9,223,372,036,854,775,807).
+Arbitrary 64-bit signed integers (from −2^63 to 2^63−1) should be accepted and
+handled losslessly. If the result of parsing a value with unlimited precision
+would result in an integer that cannot be represented losslessly, an error must
+be thrown.
 
 Float
 -----
@@ -574,9 +576,9 @@ time with a space character (as permitted by RFC 3339 section 5.6).
 odt4 = 1979-05-27 07:32:00Z
 ```
 
-The precision of fractional seconds is implementation-specific, but at least
-millisecond precision is expected. If the value contains greater precision than
-the implementation can support, the additional precision must be truncated, not
+Millisecond precision is required. Further precision of fractional seconds is
+implementation-specific. If the value contains greater precision than the
+implementation can support, the additional precision must be truncated, not
 rounded.
 
 Local Date-Time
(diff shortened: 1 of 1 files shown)


eksortso:
LGTM!

README.md
```diff
@@ -488,8 +488,10 @@ oct2 = 0o755 # useful for Unix file permissions
 bin1 = 0b11010110
 ```
 
-64 bit (signed long) range expected (−9,223,372,036,854,775,808 to
-9,223,372,036,854,775,807).
+Arbitrary 64-bit signed integers (from −2^63 to 2^63−1) should be accepted and
+handled losslessly. If the result of parsing a value with unlimited precision
+would result in an integer that cannot be represented losslessly, an error must
+be thrown.
```
pradyunsg, line 494: This feels like a very complex sentence -- I understand what this means, but it needed a couple of readings and conscious thinking to figure it out.

I think we can drop the "unlimited precision" part of this sentence without losing any of the main benefits. :)

```suggestion
handled losslessly. If an integer that cannot be represented losslessly,
an error must be thrown.
```

abelbraaksma: @pradyung I agree to an extend. Though as advocate for the devil, a very literal reading can then become: I parse an int with `Int64.TryParse` and get back `-2`. Fine, right?

Nope, because that function would give a 2's complement result, input could've been `2^63 + 2`. Hence I introduced the wording "with unlimited precision".

I would highly recommend that at some point we diverge into two documents: one for the general public, and one being the actual spec using precise spec wording. The second one intended for implementers, who are expected to be able to read such more strict documents.

pradyunsg: Hmmm... I'm pretty sure I'd consider `2^63 + 2` becoming `-2` as a case of "cannot be represented losslessly".

> I would highly recommend that at some point we diverge into two documents: one for the general public, and one being the actual spec using precise spec wording. The second one intended for implementers, who are expected to be able to read such more strict documents.

Perhaps? I understand the balance we're trying to strike here, and if it's definitely-not-possible, we should do this. I'm not convinced that it'd be a worthwhile exercise right now though. :)

pradyunsg: PS: I think the devil is pretty hunky-dory right now. It'll be fine if you don't advocate for them for a bit. :P

abelbraaksma: > I'm not convinced that it'd be a worthwhile exercise right now though. :)

Agreed, I didn't mean "right now", but more for a vNext ;)

ChristianSi:
@pradyunsg Your suggestion was not quite clear, but I've just pushed a commit (I think) fulfills it. Better now?

@abelbraaksma To me the rewritten sentence still sounds fairly unambiguous, though it's true that the old version was even more explicit.

pradyunsg:
I thought I had already!

PR #742, opened 2020-06-02
Clarify that keys are strings in README.md

### LINKED ISSUE #733: Clarification: same bare and quotted keys sematically equivalent?

eksortso:
Clarifies that keys are strings. Implies that a quoted key whose string value resolves to what amounts to a bare key is the same as that bare key. This is pending a test in toml-lang/compliance that makes this explicit. Closes #733.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,10 @@ Key/Value Pair
 
 The primary building block of a TOML document is the key/value pair.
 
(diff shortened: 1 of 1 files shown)


eksortso:
There's better than this.

PR #743, opened 2020-06-03
Update README.md: Clarify keys treated as strings

### LINKED ISSUE #733: Clarification: same bare and quotted keys sematically equivalent?

eksortso:
Clarifies that keys are treated as strings, and surrounding quotes are not included. Closes #733.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ first = "Tom" last = "Preston-Werner" # INVALID
 Keys
 ----
 
-A key may be either bare, quoted, or dotted.
+A key is treated as a string and may be either bare, quoted, or dotted.
 
 **Bare keys** may only contain ASCII letters, ASCII digits, underscores, and
 dashes (`A-Za-z0-9_-`). Note that bare keys are allowed to be composed of only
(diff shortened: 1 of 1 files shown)


ChristianSi:
Looks good to me!

README.md
```diff
@@ -157,7 +157,7 @@ first = "Tom" last = "Preston-Werner" # INVALID
 Keys
 ----
 
-A key may be either bare, quoted, or dotted.
+A key is treated as a string and may be either bare, quoted, or dotted.
```
pradyunsg, line 160: I *really* don't want us to be referencing a value type, that we describe as a much later in the specification to describe keys.

Further, I don't think this addresses the "is bare same as quoted" concern from the original issue.

pradyunsg:
#753 adds an example instead, to avoid trying to meld these words into something that fits. :)

README.md
eksortso: Good point. I can't say that I didn't think that talking about "strings" in this way may be confusing. What I was trying to convey was, if you've got a key in double-quotes, resolve it like you would a string value, in order to get the key's proper name. On reflection, that's not so obvious.

eksortso:
I prefer your approach (though check my comments) on #753. You've got this.

PR #751, opened 2020-06-23
Adopt a Code of Conduct

pradyunsg:


pradyunsg:
@toml-lang/core Anyone opposed to us doing this?

BurntSushi:
Does TOML really have a community beyond this issue tracker? Seems like overkill to me.

Otherwise, I'd recommend adopting the CoC from LLVM or FreeBSD instead.

pradyunsg:
> Does TOML really have a community beyond this issue tracker? Seems like overkill to me.

I think it doesn't hurt to explicitly set expectations around social engagement. Ideally, we won't need to use any of this, but setting expectations clearly can't hurt.

BurntSushi:
I'm not going to put up a strong objection, although its value is not clear to me. "It doesn't hurt" doesn't seem like a good reason to me either.

Again, I would urge you to reconsider using the conteibutor covenant's CoC. Please instead consider something like LLVM's: https://github.com/freebsd/core.10-public-docs/blob/master/CoC/llvm-based.md

pradyunsg:
Using LLVM's CoC sounds like a good idea -- I definitely agree that it's much better overall.

eksortso:
Looking them over, these CoC documents have positive aspects, and ones that would have attracted me to the project if I was unsure of myself. That's important.

Free BSD's and LLVM's CoC language, which share a common source, is more positive, more assertive of what it promotes. There's little mention who would mediate if serious conflict happened. It's implied that project leads hold this authority, and I'm okay with that if it does not overwhelm them. FreeBSD has a committee dedicated to these issues, which could be useful for mediation and advice.

So something closer to what FreeBSD has would suit us well,  I think.

marzer:
> I'm not going to put up a strong objection, although its value is not clear to me. "It doesn't hurt" doesn't seem like a good reason to me either.

Despite having almost 14 thousand stars the TOML repo has relatively little 'social' engagement (in the issue system, et cetera). Why do you suppose that is? Could just be relatively dry nature of a language spec, sure, but who's to say there aren't _also_ a number of people feeling unsafe to do so? The tech industry is somewhat notorious for being toxic at the best of times. If you're in a position to say that you have trouble seeing the value in a CoC, then you're likely not the intended beneficiary. Privilege etc. Taking whatever small measures available to deal with that sort of toxicity, even pre-emptively, can only be a good thing.

pradyunsg:
@BurntSushi Sorry for my terse responses earlier. I was completely in "triage mode" at that time, and should've realized that terse sentences weren't appropriate here.

FWIW, "Code of Conduct" is a term that's got a *lot* of baggage associated with it, making it a more sensitive matter than usual, and discussions on sensitive matters have a tendency to escalate even when no one acts maliciously. :)

---

> Does TOML really have a community beyond this issue tracker?

There's also a Wiki. ;)

More seriously though, my personal motivation to do this is because I think we should "formalize" expectations around interactions on this repository. (huh, [I see a pattern here](https://www.python.org/dev/peps/pep-0609/#abstract))

I've found that it helps to note down how people are expected to behave and set a clear baseline, *especially* on the internet. It helps "clear the fog" of implicit expectations in the interactions, which in case of CoCs, can ease concerns about others' behaviors for some people (as @eksortso mentioned) and can be a great tool for preventing and handling negative interactions (as @marzer mentioned).

In other words, I don't think anyone of us thinks that we should have individuals get harassed on this repository. So, I think we should clearly state the same. It certainly cannot hurt and is likely helpful, even if we won't be able to directly see the positive effects of stating these things.

---

The tricky part of this exercise, for me anyway, has been "what are these implicit expectations?". It's really tricky to word properly, and I'm certainly not the best person when it comes to things like this.

This was why I ended up using "a well-known base" here. As I've mentioned, I agree with @BurntSushi that a Django-style CoC would likely work better for us, since it's easier to read and I feel more confident in saying "this is how we operate here" with it, than the Contributor-Covenant-style CoC. I will note that while I'm 100% OK with changing the specific wording, I'm wary of doing so "just because" since we're building upon stuff that is already more than good-enough IMO.

I'll point out that I really like how well-written [Django's Conduct FAQ](https://www.djangoproject.com/conduct/faq/) is -- particularly "Why have you adopted a Code of Conduct?" and "Why do we need a Code of Conduct? Everyone knows not to be a jerk.". I think both of those answers apply to most projects with a CoC (including us).

---

I do think I've identified 2 gaps in the implicit expectations in this process though -- we don't have any clarity on "who to talk to" in case of bad conduct, and what the consequences of bad conduct would be.

Right now, in this PR, I'm basically saying the answer to this would be "Pradyun, on a dedicated email address he controls + whatever he thinks is appropriate to do". I think this is basically true today, and I've certainly been contacted over my email, in the past, by folks to tell me about conduct they've found concerning.

I am all ears to know if we're all OK with this. FWIW, this was what I'd intended to ask in my first comment mentioning @toml-lang/core -- evidently, I didn't actually state this explicitly. :)

BurntSushi:
@pradyunsg Thanks for elaborating. I guess given the discussion, and some comments already starting to get personal (which seem to imply it's impossible for someone to disagree, which I think is totally ridiculous), I'll need to clarify that I am not opposed to codes of conduct. I've been a moderator for official Rust community spaces for several years now, and my primary charge is enforcing [its Code of Conduct](https://www.rust-lang.org/policies/code-of-conduct). I think it has overall worked really well so far, and I am really happy to have the Code of Conduct in that space.

Despite my apparent support for CoCs in principle, this does not necessarily mean that I add them to every project I'm involved in. I maintain lots of projects, and because they have no established community around them (nor do I really want the burden of maintaining any such official community for those projects), I don't add a CoC. In practice, it has just not been necessary. I see CoCs as tools to be wielded. Sometimes they are appropriate and sometimes they are not. I can totally appreciate a default stance of leaning towards adding a CoC, since their costs aren't that great (depending on which CoC you use, I guess).

In any case, I haven't really perceived the TOML project as having a community per se, which is why I questioned whether a CoC was really necessary based on the community aspect. However, if you're actively looking to foster and grow a community, then I would certainly withdraw my objection, and would actually lend support toward adding a CoC. That is, I do not see a CoC as a goal unto itself, but rather, a means to an end. But building a community is a huge task, and moderating it probably shouldn't be left to a single person. So I think, at least to me, the question is not, "Should we adopt a CoC?" But rather, "Should we build a community around TOML?"

pradyunsg:
> moderator for official Rust community spaces for several years now, and my primary charge is enforcing [its Code of Conduct](https://www.rust-lang.org/policies/code-of-conduct)

Yup yup -- I'm personally well aware of your work in those spaces, and have seen some of the occasions where you've had to use this hat; albeit I'm basically an intermittent silent observer. This shared context contributed to my terseness earlier. ;)

> Despite my apparent support for CoCs in principle, this does not necessarily mean that I add them to every project I'm involved in.

Indeed, and likewise.

> I can totally appreciate a default stance of leaning towards adding a CoC, since their costs aren't that great (depending on which CoC you use, I guess). [snip] the question is not, "Should we adopt a CoC?" But rather, "Should we build a community around TOML?"

The way I see it, in this case, TOML is fairly "big" project in terms of the userbase of it, and it certainly doesn't have a large community. I'm not sure I personally intend to grow a community here, but I certainly think that it'll be good to have a clear CoC around -- as you said, the costs aren't that great and given the size of our userbase, I think it's still an overall benefit.

BurntSushi:
Okay, SGTM. I'm still mildly skeptical, but do not feel strongly about it at all. I do feel more strongly about using an LLVM/FreeBSD/Django-style CoC though (as opposed to the Contributor Covenant). Mostly for reasons that you've stated already.

eksortso:
Well, regarding the question of community, we can look at all the contributors we've seen come through. We do have our regulars. But some of the most dynamic conversations we've had, at least since the time I started watching, came from people who basically came and went, and we did try our best to engage with them. That needs to be encouraged.

This project is not so much pridelands as it is a watering hole, where all the animals come for nourishment, and peace between them is a necessity. Being a standard, it's a nexus that serves many projects, and those projects have civil expectations of both the regulars here and other passers-by.

A Code of Conduct is valuable even if the community is small. We are trying to build something useful and good, but the core members of the project and its contributors necessarily recognize in no uncertain terms that "outsiders" _per se_ have shaped this standard more than just a little bit. They'll always be welcomed, and a CoC is just a formal way for us to acknowledge that civility.

ChristianSi:
@pradyunsg If I understand you right, you'd now prefer a LLVM-style or similar CoC, so this PR itself is obsolete?

pradyunsg:
I'll be updating this PR. Haven't come around to it yet.

ChristianSi:
Looks like this has been forgotten.

bl-ue:
ping.  I'd like to see this come through.

pradyunsg:
Closing this out for now -- I'll pick this back up when I have the time to do so.

PR #757, opened 2020-06-23
Clarify what "quotation marks" means

### LINKED ISSUE #748: Clarify if basic strings must be wrapped in double-quotes specifically
I'm inferring from the spec's examples that basic strings must be wrapped in double quotes, but it never says so explicitly. Rather it just says "quotation marks" for basic strings and basic multi-line strings, while it says "single quotes" for literal strings.

"Quotation marks" and "single quotes" are not good contrast objects. If basic strings must be wrapped in double quotes, the spec should just say so. "Double quotes" and "single quotes" are the natural contrast objects, and the two types of quotes are normally distinguished in web development. Also, "quotation marks" is a cumbersome way to say quotes or double quotes.

(I didn't do a PR because I wasn't 100% certain that double quotes were required.)

pradyunsg:
Closes #748

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -240,9 +240,10 @@ String
 There are four ways to express strings: basic, multi-line basic, literal, and
 multi-line literal. All strings must contain only valid UTF-8 characters.
 
-**Basic strings** are surrounded by quotation marks (`"`). Any Unicode character may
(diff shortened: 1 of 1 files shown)


ChristianSi:
Looks good.

PR #758, opened 2020-06-23
Array of tables phrasing changes

### LINKED ISSUE #738: Clarification: Array of tables vs Array of inline tables
The question is: Do array of tables with the `[[table]]` syntax have the same meaning as inline array of tables ?

For example: do those two documents have the same value ?
```toml 
[[table]] 
a = 42
b = "test"

[[table]]
c = 4.2
```

And

```toml
table = [ { a = 42, b = "test" }, {c = 4.2} ]
```

I expect that they have the same value but a sentence in the specification mentioning that they are indeed the same seems a good idea. In particular this sentence from the specification is problematic: 

> The last type that has not yet been described is an array of tables

It is at the beginning of the "Arrays of table" and can be interpreted to imply that an array of table is a new type that wasn't defined before and in particular that it is not the same type as the type described in section "Arrays". It would be much better to describe it as a another syntax for an existing type rather than a new type. I suppose that in TOML history table of arrays were introduced before inline arrays and so that sentence is historical.

I wouldn't mention it normally but it seems that the only test suite listed on your wiki, i.e https://github.com/BurntSushi/toml-test, does a difference between double bracket array of tables and array of inline table. The former do not have the `{type="array", ...}` marker but the latter does, and so the two documents above are parsed differently in JSON according to that test suite.

Hopefully when https://github.com/toml-lang/compliance is operational that kind of problem won't exist anymore

pradyunsg:
Closes #738

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -789,7 +789,8 @@ type.name = "Nail"
 Array of Tables
 ---------------
 
-The last type that has not yet been described is an array of tables. These can
+The last syntax that has not yet been described allows writing arrays of tables.
+These can
 be expressed by using a table name in double brackets. Under that, and until the
 next table or EOF are the key/values of that table. Each table with the same
 double bracketed name will be an element in the array of tables. The tables are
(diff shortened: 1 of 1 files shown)


ChristianSi:
Looks good.

PR #760, opened 2020-07-03
Add invalid float examples

### LINKED ISSUE #759: Should floats allow omission of integer or fractional/exponent?
I recently did a conformance pass over [toml++](https://github.com/marzer/tomlplusplus) to ensure it passed any remaining test cases from both BurntSushi's and iarna's test suites, and was surprised to find that I wasn't passing some of the floating-point tests. Specifically, I was accepting some inputs that would be valid as literals in C and C++, but are apparently not in TOML. Excerpt from my test suite before I updated it:

```c++
// omitting leading integer part
parse_expected_value(FILE_LINE_ARGS, "     .1 "sv,       .1 );
parse_expected_value(FILE_LINE_ARGS, "    +.1 "sv,      +.1 );
parse_expected_value(FILE_LINE_ARGS, "    -.1 "sv,      -.1 );
parse_expected_value(FILE_LINE_ARGS, "   .1e1 "sv,     .1e1 );
parse_expected_value(FILE_LINE_ARGS, "  .1e+1 "sv,    .1e+1 );
parse_expected_value(FILE_LINE_ARGS, "  .1e-1 "sv,    .1e-1 );
parse_expected_value(FILE_LINE_ARGS, "  +.1e1 "sv,    +.1e1 );
parse_expected_value(FILE_LINE_ARGS, " +.1e+1 "sv,   +.1e+1 );
parse_expected_value(FILE_LINE_ARGS, " +.1e-1 "sv,   +.1e-1 );
parse_expected_value(FILE_LINE_ARGS, "  -.1e1 "sv,    -.1e1 );
parse_expected_value(FILE_LINE_ARGS, " -.1e+1 "sv,   -.1e+1 );
parse_expected_value(FILE_LINE_ARGS, " -.1e-1 "sv,   -.1e-1 );

// omitting trailing fractional part
parse_expected_value(FILE_LINE_ARGS, "     1. "sv,       1. );
parse_expected_value(FILE_LINE_ARGS, "    +1. "sv,      +1. );
parse_expected_value(FILE_LINE_ARGS, "    -1. "sv,      -1. );
parse_expected_value(FILE_LINE_ARGS, "   1.e1 "sv,     1.e1 );
parse_expected_value(FILE_LINE_ARGS, "  1.e+1 "sv,    1.e+1 );
parse_expected_value(FILE_LINE_ARGS, "  1.e-1 "sv,    1.e-1 );
parse_expected_value(FILE_LINE_ARGS, "  +1.e1 "sv,    +1.e1 );
parse_expected_value(FILE_LINE_ARGS, " +1.e+1 "sv,   +1.e+1 );
parse_expected_value(FILE_LINE_ARGS, " +1.e-1 "sv,   +1.e-1 );
parse_expected_value(FILE_LINE_ARGS, "  -1.e1 "sv,    -1.e1 );
parse_expected_value(FILE_LINE_ARGS, " -1.e+1 "sv,   -1.e+1 );
parse_expected_value(FILE_LINE_ARGS, " -1.e-1 "sv,   -1.e-1 );
```
All of those representations are valid in C and C++ (and I assume other languages, too), but not legal in TOML.

Upon re-reading the TOML spec I realized that it is fairly clear that the integer part is always required (it's never subject to an _'or'_), so my knee-jerk assumption that followed the same rules as C and C++ wasn't particularly clever. Despite that, prefixing or suffixing an integer with a '.' character is a common short-hand for expressing that the value is in fact a float of some sort, and until encountering these test case violations, my assumption was that it was allowed in TOML seemed like a fairly natural one.

I guess the question is: should these representations, or some subset of them, be legal? If the answer is ultimately 'No', I think it would be wise to add a few invalid examples to the spec doc to explicitly call them out.

marzer:
Clarifies 'floats' section to explicitly demonstrate that omission of integer or fractional+exponent parts isn't legal (closes #759).

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -477,6 +477,15 @@ An exponent part is an E (upper or lower case) followed by an integer part
 (which follows the same rules as decimal integer values but may include leading
 zeros).
 
+The decimal point, if used, must be surrounded by at least one digit on each side.
+
+```
+# INVALID FLOATS
+invalid_float_1 = .7
+invalid_float_2 = 7.
+invalid_float_3 = 3.e+20
+```
+
 Similar to integers, you may use underscores to enhance readability. Each
 underscore must be surrounded by at least one digit.
 


marzer:
I'm not married to the wording here. Being both clear and concise about this particular point leaves a lot of room for bike-shedding.

toml.md
```diff
@@ -454,7 +454,8 @@ Floats should be implemented as IEEE 754 binary64 values.
 A float consists of an integer part (which follows the same rules as decimal
 integer values) followed by a fractional part and/or an exponent part. If both a
 fractional part and exponent part are present, the fractional part must precede
-the exponent part.
+the exponent part. Omitting the integer part is invalid, as is omitting both
+of the fractional and exponent parts.
```
ChristianSi, line 458: I would drop this change. The new sentence just repeats what is already said clearly in the first sentence of the paragraph.

toml.md
```diff
@@ -469,6 +470,10 @@ flt6 = -2E-2
 
 # both
 flt7 = 6.626e-34
+
+# omissions
+# fltX = .7   # INVALID - missing the integer part
+# fltY = 7.   # INVALID - missing both fractional and exponent parts
```
ChristianSi, line 476: The second comment is misleading since `7.e2` would be equally invalid. Rather I would propose to move these invalid examples into a separate example preceded by an introductory sentence, approximately as follows (for clarity and consistency with other examples, I would NOT comment-out the invalid examples, but leave off the TOML syntax highlighting):

Note that the decimal dot, if used, must be surrounded by at least one digit on each side.

```
# DO NOT DO THIS
invalid_float_1 = .7
invalid_float_2 = 7.
invalid_float_3 = 3.e+20
```

ChristianSi:
Complementing my comments above I just noticed that the invalid floats example should be moved after the paragraphs explaining how fractional part and exponent part look like, that is in front of the sentence "Similar to integers, you may use...".

toml.md
marzer: I disagree. The fact that it isn't explicitly called out is what led me to file issue #759 in the first place.

toml.md
marzer: The placement and formatting of invalid examples throughout the whole spec needs to 'pick a lane', really. I Chose this way of representing them because it's used elsewhere in the document, but then so is the way you've chosen above.

The reason I chose to do them 'inline' was because the existing float examples were already given comment 'headings' demonstrating the different representations of floats, so it made sense that 'invalid' float representations should also get a little heading.

marzer: I also think the existing use of "DO NOT DO THIS" in the spec doc is weird and nasty, and shouldn't be in the document at all. Obviously that's out-of-scope for this PR, but it warrants visiting at some point methinks.

ChristianSi:
I still consider the first sentence added by the PR as redundant. If I say: "X consists of A followed by B and/or C", then, in my mind, this clearly means that, when talking about X, A will always be there, and at least one of B or C will too. What's unclear about that?

Admittedly saying the same thing twice is not so bad and teachers do it quite often – it may help with learning. So I'm not strictly opposed to adding the new sentence – but still feel that, since the first sentence of the same paragraph already says as much and since the new example (added by this PR) makes everything crystal-clear, it would be better to leave it out for the sake of brevity.

The rest of the PR is now fine; however, I would still propose to insert the new text ("Note that the decimal point..." and the following example) above the sentence "Similar to integers, you may use..." (leading to the flt8 example) rather than below it. Since the text before that explicitly speaks about the fractional and exponent parts it feels more logical to have the new text there; the underscores issue is something else again, hence it would prefer to see it afterwards.

marzer:
> I still consider the first sentence added by the PR as redundant. If I say: "X consists of A followed by B and/or C", then, in my mind, this clearly means that, when talking about X, A will always be there, and at least one of B or C will too. What's unclear about that?

It's not a hypothetical problem; I'll remind you again that I made this PR as a result of a _real_ experience I had personally. I've spent a lot of time reading this spec and still the nuance of that point was a surprise to me since I came in with a programming-oriented preconception of what a 'float' is and how it may be represented. My natural feeling was one of "oh I know this!" and I missed a small but significant detail because of it. This sort of mental hop-scotch phenomenon is not unique to me or this situation, mind you; you would have fallen victim to it countless times in your own life, as has everyone. If being a touch more explicit than _you_ feel is necessary helps solve this problem in this particular case going forward, then it's not redundant at all (insofar as redundancy implies waste).

> I would still propose to insert the new text ("Note that the decimal point..." and the following example) above the sentence "Similar to integers, you may use..." 

Sure, why not.

eksortso:
The redundancy doesn't bother me, except it's not technically correct. Leaving off both the fractional part and the exponent part is actually not invalid, but only because doing so makes the value an integer, not a float. Integers are valid TOML, but the consumer of the value could complain if it was expecting a float. So I'm not sure how to best rewrite that latter phrase.

marzer:
@eksortso `1.` is an example of leaving both of while not being an integer

eksortso:
> @eksortso `1.` is an example of leaving both off while not being an integer.

I've been using `toml.abnf` for the definition of floats,  and it explicitly includes the decimal point with the fractional part. Here's the section of that document that's currently relevant:

```abnf
float = float-int-part ( exp / frac [ exp ] )
float =/ special-float

float-int-part = dec-int
frac = decimal-point zero-prefixable-int
decimal-point = %x2E ; .
zero-prefixable-int = DIGIT *( DIGIT / underscore DIGIT )
```

Though I was aware that many languages allow literals with hanging decimal points for float values, I did consider it better form to require that a decimal point always include at least one digit, for the sake of readability and unobscured identification. That's why I went along with the conversation and didn't say much until now.

The ABNF has been moving towards being fully in line with the spec. I'm not sure where its official status stands at this point, but I think we're close to done, so that anything valid per `toml.md` is necessarily well-formed per `toml.abnf`.

marzer:
@eksortso Yup, I get all that. I'm trying to clearly, *explicitly* disambiguate the difference in what TOML considers a float versus what many other languages do. When I think "float" I think "a floating-point number as per IEEE754 and a thousand different programming languages", not "a floating-point number as TOML would have it", and I'll not be the first to make the mistake of conflating the two. The whole point of this PR is basically to make the point you just did in fewer words.

marzer:
@pradyunsg I've moved the example, as well as adjusting the wording to essentially combine the two and make it read as less formal in the process. Hopefully closer to what you were driving at.

Incidentally, in working on this I've noticed that the repository lacks a `.gitattributes` and `.editorconfig`, mainly because my git client and text editor on my windows machine were both initally trying to convert everything to CRLF. I'm happy to add those too if that would be welcome.

toml.md
```diff
@@ -477,6 +477,16 @@ An exponent part is an E (upper or lower case) followed by an integer part
 (which follows the same rules as decimal integer values but may include leading
 zeros).
 
+Note that unlike other languages, TOML does not allow abbreviation of floats;
+the decimal point, if used, must be surrounded by at least one digit on each side.
```
pradyunsg, line 481:
```suggestion
The decimal point, if used, must be surrounded by at least one digit on each side.
```

eksortso:
@marzer Please address the issue that I brought up. If both the fractional part (including the decimal point) and the exponent part are excluded, then what remains is the integer part, not invalid syntax. Could you please revise that part of your proposal?

marzer:
@eksortso Huh? Haven't I? The sentence you took issue with isn't in the PR any more.

eksortso:
My apologies, @marzer. I couldn't tell. I'm still getting used to the app, which is no excuse. Didn't mean to devalue your work, and I'm sorry.

pradyunsg:
Thanks @marzer for the PR, and thanks all for an interesting discussion. :)

Issue #750, opened 2020-06-22
TOML 1.0.0-rc2?


ChristianSi:
As v1.0.0-rc.1 was published more than 2.5 months ago and various issues have been discovered and (partially) addressed since than, I propose it's time to proceed to publishing v1.0.0-rc.2 – which might then, if we're lucky, be promoted to 1.0 Final after two more months or so.

Prior to tagging and announcing RC 2, I think the following actions are needed:

- Merge this PR:
  - [x] #743 (would fix #733)
- Decide on these open issues (close / mark them as "post-1.0"):
  - [x] #685
  - [x] #738
  - [x] #748
  - [x] #749
  - [x] #707

---

These issues should be addressed as well, but probably need not block a new RC:

* #585
* #668

marzer:
Re #668: I took a look at the `toml.tmbundle` repo and it looks like abandonware at this point. _All_ of their repos do, actually. Lots of PRs waiting to be merged with very little activity from the maintainers. Is this _really_ where github gets their syntax highlighting juju from??

pradyunsg:
@ChristianSi Thanks for filing this! I've taken the liberty of editing the issue description, to use it as a checklist.

I actually think that nearly all the changes flagged since 1.0.0-rc1 have been *extremely* trivial, and that we can just hop to 1.0 directly. The only blocker to us doing so is #585, which, well, is 1-weekend-of-work away from completion. :)

ChristianSi:
@pradyunsg Wow, that was a quicker and more thorough reaction than I could have hoped for!

Since the changes since RC 1 went beyond mere typofixes, my own inclination would be NOT to jump to 1.0 final immediately, but to release an RC 2 first. If nothing (or only typos or postponable things) comes up, it could be promoted to 1.0 final after (at least) 4 weeks. What do others think?

pradyunsg:
Hmm... I think that'll work well actually.

It's probably a good idea to run through do-a-release at least once before 1.0.0 (since that's changed with the new website) and I'll also get a bit more time to tackle #585. Consider me convinced that an 1.0.0-rc2 is a good idea. :)

ChristianSi:
What's the status of this? Of all the tracked issues, only #685 is still open, as far as I can tell. I believe that one just needs a maintainer decision, then it could be closed quickly and the RC released...

pradyunsg:
I'd like to add in #759 as well.

ChristianSi:
[...]

PR #765, opened 2020-08-07
Add changelog entries for recent changes

pradyunsg:
Should be the last thing needed for #750.


pradyunsg:
Merged as I'd like to not delay this further + limited free time on my end.

Issue #766, opened 2020-08-14
Can comments appear before commas in in-line arrays?


Validark:
> Arrays can span multiple lines. A terminating comma (also called trailing comma) is ok after the last value of the array. There can be an arbitrary number of newlines and comments before a value and before the closing bracket.

My question is this: Can comments appear before commas in in-line arrays?

```toml
integers3 = [
  1,
  2 # is this a valid toml document?
  , 3
]
```

According to the [ABNF file](https://github.com/toml-lang/toml/blob/master/toml.abnf), the above is not a valid toml document.

However, [tomlplusplus](https://github.com/marzer/tomlplusplus) parses the above document just fine: https://godbolt.org/z/f6MG38

marzer:
This is a good question. Given that I've recently tested toml++'s compliance against the two canonical TOML compliance test suites (BurntSushi's and iarna's), this implicitly means that any other parser passing those might still have a discrepancy here as well.

My _hope_ is that this part of the spec is just a case of unclear wording and this syntax should actually be allowed. As far as the parser is concerned it's much simpler to allow this than to prohibit it. It's also not unheard-of for people to use leading commas instead of trailing commas, as odd as it looks.

eksortso:
I'm inclined to agree with @marzer here. SQL coders tend to use preceding comments in select clauses, and I've written enough SQL to understand why.

The ABNF would be easy to change. Replace `ws` with `ws-comment-newline` on two lines, then shorten up the lines with further refinement. The language of the spec is oddly specific: comments are allowed before values, and before closing brackets. That ought to be changed so that comments (with newlines) are allowed between values, commas, and brackets

Validark:
I agree, every major language I'm aware of allows this. The spec likely just forgot to mention this case. Perhaps it should instead say something like:

"Arrays can span multiple lines. A terminating comma (also called trailing comma) is permitted after the last value of the array. Any number of newlines and comments may precede values, commas, and the closing bracket."

ChristianSi:
As it stands, the spec (both written as ABNF) is unambiguous: newlines and comments may appear before values and before the closing bracket, BUT NOT before commas.

This also means that this (odd, but occasionally used) style is forbidden even in the absence of comments:

```toml
# THIS IS (as of now) NOT A VALID TOML DOC!
values = [
    1
  , 2
  , 3
]
```

[...]

PR #767, opened 2020-08-16
Allow comments w/ newlines before commas in arrays

eksortso:
Per #766 this change updates the specification to allow comments with newlines before commas in arrays. Modifies toml.md, toml.abnf, and CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 
-*  Clarify that comments with newlines are allowed before commas in arrays.
+*  Clarify that comments and newlines are allowed before commas in arrays.
 
 ## 1.0.0-rc.2 / 2020-08-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 
-Nothing.
+*  Clarify that comments with newlines are allowed before commas in arrays.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## unreleased
 
-Nothing.
+*  Clarify that comments with newlines are allowed before commas in arrays.
```
marzer, line 5: This change isn't just comments with newlines, but comments _and/or_ newlines, right?

Nit-pick, I guess, since this is just the changelog.

eksortso: Oh I didn't check if newlines on their own would be allowed. They should be, since arrays permit them between values. Maybe we need to bring this up in #766.

eksortso: You're right. The change to the ABNF does allow newlines on their own. So both the `CHANGELOG.md` and `toml.md` require changes.

marzer: I think the changes in toml.md are already sufficient... it reads correctly to me anyways O_o

ChristianSi (approved):

ChristianSi:
Looks good to me.

pradyunsg (approved):

pradyunsg:
Thanks @eksortso for a fully fleshed out PR for this! ^>^

PR #780, opened 2020-10-26
Clarify indentation usage and remove indents from table examples

### LINKED ISSUE #744: Proposal: Reference shortcut for nesting tables
Consider the following exampe:

```toml
[servers]
 
  [servers.alpha]
  ip = "10.0.0.1"
  dc = "eqdc10"
   
  [servers.beta]
  ip = "10.0.0.2"
  dc = "eqdc10"
```

**Idea:** add a shortcut character to reference the outer table.

Example:
```toml
[servers]
 
  [#.alpha]
  ip = "10.0.0.1"
  dc = "eqdc10"
   
  [#.beta]
  ip = "10.0.0.2"
  dc = "eqdc10"
```

The character `#` is illustrative and may have to be reconsidered giving it's used already for comments.

The proposal also considers multiple levels:

```toml
[servers]
 
  [#.alpha]
  ip = "10.0.0.1"
  dc = "eqdc10"

[#.#.firewall]
inbound = "0.0.0.0/24" # this would become `servers.alpha.firewall.inbound`
outbound = "0.0.0.0/24" # similarly, `servers.alpha.firewall.outbount`

  [#.beta]
  ip = "10.0.0.2"
  dc = "eqdc10"
```

The multi-level indicator, e.g. `#.#.#...`, must match the current structure. For example, `#.#.#.#.foo` right below a single-level table `[bar]` must be considered invalid. IDEs and plugins for text editors should be able to provide tooltip to indicate the expanded name of the referenced tables, expanding the value.

The benefit of this feature is to simplify typing while keeping explicitly defined the level of the table, and also in reducing the chances of typos when nesting tables. Can also facilitate when copying/pasting snippets.

eksortso:
During discussion [in issue 744](https://github.com/toml-lang/toml/issues/744#issuecomment-706912461) and elsewhere, users coming from Python with no experience with INI file formats were confused by the presence of indentation in a few of the examples in `toml.md`. Indentation is essentially whitespace (except inside strings) and is discouraged before keys and table headers, but this wasn't being stated clearly.

This PR makes these facts clear about indentation. For keys and table headers in examples, all indentation was stripped, and indentation is clearly treated as whitespace and its use is discouraged. For consistency, similar language was applied to array values, though indentation was preserved and it isn't discouraged there. A few minor grammar corrections were also made.

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -172,12 +172,6 @@ fruit . flavor = "banana"   # same as fruit.flavor
 
 Indentation is treated as whitespace. Tabs and spaces before keys are ignored.
 
-```toml
-bad_fruit.name = "banana"
-  bad_fruit.color = "green"  # Indentation is VALID, BUT DISCOURAGED.
-bad_fruit.flavor = "blah"
-```
-
 Defining a key multiple times is invalid.
 
 ```
@@ -631,7 +625,7 @@ contributors = [
 Arrays can span multiple lines. A terminating comma (also called a trailing
 comma) is permitted after the last value of the array. Any number of newlines
 and comments may precede values, commas, and the closing bracket. Indentation
-between array values and commas is treated as whitespace, and is ignored.
+between array values and commas is treated as whitespace and ignored.
 
 ```toml
 integers2 = [
@@ -693,15 +687,6 @@ extraneous whitespace.
 ```
 
 Indentation is treated as whitespace. Tabs and spaces before keys are ignored.
-
-```toml
-[m]      # how to write subtables
-[m.n]    # use dotted keys to nest tables
-[m.n.o]  # best practice: write all headers flush-left
-
-[p]      # Indentation is VALID,
-  [p.q]    # BUT DISCOURAGED.
-```
 
 You don't need to specify all the super-tables if you don't want to. TOML knows
 how to do it for you.
--- a/toml.md
+++ b/toml.md
@@ -170,7 +170,7 @@ fruit. color = "yellow"    # same as fruit.color
 fruit . flavor = "banana"   # same as fruit.flavor
 ```
 
-Indentation is treated as whitespace. Tabs and spaces before keys are ignored.
+Indentation is treated as whitespace and ignored.
 
 Defining a key multiple times is invalid.
 
--- a/toml.md
+++ b/toml.md
@@ -171,7 +171,6 @@ fruit . flavor = "banana"   # same as fruit.flavor
 ```
 
 Indentation is treated as whitespace. Tabs and spaces before keys are ignored.
-You shouldn't use them. But if you must, use them sparingly.
 
 ```toml
 bad_fruit.name = "banana"
@@ -694,7 +693,6 @@ extraneous whitespace.
 ```
 
 Indentation is treated as whitespace. Tabs and spaces before keys are ignored.
-You shouldn't use them. But if you must, use them sparingly.
 
 ```toml
 [m]      # how to write subtables
--- a/toml.md
+++ b/toml.md
@@ -686,7 +686,7 @@ extraneous whitespace.
 [ j . "ʞ" . 'l' ]  # same as [j."ʞ".'l']
 ```
 
-Indentation is treated as whitespace. Tabs and spaces before keys are ignored.
+Indentation is treated as whitespace and ignored.
 
 You don't need to specify all the super-tables if you don't want to. TOML knows
 how to do it for you.
--- a/toml.md
+++ b/toml.md
@@ -631,8 +631,8 @@ contributors = [
 
 Arrays can span multiple lines. A terminating comma (also called a trailing
 comma) is permitted after the last value of the array. Any number of newlines
-and comments may precede values, commas, and the closing bracket. Indentation is
-treated as whitespace.
+and comments may precede values, commas, and the closing bracket. Indentation
+between array values and commas is treated as whitespace, and is ignored.
 
 ```toml
 integers2 = [
--- a/toml.md
+++ b/toml.md
@@ -161,8 +161,23 @@ In JSON land, that would give you the following structure:
 }
 ```
 
-Whitespace around dot-separated parts is ignored, however, best practice is to
+Whitespace around dot-separated parts is ignored. However, best practice is to
 not use any extraneous whitespace.
+
+```toml
+fruit.name = "banana"     # this is best practice
+fruit. color = "yellow"    # same as fruit.color
+fruit . flavor = "banana"   # same as fruit.flavor
+```
+
+Indentation is treated as whitespace. Tabs and spaces before keys are ignored.
+You shouldn't use them. But if you must, use them sparingly.
+
+```toml
+bad_fruit.name = "banana"
+  bad_fruit.color = "green"  # Indentation is VALID, BUT DISCOURAGED.
+bad_fruit.flavor = "blah"
+```
 
 Defining a key multiple times is invalid.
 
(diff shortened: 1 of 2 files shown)


toml.md
```diff
@@ -614,9 +629,10 @@ contributors = [
 ]
 ```
 
-Arrays can span multiple lines. A terminating comma (also called trailing comma)
-is permitted after the last value of the array. Any number of newlines and
-comments may precede values, commas, and the closing bracket.
+Arrays can span multiple lines. A terminating comma (also called a trailing
+comma) is permitted after the last value of the array. Any number of newlines
+and comments may precede values, commas, and the closing bracket. Indentation is
+treated as whitespace.
```
marzer, line 635: > Indentation is treated as whitespace.

Will that be sufficient to eliminate confusion for the Python/YAML folk, given their relationship with "whitespace"? Suggest it include the word "ignored" somewhere.

marzer (changes requested):
My gut feel is that a blanket declaration like "Whitespace is insignificant and ignored everywhere except inside strings" somewhere near the top of the doc (perhaps the `Spec` section) might be more concise than the spec having to explicitly state that whitespace is ignored. A single source of truth, etc.

This PR is maybe not the right place to flesh that out, though.

toml.md
eksortso: Good idea. I'll add "ignored" to the language here.

eksortso: Whoops, I missed your previous comment. But since we're introducing this clarification for pedantic reasons, the duplication language is essential. We could come back in the future and sort-of refactor, but not just yet.

marzer: Roger that

(incidentally you saw my notes in the intended order; my inline comment was actually first and the general commentary on the duplication was the final "summary" comment at the bottom of the review, but github displays them like this after-the-fact, which is super weird)

marzer (approved):

toml.md
```diff
@@ -161,9 +161,24 @@ In JSON land, that would give you the following structure:
...
+Whitespace around dot-separated parts is ignored. However, best practice is to
 not use any extraneous whitespace.
 
+```toml
+fruit.name = "banana"     # this is best practice
+fruit. color = "yellow"    # same as fruit.color
+fruit . flavor = "banana"   # same as fruit.flavor
+```
+
+Indentation is treated as whitespace. Tabs and spaces before keys are ignored.
```
ChristianSi, line 173: I would change the first sentence to "Indentation is ignored" – indentation IS whitespace (even in Python), so otherwise the sentence is just repeating obvious.

toml.md
```diff
@@ -161,9 +161,24 @@ In JSON land, that would give you the following structure:
...
+You shouldn't use them. But if you must, use them sparingly.
```
ChristianSi, line 174: I would delete this line (and similar insertions elsewhere in the PR). Whether and how to indent is a stylistic issue and we as spec writers should not take a line here. OUR style in the spec should be not to indent – I agree on that. But if others want to use indentation and consider is good for readability, who are we to prevent/discourage them?

ChristianSi (changes requested):

toml.md
eksortso: It may be repeating the obvious to us, but not necessarily to Pythonistas and YAML punks. As I mentioned to @marzer, the repetition is deliberate. It's whitespace, it's ignored. Those two words always come together.

Call it overkill, but I'd rather have that in there now, going into v1.0.0, than leave any glimmer of doubt about what indentation is in TOML.

toml.md
eksortso: That line was a holdover from a previous draft. I just forgot to clean it up. I'll do that now.

toml.md
```diff
@@ -161,9 +161,23 @@ In JSON land, that would give you the following structure:
...
+fruit.name = "banana"     # this is best practice
+fruit. color = "yellow"    # same as fruit.color
+fruit . flavor = "banana"   # same as fruit.flavor
+```
+
+Indentation is treated as whitespace. Tabs and spaces before keys are ignored.
+
+```toml
+bad_fruit.name = "banana"
+  bad_fruit.color = "green"  # Indentation is VALID, BUT DISCOURAGED.
```
ChristianSi, line 177: There is still a DISCOURAGED here, as well as in the CHANGELOG (twice). I certainly agree that the indentation shown in this example would be nonsensical, but indenting all key-value pairs two spaces (say) or indenting whole nested tables depending on the number of name parts they contain seem reasonable stylistic choices.

toml.md
```diff
@@ -614,9 +628,10 @@ contributors = [
 ]
 ```
 
-Arrays can span multiple lines. A terminating comma (also called trailing comma)
-is permitted after the last value of the array. Any number of newlines and
-comments may precede values, commas, and the closing bracket.
+Arrays can span multiple lines. A terminating comma (also called a trailing
+comma) is permitted after the last value of the array. Any number of newlines
+and comments may precede values, commas, and the closing bracket. Indentation
+between array values and commas is treated as whitespace, and is ignored.
```
ChristianSi, line 634: This can be simplified to "... treated as whitespace and ignored."

toml.md
```diff
@@ -677,6 +692,17 @@ extraneous whitespace.
...
 
+Indentation is treated as whitespace. Tabs and spaces before keys are ignored.
+
+```toml
+[m]      # how to write subtables
+[m.n]    # use dotted keys to nest tables
+[m.n.o]  # best practice: write all headers flush-left
+
+[p]      # Indentation is VALID,
+  [p.q]    # BUT DISCOURAGED.
```
ChristianSi, line 703: I would remove this whole example, since it gives only stylistic advice. Not our cup of tea!

ChristianSi (changes requested):

toml.md
eksortso: I do agree with you, to an extent. But when the indentation doesn't match the depth of the specified subtable, there could be chaos.

I'll take out both of those commentaries. Not sure what to replace them with, though.

toml.md
eksortso: Sounds good.

toml.md
eksortso: Alright.

toml.md
eksortso: Replaced them with nothing. No need for examples.

ChristianSi (approved):

eksortso:
It's been a while. @pradyunsg your thoughts on this?

toml.md
```diff
@@ -677,6 +686,8 @@ extraneous whitespace.
 [ j . "ʞ" . 'l' ]  # same as [j."ʞ".'l']
 ```
 
+Indentation is treated as whitespace. Tabs and spaces before keys are ignored.
```
pradyunsg, line 689:
```suggestion
Indentation is treated as whitespace. Whitespace before keys is ignored.
```

toml.md
```diff
@@ -161,9 +161,17 @@ In JSON land, that would give you the following structure:
...
+Whitespace around dot-separated parts is ignored. However, best practice is to
 not use any extraneous whitespace.
 
+```toml
+fruit.name = "banana"     # this is best practice
+fruit. color = "yellow"    # same as fruit.color
+fruit . flavor = "banana"   # same as fruit.flavor
+```
+
+Indentation is treated as whitespace. Tabs and spaces before keys are ignored.
```
pradyunsg, line 173:
```suggestion
Indentation is treated as whitespace. Whitespace before keys is ignored.
```

pradyunsg: Or even:
```suggestion
Indentation is treated as whitespace and ignored.
```

pradyunsg (commented):
Looks great, although I would prefer to say "whitespace" instead of "tabs and spaces".

toml.md
pradyunsg: Or even:
```suggestion
Indentation is treated as whitespace and ignored.
```

toml.md
eksortso: I picked your second suggestion. It keeps the note about indentation and only repeats that.

eksortso: The same redundancy appeared under the dotted keys section as well, so I applied the same change there.

pradyunsg:
Hurray! Thanks! ^>^

PR #789, opened 2020-11-23
Clarify and describe the top-level table

eksortso:
This PR contains a paragraph and new example in the Table section of `toml.md` that explicitly describes the top-level table. The text and the example makes it clear that this table is nameless and cannot be relocated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## unreleased
 
+* Clarify and describe the top-level table.
 * Clarify that indentation before keys is ignored.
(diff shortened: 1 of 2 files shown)


ChristianSi:
I would move the whole new block further down, probable right above "Dotted keys define everything". Right now in comes in the middle of explaining how tables work, which is a bit odd.

eksortso:
Good idea. I'll move the changes down now.

pradyunsg (approved):

pradyunsg:
Thank you @eksortso and @ChristianSi! :)

PR #790, opened 2020-11-24
Add shorthand for escape character

### LINKED ISSUE #715: Support \e ESC Escape Character
I'm excited to see the progress of #709, however one other escape that seems to really be missing is `\e` (`ESC`) which is `\x1b`.

For example, `echo -e '\e[31mfoo\e[0m'` on an ANSI compatible terminal will print `foo` in red.

I'm wondering how people here feel about adding this to TOML?

For what it's worth, the maintainers of [Alacritty](https://github.com/alacritty/alacritty) are considering switching from YAML to TOML, and being able to express terminal escape sequences easily would be very nice for us.

chrisduerr:
This adds a shorthand for the escape character `\e` (0x1B).

This is useful especially when talking about things like terminal escape
sequences, which use this character regularly.

Fixes #715.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Clarify that indentation before keys is ignored.
 * Clarify that indentation before table headers is ignored.
 * Clarify that indentation between array values is ignored.
+* Add new `\e` shorthand for the escape character.
 
 ## 1.0.0-rc.3 / 2020-10-07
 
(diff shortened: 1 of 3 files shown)


chrisduerr:
I'm aware that the inclusion of this shorthand is not agreed upon yet. But I thought I might as well just send a PR right away so it doesn't just sit around as an issue nobody cares about.

Feel free to just close this if the decision has been made against supporting it. But I think it would be a nice addition.

eksortso:
If it does pass muster (and I personally support its inclusion), it probably won't come in before v1.0.0 is released.

chrisduerr:
1.0 has been released, so an evaluation of this feature (and the related issue) might be relevant now?

pradyunsg:
Thanks for filing a complete PR for this @chrisduerr! :)

PR #798, opened 2020-12-25
ABNF: Use uppercase hex digit consistently

dduan:
Although lowercase is not technically wrong, this document uses
uppercase hex digit in majority. So make that consistent.


pradyunsg (approved):

pradyunsg:
Thanks!

PR #801, opened 2021-01-01
ABNF: Reflect first newline in multiline string explicitly

dduan:
The spec states that, for both multiline basic and literal strings:

> A newline immediately following the opening delimiter will be trimmed.

This is currently not reflected in the ABNF definition, using basic
string as an example:

```abnf
ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
ml-basic-string-delim = 3quotation-mark
ml-basic-body = *mlb-content *( mlb-quotes 1*mlb-content ) [ mlb-quotes ]
```

Implemented directly (or via an ABNF parser generator), the initial new line 
sequence after `"""` will be part of `ml-basic-body`. This is misleading.

IMHO, this special newline should have its own place in the ABNF syntax.
Hence this patch.

### DIFF
--- a/toml.abnf
+++ b/toml.abnf
@@ -83,7 +83,8 @@ escape-seq-char =/ %x55 8HEXDIG ; UXXXXXXXX            U+XXXXXXXX
 
 ;; Multiline Basic String
 
-ml-basic-string = ml-basic-string-delim ml-basic-body ml-basic-string-delim
+ml-basic-string = ml-basic-string-delim [ newline ] ml-basic-body
+                  ml-basic-string-delim
 ml-basic-string-delim = 3quotation-mark
 ml-basic-body = *mlb-content *( mlb-quotes 1*mlb-content ) [ mlb-quotes ]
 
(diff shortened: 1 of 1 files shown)


eksortso (commented):
LGTM, but the lines are running long. Could you split each of the lines into two lines and indent the second of each pair of lines?

Like this but using your text:
```abnf
part1 = first-half [newline]
        second-half
```

dduan:
@eksortso done

eksortso (approved):

pradyunsg (approved):
Thanks for the PR!

[skipping my thinking aloud, that comes down to: sure, why not] 🤷🏻‍♂️

cicakkobeng8p (commented):
Summit

cicakkobeng8p (commented):
Summit

cicakkobeng8p (commented):
Summit

genering0721-png (approved):

PR #805, opened 2021-01-23
Clarify sub-millisecond precision is allowed

aduh95:
"Millisecond precision is required" could be interpreted as a requirement for input to be valid TOML. This commit adds examples of valid TOML time values that uses tens of second to remove the ambiguity.

Refs: https://github.com/alexcrichton/toml-rs/pull/416
Co-Authored-By: Dave Ostroske <eksortso@users.noreply.github.com>

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -563,7 +563,8 @@ To unambiguously represent a specific instant in time, you may use an
 ```toml
 odt1 = 1979-05-27T07:32:00Z
 odt2 = 1979-05-27T00:32:00-07:00
-odt3 = 1979-05-27T00:32:00.999999-07:00
+odt3 = 1979-05-27T00:32:00.5-07:00
+odt4 = 1979-05-27T00:32:00.999999-07:00
 ```
 
 For the sake of readability, you may replace the T delimiter between date and
@@ -581,10 +582,9 @@ odt5 = 1979-05-27 07:32Z
 odt6 = 1979-05-27 07:32-07:00
 ```
 
-Millisecond precision is required. Further precision of fractional seconds is
-implementation-specific. If the value contains greater precision than the
-implementation can support, the additional precision must be truncated, not
-rounded.
+Implementations are required to support at least millisecond precision.
+Additional digits of precision may be specified, but if they exceed the
+supported precision then the extra digits must be truncated, not rounded.
 
 ## Local Date-Time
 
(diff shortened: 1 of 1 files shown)


ChristianSi:
I don't think this is necessary, but it certainly cannot hurt.

bl-ue (approved):
👌🏻

eksortso (approved):
LGTM

eksortso:
However, upon reflection, I think a little more explanation in the text is necessary. We currently state that "millisecond precision is required," but it's unclear that we are referring to the parsers' level of support for precision, not the users' precision, which should be arbitrary. Perhaps we ought to change the precision text to read as follows:

> Implementations should support at least millisecond precision. Additional digits of precision may be specified, but if they exceed the supported precision, then the extra digits must be _truncated_, not rounded.

Then, if it wouldn't be too much detail, we could indicate what the resulting values would look like by referring back to the examples, to make this crystal clear.

> For instance, a specified value of `00:32:00.999999` would be truncated to `00:32:00.999` when parsed with millisecond precision.

Thoughts?

ChristianSi:
I would write "must support" instead of "should support", otherwise this sounds good. But I would tend not to the include the truncation example – I think the spec is perfectly clear even without it.

Also keep in mind that we repeat this language three times (for each of the relevant date/time types), so brevity is important. I suppose you mean to include the truncation example just once. Sure, that would be possible, but I don't think it's necessary.

eksortso:
It was intended to be succinct; I wrote it knowing that it would be repeated. We don't need to show the truncation example, but I do insist that we put emphasis on the word _truncated_.

I didn't intend to relax the millisecond precision requirement, which I'm almost certain some implementations will do anyway. I was only trying to avoid saying "must" twice, and I forgot the phrases in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.html) that meant the same thing. We technically aren't using RFC 2119 now, but #870 may push us that way.

Let's return to the original wording somewhat for that first sentence.

> Implementations are required to support at least millisecond precision. Additional digits of precision may be specified, but if they exceed the supported precision, then the extra digits must be _truncated_, not rounded.

eksortso (changes requested):
[toml.md](https://github.com/toml-lang/toml/files/10007017/toml.md)
Here are the changes to clarify that the sub-millisecond precision is a requirement specifically for the implementations to support.

eksortso (approved):
Thank you @aduh95 for including the changes.

eksortso:
@pradyunsg This PR draws an important distinction and relieves some potential confusion. What are your thoughts?

ChristianSi (approved):
LGMT!

eksortso:
@pradyunsg This has been sitting dormant for four months. Does anything more need done?

eksortso:
@pradyunsg It's been six months. What are your thoughts on this? [Can we merge it for v1.1.0?](https://github.com/toml-lang/toml/issues/928#issuecomment-1294659512)

ChristianSi:
Pleeeeese!

PR #811, opened 2021-02-24
Update ports in README

ilyakam:
It doesn't make sense to have the same port `8001` appear twice in the array. I've updated the README with the `ports` array to be `[ 8000, 8001, 8002 ]`, presumably as it was originally intended.

For more on this topic, please see the conversation in https://github.com/toml-lang/toml.io/pull/37.


eksortso (approved):
LGTM, thanks!

PR #812, opened 2021-03-26
Correct ABNF to prohibit DEL in comments

yawkat:
The spec says:

> Control characters other than tab (U+0000 to U+0008, U+000A to U+001F, U+007F) are not permitted in comments.

non-eol did not exclude 7F (DEL) properly.


pradyunsg (approved):
Ah! Nice catch! ^>^

PR #818, opened 2021-04-24
Tell GitHub's linguist to ignore scripts

LongTengDao:


jaceppo (approved):

eksortso:
It's my first time looking at Linguist attributes. So based on [all I know](https://link.medium.com/2g6ikTb31gb), here's what this is.

If I'm understanding this correctly, this PR intends to declare to Github Linguist that the scripts are all vendored, so the project won't be confused for a regular Python project.

If so, then LGTM. The scripts technically aren't vendored, but the `toml-lang` folks could make new projects out of them, if that need ever arose, and could then vendor them in properly.

.gitattributes
```diff
@@ -0,0 +1 @@
+/scripts/* linguist-vendored=true
```
pradyunsg, line 1:
```suggestion
/scripts/* linguist-detectable=false
```

LongTengDao:
Oh, you are right, `linguist-detectable` is be more suitable than `linguist-vendored`!

pradyunsg:
Thanks @LongTengDao! ^.^

PR #833, opened 2021-07-04
Allow 0X, 0O, and 0B literals as capitals

arp242:
Currently only "0x42" is accepted as a valid hex literal; "0X42" is
invalid.

I'm not sure if there's a good reason to not accept capital letters?
This is something that already works in many programming languages (not
all though, for example Go doesn't) and I don't see any downsides. I
can't find any previous discussions on this.

This also matches with the datetime parsing where "t" and "z" are
explicitly accepted as well (i.e. "1970-01-01t00:00:00z").

Since ABNF is case-insensitive we can use the "x" string for this
instead of specifying the codepoints, which is also a wee bit clearer.

### DIFF
--- a/toml.abnf
+++ b/toml.abnf
@@ -124,9 +124,9 @@ digit1-9 = %x31-39                 ; 1-9
 digit0-7 = %x30-37                 ; 0-7
 digit0-1 = %x30-31                 ; 0-1
 
-hex-prefix = %x30.78               ; 0x
-oct-prefix = %x30.6F               ; 0o
-bin-prefix = %x30.62               ; 0b
+hex-prefix = %x30."x"              ; 0x, 0X
+oct-prefix = %x30."o"              ; 0o, 0O
+bin-prefix = %x30."b"              ; 0b, 0B
 
 dec-int = [ minus / plus ] unsigned-dec-int
 unsigned-dec-int = DIGIT / digit1-9 1*( DIGIT / underscore DIGIT )


eksortso:
When the prefixes were [first proposed](https://github.com/toml-lang/toml/issues/409#issuecomment-215333474), we chose lowercase-only to prevent confusion. An octal prefix of `0O` doesn't read clearly with some fonts. A similar case could be made against `0B` (`08`?) but not against `0X` in my opinion. The clarity that lowercase prefixes provide is frankly worth keeping. If it became optional instead of mandatory, I'd still keep them lowercase.

That being said, our tendency is that if we place technical restrictions at the start, then it's easier to loosen them later. Better to do that than to start with an irreversible design mistake.

I'm against this particular change, but since it's mostly harmless, I won't stand in its way.

arp242:
> That being said, our tendency is that if we place technical restrictions at the start, then it's easier to loosen them later. Better to do that than to start with an irreversible design mistake.

Yeah, that makes sense. And I missed that particular issue on this; there were a bunch of them 😅

It just came up on the toml-test repo and I figured I'd forward it here. To be honest I'm kind of apathetic about this chance myself: I don't really care. Maybe I should have just forwarded that person here for them to argue the case themselves – it's just that I didn't want to "this is how the TOML spec says it is, issue closed, not my problem" 😅

ChristianSi:
I don't think this is "mostly harmless" – rather, it could really cause confusion if one could write octal 42 as `0O42`. So, since the prefix `0O` **must** remain forbidden, for consistency's sake it is much better to forbid `0X` and `0B` as well.

eksortso:
> I don't think this is "mostly harmless" – rather, it could really cause confusion if one could write octal 42 as `0O42`.

Well, you read my mind, there. That indeed is confusing. But octal, being a limited use case yet substantial within that context, is the only thing standing in the way of allowing `0X` and `08`.

> So, since the prefix `0O` **must** remain forbidden, for consistency's sake it is much better to forbid `0X` and `0B` as well.

@ChristianSi You're right, and you have swayed me. Let's reject this PR.

@arp242 Thanks for offering this, but although there are languages that permit uppercase in the headers, it's also the case that those same languages, and plenty of others, allow lowercase, and I cannot think of one language that only allows uppercase.

pradyunsg:
Looks like there’s consensus already before I got time to come to this PR’s notification. :)

Thanks for the PR @arp242, and for the discussion here @eksortso and @ChristianSi! ^.^

Issue #837, opened 2021-07-22
Proposal: Allow carriage return as valid newline sequence


hukkin:
Related to some of the discussion in https://github.com/toml-lang/toml/issues/835

I'd like to propose adding CR (carriage return) character that is not followed by an LF character to the list of allowed newline sequences. This is how e.g. CommonMark [defines line ending](https://spec.commonmark.org/0.30/#line-ending).

The reasoning for the change is:
1. ~~Avoid [roundtrip instability](https://github.com/toml-lang/toml/issues/835#issuecomment-877755889) when an LF is prefixed by more than one CR characters.~~ EDIT: Not an issue as per `.abnf`
1. In Python (perhaps other languages?) the standard way of opening files supports something called [Universal Newline Support](https://www.python.org/dev/peps/pep-0278/), meaning that LF, CR, and CRLF are all normalised to LF. This normalisation should NOT be used with the current TOML spec because
    - ~~CR characters in strings will be converted to LFs when not allowed to do so.~~ EDIT: Not an issue as per `.abnf`
    - Files where CRs are used as newline sequence are parsed successfully even though they are invalid TOML (spec v1.0.0).

   As far as I know, all popular Python parsers and their documented API make the mistake of doing this incorrect newline normalisation.

Now for point 2, a valid counter-argument is to "just fix all Python parsers". :smile: However, perhaps other languages do similar normalisation by default or otherwise we agree that a spec change makes more sense here instead.

hukkin:
If I interpret the `.abnf` correctly, CRs not followed by an LF character are already prohibited in MLB strings, in which case point 1 in the original post is actually not an issue.

I think this should be clarified in the text spec though. Currently the text spec explicitly mentions that unescaped CRs can be used in MLBs.

eksortso:
There's no such thing as an "escaped" carriage return in TOML. There's a limited set of characters that can be escaped (with a backslash) in basic strings, and carriage returns aren't among them. (Even inside escaped end-of-line whitespace in MLBs, carriage returns are parts of newline sequences.) And other representations of carriage returns like `\r` and `\u000d` use the escape backslash but the sequences aren't considered "eacaped."

hukkin:
> But to be clear, there's no such thing as an "escaped" carriage return

Thanks for the correction.

FWIW, I currently think it may be best to reject this proposal in favor of the clarifying PR https://github.com/toml-lang/toml/pull/838

ethanhs:
[...]

PR #838, opened 2021-07-23
Clarify that CR not followed by LF is not allowed in an MLB

hukkin:
A possible fix to https://github.com/toml-lang/toml/issues/837

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -341,8 +341,8 @@ str3 = """\
 
 Any Unicode character may be used except those that must be escaped: backslash
 and the control characters other than tab, line feed, and carriage return
-(U+0000 to U+0008, U+000B, U+000C, U+000E to U+001F, U+007F). Unescaped
-carriage returns (U+000D) are only allowed as part of a newline sequence.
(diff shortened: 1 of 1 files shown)


toml.md
```diff
@@ -341,7 +341,8 @@ str3 = """\
 
 Any Unicode character may be used except those that must be escaped: backslash
 and the control characters other than tab, line feed, and carriage return
-(U+0000 to U+0008, U+000B, U+000C, U+000E to U+001F, U+007F).
+(U+0000 to U+0008, U+000B, U+000C, U+000E to U+001F, U+007F). Unescaped
+carriage returns (U+000D) are only allowed as part of a newline sequence.
```
hukkin, line 345:
```suggestion
(U+0000 to U+0008, U+000B, U+000C, U+000E to U+001F, U+007F). Carriage returns
(U+000D) are only allowed as part of a newline sequence.
```

pradyunsg:
I don't think this is the approach to take here. I'd rather permit lone `CR`s as a newline character.

PR #840, opened 2021-08-13
Allow lone CR as valid newline

hukkin:
Closes https://github.com/toml-lang/toml/issues/837

### DIFF
--- a/toml.abnf
+++ b/toml.abnf
@@ -30,6 +30,7 @@ wschar =/ %x09  ; Horizontal tab
 ;; Newline
 
 newline =  %x0A     ; LF
+newline =/ %x0D     ; CR
 newline =/ %x0D.0A  ; CRLF
 
 ;; Comment
--- a/toml.abnf
+++ b/toml.abnf
@@ -30,8 +30,8 @@ wschar =/ %x09  ; Horizontal tab
 ;; Newline
 
 newline =  %x0A     ; LF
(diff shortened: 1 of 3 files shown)


toml.abnf
```diff
@@ -31,6 +31,7 @@ wschar =/ %x09  ; Horizontal tab
 
 newline =  %x0A     ; LF
 newline =/ %x0D.0A  ; CRLF
+newline =/ %x0D     ; CR
```
hukkin, line 34: Not sure if there's a good way to say "followed by EOF or any codepoint except LF", or if that's even necessary.

ChristianSi: newline =/ %x0D [%0A]  ; CR or CRLF

That should work, I think?

hukkin: Oh, thanks, how did I miss that.

The ABNF spec doesn't seem to mention anything about the greediness of that expression, so wondering if a CRLF could in theory match two `newline`s instead of one? Note that I think the same problem may apply to current solution using alternatives, because AFAIK alternatives are unordered.

There is a place where this could potentially be an issue, and that is the single optional newline that prefixes MLB and MLL strings. That said, I'm not sure if it's necessary for the ABNF to explicitly express that newline in the way it currently does.

septatrix:
The general consensus from the referenced issue seems to be that this is not desired

pradyunsg:
Closing per #837.

Issue #835, opened 2021-07-10
Question: Can newlines in multi-line literals be normalized?


hukkin:
The spec only speaks of newline normalization in the context of multi-line basic strings and is in fact quite literal about the fact that "content between the [multi-line literal] delimiters is interpreted as-is without modification".

I'm not an expert on reading ABNF, but the ABNF spec says
```
mll-content = mll-char / newline
```
that is, it does not include `newline` in `mll-char` (multi-line literal char), making me think perhaps the normalization is, nonetheless, allowed in multi-line literals as well.

I'd much appreciate clarification/someone else's interpretation.

arp242:
What exactly do you mean with "normalized"?

ABNF only describes the syntax (that it, what characters are allowed), and doesn't really describe the semantics (i.e. how it's interpreted or processed).

hukkin:
Normalize as in convert CRLF to LF or vice versa.

arp242:
Oh right. That's basically just up to the implementation; from https://toml.io/en/v1.0.0 :

> TOML parsers should feel free to normalize newline to whatever makes sense for their platform.

eksortso:
@hukkin Good catch. Definitely a subtle point.

My personal take is that what matters most about newlines in MLL strings isn't how they appear in the document but how the resulting string represents them, for two reasons.

Primarily, aside from newline characters, the only control character permitted in an MLL string is the tab. So fine control over the special characters `\n` and `\r` ought to be better suited to basic string use, where they can expressed clearly.

Second, as noted, the precedent for newline normalization is set in the Multiline Basic string definition, which should be repeated in the MLL string definition:

```
TOML parsers should feel free to normalize newline to whatever makes sense for their platform.
```

I'm willing to draft a PR to set this straight.

hukkin:
> I'm willing to draft a PR to set this straight.

Would be great, thanks! Should we make the sentence I quoted in the first post (that says "as-is without modification" etc.) mention this exception?

Now that I have your attention :smile: I've got another question related to newline normalization: is a parser allowed to normalize more than once? I'll try to explain what this means with an example where I use `\r` and `\r` to denote literal CR and LF characters:

```
str = '''Here's a newline\r\r\n'''
```
A parser is allowed to convert the `\r\n` to `\n`. We can express the normalized string with the following TOML:
```
str = '''Here's a newline\r\n'''
```
And on second parse we're allowed to parse
```
str = '''Here's a newline\n'''
```

[...]

PR #842, opened 2021-08-15
Allow newline normalization in multi-line literal strings

hukkin:
Closes https://github.com/toml-lang/toml/issues/835

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -387,8 +387,11 @@ version of literal strings that solves this problem.
 
 **Multi-line literal strings** are surrounded by three single quotes on each
 side and allow newlines. Like literal strings, there is no escaping whatsoever.
-A newline immediately following the opening delimiter will be trimmed. All other
-content between the delimiters is interpreted as-is without modification.
+A newline immediately following the opening delimiter will be trimmed. TOML
+parsers must normalize newlines in the same manner as multi-line basic strings.
+
+All other content between the delimiters is interpreted as-is without
+modification.
 
 ```toml
 regex2 = '''I [dw]on't need \d{2} apples'''
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Add \x00 notation to basic strings.
 - Seconds in Date-Time and Time values are now optional.
 - Allow non-English scripts in unquoted (bare) keys
+- Clarify newline normalization in multi-line literal strings.
 
 ## 1.0.0 / 2021-01-11
 


toml.md
```diff
@@ -378,7 +378,8 @@ version of literal strings that solves this problem.
 **Multi-line literal strings** are surrounded by three single quotes on each
 side and allow newlines. Like literal strings, there is no escaping whatsoever.
 A newline immediately following the opening delimiter will be trimmed. All other
-content between the delimiters is interpreted as-is without modification.
+whitespace and newline characters remain intact. TOML parsers should feel free
+to normalize newline to whatever makes sense for their platform.
```
pradyunsg, line 382: How about the following?

```
**Multi-line literal strings** are surrounded by three single quotes on each
side and allow newlines. Like literal strings, there is no escaping whatsoever.
A newline immediately following the opening delimiter will be trimmed. Newline
normalisation is handled in the same manner as multi-line basic strings.

All other content between the delimiters is interpreted as-is without modification.
```

hukkin: Seems also good to me, however isn't the word "must" too strong here? Should be fine if a parser doesn't normalize?

Some context why I wrote the text the way I did:
- I reused sentences from multi-line basic string description to avoid any chance of discrepancy.
- I left out the "as-is without modification" sentence because it should be implied by "Like literal strings, there is no escaping whatsoever".

pradyunsg: > Should be fine if a parser doesn't normalize?

Yes.

If they don't normalise in multiline basic strings, then they must not normalise multiline literal strings either. And if they do normalise, they must normalise in both.

Doing it in one but not the other -- I can't think of any good reason to do that.

hukkin: Oh I see.

> Doing it in one but not the other -- I can't think of any good reason to do that.

Totally agreed. However, I'm not sure this is something that the spec needs to enforce. It's a small gotcha that an implementation could in theory get wrong. Since normalization is optional to begin with, it is still completely possible that the parsed content has mixed LFs and CRLFs (soon also CRs), so enforcing doesn't give much extra guarantees of the parsed output. Just sharing my thoughts, don't feel strongly at all about this :smile: 

The word "other" in the last sentence ("All other content...") used to reference the word "newline" ("A newline immediately..."). Now it's in another paragraph and there is an extra sentence in between so the meaning of "other" seems a lot more vague. Would it make sense to replace "other content" with "non-newline content" or similar expression? Assuming we want to keep the sentence, as I mentioned above, I feel it should already be implied.

eksortso:
>Would it make sense to replace "other content" with "non-newline content" or similar expression?

The language suggested by @pradyunsg as stated is good. Its use of "other content" is clear in the context of the previous paragraph.

toml.md
eksortso: >Would it make sense to replace "other content" with "non-newline content" or similar expression?

The language suggested by @pradyunsg as stated is good. Its use of "other content" is clear in the context of the previous paragraph.

pradyunsg:
Thanks for the initial PR and the discussion on this @hukkin! ^>^

PR #847, opened 2021-09-05
Remove redundant rules

jun-sheaf:


abelbraaksma:
I think this is a good change, it simplifies the parsing and hence the parse-tree a bit. Redefinition of existing rules as is currently the case doesn't really seem necessary (unless there was a clear reason to do so, i.e. for error reporting or something?). @pradyunsg, what do you think?

pradyunsg:
Seems like a good idea! :)

PR #848, opened 2021-09-06
Clarify where and how dotted keys define tables

### LINKED ISSUE #846: Can dotted keys insert into already-defined [tables]?
Had a user file a bug report for the way my parser handles the following two snippets:

#### Snippet A
```toml
[a.b.c]
  z = 9
[a]
  b.c.t = 9
```

#### Snippet B
```toml
[a.b.c.d]
  z = 9
[a]
  b.c.d.k.t = 8
```

And honestly after reading and re-reading the table and dotted key sections of the spec I'm not really any closer to deciding if these are legal. Currently my implementation treats them both as being illegal.

I can understand how a case might be made for Snippet B to be legal, since it's effectively just an alternate way of expressing this:
```toml
[a.b.c.d]
  z = 9
[a.b.c.d.k]
t = 8
```
Which is pretty harmless. Snippet A, on the other hand, just feels wrong to me. If you can inject **values** into already-defined `[tables]` (and thus spread their actual definition all throughout the document), what then is the point of table headers?

eksortso:
Removes and replaces language that could potentially allow dotted keys in a parent table of an existing table to inject key/value pairs into subtables of the existing table. Closes #846.

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -128,12 +128,13 @@ to use bare keys except when absolutely necessary.
 ```
 
 A bare key must be non-empty, but an empty quoted key is allowed (though
-discouraged).
+discouraged). You cannot use multi-line strings to define quoted keys.
 
 ```toml
-= "no key name"  # INVALID
-"" = "blank"     # VALID but discouraged
-'' = 'blank'     # VALID but discouraged
+= "no key name"           # INVALID
+"""key""" = "not allowed" # INVALID
+"" = "blank"              # VALID but discouraged
+'' = 'blank'              # VALID but discouraged
 ```
 
 **Dotted keys** are a sequence of bare or quoted keys joined with a dot. This


marzer:
Maybe an additional counter-example too? Both @hukkin and I would probably have benefited from one during implementations. One good example is always better than wording tweakery imo.

hukkin:
Are we accidentally banning something like the following snippet?
```toml
a.x = 0
a.y = 0
```

If dotted keys create and define tables unconditionally now, while still (quoting the spec)
> Since tables cannot be defined more than once

do we have the problem that line 2 defines table "a" for the second time, rendering the snippet illegal?

ChristianSi:
Yeah, I think that the "provided that" clause was indeed meant to cover the case mentioned by @hukkin : Two or more dotted keys given *in the same `[table]` block* are of course allowed to refer to the same table name without the second being considered an illegal "redefinition" of the table defined by the first one – otherwise dotted keys would be next to useless! So, the "provided that" clause is indeed useful and should stay, but to avoid ambiguities I would propose to extend it as follows:

> provided that such tables were not previously created *by another dotted key given under the same `[table]` header.*

Strictly speaking that should already be sufficient, but for clarity @eksortso 's new sentence is certainly a good idea, though I would suggest to change the wording slightly:

> Any such table must have all its key/value pairs defined under the *same* `[table]` header, or in the root table if defined before all headers.

(Emphasis is only used to change which words were added or changes – it's not meant to actually go in the spec.)

Adding some more examples of illegal stuff would be useful too – maybe based on Snippet B from #846.

pelletier:
Clarifying question on this change: how does this impact inline tables?

For example:

```
a = { b.c = 1, b.d = 2 }
```

Is this valid? Or is it considered invalid because `b.c` defines the table at `a.b`, and therefore `b.d` would re-define it, which would be an error?

ChristianSi:
@pelletier: Your example is valid. All keys within the same block jointly define tables. An inline table is a block, while outside of inline tables a block extends from one explicitly written `[table header]` to the next.

(Note that the spec doesn't talk about "blocks" – that's just my way of explaining it.)

toml.md
```diff
@@ -759,8 +759,9 @@ name = "Regina Dogman"
 member_since = 1999-08-04
 ```
 
-Dotted keys create and define a table for each key part before the last one,
-provided that such tables were not previously created.
+Dotted keys create and define a table for each key part before the last one. Any
+such table must have all its key/value pairs defined under the current `[table]`
+header, or in the root table if defined before all headers.
```
pelletier, line 764:
```suggestion
header, or in the root table if defined before all headers, or in one inline table.
```

@ChristianSi thank you! I think this change would have helped me. At that point I wonder if your concept of block should be defined in the spec, it seems a clearer way to express that behavior.

ChristianSi:
Why is this not yet merged, by the way?

eksortso:
I accidentally opened a PR on my master branch. I'll follow up with a new PR with the same changes.

PR #849, opened 2021-09-21
Clarify that multi-line strings are not allowed in key names

arp242:
There have been two cases recently where I had someone who was confused
about this.

This clarifies that multi-line strings are not allowed as key names for
the casual reader; the phrases "basic string" and "literal string" have
a specific meaning in the TOML spec, but that's not really obvious if
you're just looking up "what is the valid syntax for a key?" as the
rules are explained in a different section, and it requires some careful
reading.

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -128,12 +128,13 @@ to use bare keys except when absolutely necessary.
 ```
 
 A bare key must be non-empty, but an empty quoted key is allowed (though
-discouraged).
+discouraged). You cannot use multi-line strings to define quoted keys.
 
 ```toml
-= "no key name"  # INVALID
-"" = "blank"     # VALID but discouraged
-'' = 'blank'     # VALID but discouraged
+= "no key name"           # INVALID
+"""key""" = "not allowed" # INVALID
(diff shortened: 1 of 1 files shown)


toml.md
```diff
@@ -128,12 +128,13 @@ to use bare keys except when absolutely necessary.
 ```
 
 A bare key must be non-empty, but an empty quoted key is allowed (though
-discouraged).
+discouraged). You cannot use multi-line strings in a quoted key.
```
ChristianSi, line 131: I'd change this to one of: "You cannot use multi-line strings to make/write a quoted key."

ChristianSi (changes requested):

toml.md
arp242: Yeah, "in a quoted key" is perhaps not the best. After going back and forth on a few alternatives I ended up with:

> You cannot use multi-line strings to define quoted keys.

Whatdayathink?

marzer: Has my vote 👍🏻

pradyunsg (approved):

PR #859, opened 2021-11-18
Clarify where and how dotted keys define tables

### LINKED ISSUE #846: Can dotted keys insert into already-defined [tables]?

eksortso:
Removes and replaces language that could potentially allow dotted keys in a parent table of an existing table to inject key/value pairs into subtables of the existing table. Closes #846.

This PR replaces the previous PR #848.

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -759,8 +759,9 @@ name = "Regina Dogman"
 member_since = 1999-08-04
 ```
 
-Dotted keys create and define a table for each key part before the last one,
-provided that such tables were not previously created.
+Dotted keys create and define a table for each key part before the last one. Any
+such table must have all its key/value pairs defined under the current `[table]`
+header, or in the root table if defined before all headers.
 
 ```toml
 fruit.apple.color = "red"
--- a/toml.md
+++ b/toml.md
@@ -762,7 +762,7 @@ member_since = 1999-08-04
 
 Dotted keys create and define a table for each key part before the last one. Any
 such table must have all its key/value pairs defined under the current `[table]`
-header, or in the root table if defined before all headers.
+header, or in the root table if defined before all headers, or in one inline table.
 
 ```toml
 fruit.apple.color = "red"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 
-Nothing.
+* Clarify where and how dotted keys define tables.
 
 ## 1.0.0 / 2021-01-11
 


eksortso:
@pelletier Can you make the same suggested change here that you suggested on #848? That way, you'll be credited with the change.

@ChristianSi Once this comes together, do you want to expedite the merge?

toml.md
```diff
@@ -760,8 +760,9 @@ name = "Regina Dogman"
 member_since = 1999-08-04
 ```
 
-Dotted keys create and define a table for each key part before the last one,
-provided that such tables were not previously created.
+Dotted keys create and define a table for each key part before the last one. Any
+such table must have all its key/value pairs defined under the current `[table]`
+header, or in the root table if defined before all headers.
```
pelletier, line 765:
```suggestion
header, or in the root table if defined before all headers, or in one inline table.
```

ChristianSi (approved):

hukkin:
Should we still address the concern in https://github.com/toml-lang/toml/pull/848#issuecomment-913452090 (and possible solution in https://github.com/toml-lang/toml/pull/848#issuecomment-913559789) somehow?

ChristianSi:
@hukkin My opinion is that the language in this PR is clear enough and there is no need for further changes.

eksortso:
Agreed. A table can only be defined in one section using only one method. So writing both keys `a.b` and `a.c` under the same header defines the subtable `a` using dotted keys. The same applies inside an inline table.

moorereason:
Asking a clarifying question.  Given the follow TOML document, are the comments within it true?

```toml
# Given
a = {b.c = 0}

# VALID since the keys of "b" are not entirely defined by the "b.c" reference above
a.b.z = 0

# INVALID since the keys of "a" are defined entirely by the inline table
a.c = 0
[a.d]
[[a.e]]
```

hukkin:
@moorereason My understading is that both comments in your example document should say `INVALID`. Inline tables are fully defined within the braces, including any nested data structures.

ChristianSi:
@moorereason: Yes, both your examples are invalid, like @hukkin says.

```toml
# Given
a = {b.c = 0}   # a is completely defined as an inline table

# INVALID since the keys of "a" are defined entirely by the inline table
a.b.z = 0

# INVALID since the keys of "a" are defined entirely by the inline table
a.c = 0
[a.d]
[[a.e]]
```

Actually dotted keys aren't really relevant or needed for this example. Inline tables are assumed to be complete, hence, after your `a = {...}` line, nothing under `a.` can be redefined elsewhere.

eksortso:
@mojombo @pradyunsg Can you review this PR and check if it's ready to merge?

pradyunsg (approved):

PR #867, opened 2022-01-15
Clarify that CR not followed by LF is not allowed in an MLB

hukkin:
Reopening https://github.com/toml-lang/toml/pull/838 now that https://github.com/toml-lang/toml/issues/837 was rejected.

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -341,8 +341,8 @@ str3 = """\
 
 Any Unicode character may be used except those that must be escaped: backslash
 and the control characters other than tab, line feed, and carriage return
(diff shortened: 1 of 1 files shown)


ChristianSi (approved):

pradyunsg (approved):

ChristianSi:
Should a small note be added to the changelog?

Or is that not necessary because it's only a more explicit wording, not a change in how documents are parsed?

pradyunsg:
> Or is that not necessary because it's only a more explicit wording, not a change in how documents are parsed?

Pretty much this. :)

Issue #447, opened 2017-02-05
How do you pronounce "toml" when you are reading it out loud?  As in "Cargo.toml"


CleanCut:
I'm making an instructional video course for Rust, which uses .toml files in it's package management system.  When I'm saying "Cargo.toml" out loud to my learners, how should I pronounce .toml?

I'm assuming (and hoping) that there's "one right way" to say it, so we don't end up with perpetual gif-vs-jif-like wars.

BurntSushi:
tom-el

CleanCut:
Thanks!

PR #868, opened 2022-01-16
Add pronunciation to README

amirkarimi:
https://github.com/toml-lang/toml/issues/447

Not sure if that's the best place to put the pronounciation?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 TOML
 ====
 
-TOML (pronounced /tɒməl/) stands for Tom's Obvious, Minimal Language.
+TOML (usually pronounced /tɒməl/) stands for Tom's Obvious, Minimal Language.
 
 By Tom Preston-Werner, Pradyun Gedam, et al.
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <img align="right" src="logos/toml-200.png" alt="TOML logo">
 
-TOML
+TOML /tɑməl/
 ====
 
 Tom's Obvious, Minimal Language.
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 <img align="right" src="logos/toml-200.png" alt="TOML logo">
 
-TOML /tɑməl/
+TOML
 ====
 
-Tom's Obvious, Minimal Language.
+TOML (pronounced /tɒməl/) stands for Tom's Obvious, Minimal Language.
 
 By Tom Preston-Werner, Pradyun Gedam, et al.
 
--- a/toml.md
+++ b/toml.md
@@ -3,7 +3,7 @@
 TOML
 ====
 
-TOML (pronounced /tɒməl/) stands for Tom's Obvious, Minimal Language.
+TOML (usually pronounced /tɒməl/) stands for Tom's Obvious, Minimal Language.
 
 By Tom Preston-Werner, Pradyun Gedam, et al.
 
--- a/toml.md
+++ b/toml.md
@@ -3,7 +3,7 @@
 TOML
 ====
 
-Tom's Obvious, Minimal Language.
+TOML (pronounced /tɒməl/) stands for Tom's Obvious, Minimal Language.
 
 By Tom Preston-Werner, Pradyun Gedam, et al.
 


ChristianSi:
It certainly isn't.

And I wouldn't pronounce it like that – definitively not with an /ɑ/ as in *bath*.

amirkarimi:
My bad, probably copied from the wrong place. I guess the right one is `/ɒ/`?

And about the location, what about putting it in the first line like this:

> Tom's Obvious, Minimal Language (pronounced `/tɒməl/`).

Or

> TOML (pronounced `/tɒməl/`) stands for Tom's Obvious, Minimal Language.

marzer:
The second option gets my vote.

amirkarimi:
Apply a feedback

arp242:
IMO the concept there is *One True Correct Pronunciation™* is kind of misguided. It depends on your dialect, accent, preference, etc.

"You say potato, I say pot-ah-to". Or I guess the tech equivalent of that is "you say S-Q-L, I say sequel".

amirkarimi:
@arp242 I agree, but as a non-native English speaker I (and probably many) prefer to get a hint. Would be nice to know for example it can be called tom-el.

Maybe the text can be changed to something like "can be pronounced..." or "widely pronounced..." (Though I don't have enough data to back it up)?

marzer:
@arp242 Providing a disambiguation hint in service of avoiding "gif vs jif" scenarios seems harmless to me. There are enough silly hills people choose to die on in tech; removing one can't hurt.

ChristianSi:
We could change it into "*typically* pronounced /tɒməl/". Otherwise I'm fine with the proposal as it stands now and would consider it a useful addition.

ChristianSi (approved):

abelbraaksma:
Looks like this can be merged? At least now I know how to pronounce TOML ;).

pradyunsg:
> IMO the concept there is _One True Correct Pronunciation™_ is kind of misguided. It depends on your dialect, accent, preference, etc.

I agree. As an example, I pronounce it as /toməl/ -- which is not how this PR deems it should be pronounced. :)

Overall though, I don't think policing how people say/read TOML is particularly useful -- as long as everyone in the discussion understands what's being talked about, it's fine.

pradyunsg:
Anyway, long story short, I'm wary of merging a PR that guides people to do something that I personally certainly won't be doing. 🤷🏽‍♂️

amirkarimi:
So dictionaries have been doing it wrong all this time policing the pronunciation even for abbreviations like [ESOL](https://dictionary.cambridge.org/dictionary/english/esol)?

We even changed the wording to reflect the ideas in this thread.

I would never imagine such a small change would turn into this!

abelbraaksma:
I do think it makes sense to give a hint on pronunciation. Tee-Oh-Em-El or Tommel are two ways of pronouncing it. When I first encountered it, I was curious how people would pronounce it, so I wasn't the odd guy in the next Friday drink when bringing it up. A guide is just that, a guide. If people want to say it differently, fine, but at least we know how it was intended to be pronounced by the original creator.

Just my 2p. Also, people do ask this: https://en.m.wikipedia.org/wiki/Talk:TOML#Pronounciation_of_TOML.

ChristianSi:
I too think it would be good to re-open this and merge it (possibly after some modification, if needed). That even the people here disagree on how TOML is to be pronounced in my while indicates that such guidance would indeed be useful.

Of course, it would still possible it disregard it, which is of course perfectly fine. The text of the PR now has "usually pronounce" which leaves options for unusual pronunciations.

As for @pradyunsg's comment, I'm a bit intrigued by where you get the /o/ from? Since TOML is supposedly an English word, I would expect the *O* to be pronounced either "short" as in *hot* – which results in /tɒməl/, as this PR has it, or else long as in *old*, which would lead to /toʊməl/. (I'm using the common IPA conventions from [Wikipedia](https://en.wikipedia.org/wiki/Help:IPA/English#Key) here, local pronunciations will vary a bit.) If you write /o/, do you mean the "long O" as in *old*, or it is an altogether different sound for you?

If it's the long O, maybe some more discussions on whether the "short" or the "long" version should be favored (and recommended) might be in order.

PR #871, opened 2022-02-08
Update toml.abnf

LongTengDao:


marzer:
Aren't ABNFs case-insensitive? If so, the comment was correct already.

marzer:
Also, from RFC 3339:

> NOTE: Per [ABNF] and ISO8601, the "T" and "Z" characters in this
      syntax may alternatively be lower case "t" or "z" respectively.

LongTengDao:
> Aren't string literals in ABNFs case-insensitive? If so, the comment was correct already.

Oh! Thank you to figure out! Then my implementation didn't do right thing about `T` and `Z` from the beginning...

marzer:
@LongTengDao neither did mine until recently 😅

LongTengDao:
> @LongTengDao neither did mine until recently 😅

How did you notice that... T_T @marzer

marzer:
Good question, I don't remember. I _think_ it was one of the more recent tests added to the toml-test suite, but I can't be sure.

Issue #873, opened 2022-02-10
abnf grammar has incorrect line endings


bendyarm:
The file `toml.abnf` uses LF between lines.  
The abnf spec requires CRLF between lines:  
https://datatracker.ietf.org/doc/html/rfc5234#section-2.2  
(Some checkers and consumers of abnf grammars permit LF instead, but that is off-spec.)  
I will also submit a PR for this issue.

pradyunsg:
Fixed in #876.

PR #874, opened 2022-02-10
LF -> CRLF; add comment about grammar applying to decoded UTF-8; add …

bendyarm:
….gitattributes so CRLF stays
Fixes issue #873 .


bendyarm:
Closing since I meant to have only one commit in this PR.  Will open a new one.

arp242:
If you `git rebase -i master`, select "s" for all commits except the first one, and then force push it will just show up as one commit here.

bendyarm:
@arp242 thanks for the tip!

PR #875, opened 2022-02-11
clarify string descriptions

bendyarm:
Restructure the list of disallowed code points so the control characters that are disallowed
in all types of strings are listed at the beginning, and just provide the differences for the various types.  
Clarify that backslash and quotation mark can occur literally, but only as part of an escape sequence.
Also put newlines in a couple of other places where the lines exceeded 80 characters.

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -259,12 +259,20 @@ String
 ------
 
 There are four ways to express strings: basic, multi-line basic, literal, and
-multi-line literal. All strings must contain only valid UTF-8 characters.
+multi-line literal.
 
-**Basic strings** are surrounded by quotation marks (`"`). Any Unicode character
-may be used except those that must be escaped: quotation mark, backslash, and
-the control characters other than tab (U+0000 to U+0008, U+000A to U+001F,
-U+007F).
+All strings must contain only valid UTF-8 encoded characters as is the case for
+the TOML document as a whole.  Certain control characters are not allowed to
+occur literally in any kind of string: U+0000 to U+0008, U+000B, U+000C, U+000E
+to U+001F, and U+007F. In basic strings and multi-line basic strings, but not in
+literal strings or multi-line literal strings, those control characters can be
+described with escapes as specified below. Additional restrictions are described
+below.
+
+**Basic strings** are surrounded by quotation marks (`"`). In addition to the
+characters disallowed for all strings mentioned above, U+000A (LF) and U+000D
+(CR) may not occur literally in basic strings.  Backslash and quotation mark may
+only occur literally if they are part of a valid escape sequence.
 
 ```toml
 str = "I'm a string. \"You can quote me\". Name\tJos\u00E9\nLocation\tSF."
@@ -340,10 +348,10 @@ str3 = """\
        """
 ```
 
-Any Unicode character may be used except those that must be escaped: backslash
-and the control characters other than tab, line feed, and carriage return
-(U+0000 to U+0008, U+000B, U+000C, U+000E to U+001F, U+007F). Carriage returns
-(U+000D) are only allowed as part of a newline sequence.
+In addition to the characters disallowed for all strings mentioned above, U+000D
+(CR) is allowed only as part of a newline sequence U+000D U+000A (CRLF).  As
+with basic strings, backslash and quotation mark may only occur literally if
+they are part of a valid escape sequence.
 
 You can write a quotation mark, or two adjacent quotation marks, anywhere inside
 a multi-line basic string. They can also be written just inside the delimiters.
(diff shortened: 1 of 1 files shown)


marzer:
I don't believe these changes make the spec any any clearer. In fact I'd argue it's a clear regression in clarity; aggregating the string information for all strings and listing exceptions requires the reader to do the heavy lifting. Repeatedly listing it for each string type is redundant, yes, but expects less cognitive load from the reader and fits the flow of the document better.

toml.md
```diff
@@ -259,12 +259,20 @@ String
...
-may be used except those that must be escaped: quotation mark, backslash, and
-the control characters other than tab (U+0000 to U+0008, U+000A to U+001F,
-U+007F).
+All strings must contain only valid UTF-8 encoded characters as is the case for
+the TOML document as a whole.  Certain control characters are not allowed to
+occur literally in any kind of string: U+0000 to U+0008, U+000B, U+000C, U+000E
+to U+001F, and U+007F. In basic strings and multi-line basic strings, but not in
+literal strings or multi-line literal strings, those control characters can be
+described with escapes as specified below. Additional restrictions are described
+below.
```
marzer, line 270: If you do persist with this change, then I'd simplify this paragraph. There's just too much here.

```suggestion
Strings must contain only valid UTF-8 encoded characters. Certain control characters are not allowed to occur literally in any kind of string: U+0000 to U+0008, U+000B, U+000C, U+000E to U+001F, and U+007F.
```

The point about the basic strings supporting escaped control characters is already covered in the "basic strings" section.

(An argument can be made that the list of disallowed characters should be represented as an actual bullet-point list, though that's a matter of taste, and beyond the scope of this PR since it wasn't that way to begin with.)

marzer (changes requested):

toml.md
marzer: @arp242 I like that wording.

bendyarm: @arp242  Thanks, that is an improvement.  Putting those details at the beginning rather than at the end of each section makes all the sections easier to understand.  I have also reworded the last paragraph in strings to make it clear it is not a another part of the spec but just advice (paragraph starting "Because most control characters are not permitted...").

ChristianSi:
I would avoid the phrase "All strings must be encoded as valid UTF-8" since it suggests that strings can be encoded independently of the rest of a TOML document, which of course is not the case. The *whole document* is encoded as UTF-8; so, when looking at strings, we can only see them as series of Unicode codepoints, not of bytes. So, instead of

> "All strings must be encoded as valid UTF-8, and can contain any codepoint except ..."

I'd propose

> "Strings can contain any valid Unicode codepoint except ..."

bendyarm:
@ChristianSi  Good idea!  I also simplified the language to avoid the double negative, clarifying the situation with tab.  

This is what the paragraph looks like in this PR now:

> There are four ways to express strings: basic, multi-line basic, literal, and
> multi-line literal. Strings can contain any valid Unicode codepoint except the
> following control characters: U+0000 to U+0008, U+000A to U+001F, and
> U+007F. Note that tab (U+0009) is allowed. Multi-line strings can also contain
> newlines (U+000A) and carriage returns (U+000D).

@arp242  What do you think about this change to your rewording?  
@abravalheri Does this rewording fix the issue about tabs implicit in PR #878 ?

ChristianSi (approved):

toml.md
```diff
@@ -259,12 +259,13 @@ String
 ------
 
 There are four ways to express strings: basic, multi-line basic, literal, and
-multi-line literal. All strings must contain only valid UTF-8 characters.
+multi-line literal. Strings can contain any valid Unicode codepoint except the
+following control characters: U+0000 to U+0008, U+000A to U+001F, and
+U+007F. Note that tab (U+0009) is allowed. Multi-line strings can also contain
+newlines (U+000A) and carriage returns (U+000D).
```
abravalheri, line 265: I think that saying first that `U+000A` and `U+000D` are not allowed first (they are included in the ranges) and then adding an exception for Multi-line strings is kind of a double negative...

My opinion is that it would be better to not include these characters in the range and then specifically say that "basic" and "literal" (single-line) also don't allow them.

For example:

```markdown
multi-line literal. Strings can contain any valid Unicode codepoint except the
following control characters: U+0000 to U+0008, U+000B, U+000C, U+000E, U+001F, and
U+007F. Note that tab (U+0009) is allowed. Newlines (U+000A) and carriage returns (U+000D)
characters are forbidden in basic and literal string (non multi-line).
```

abravalheri:
> Does this rewording fix the issue about tabs implicit in PR #878 ?

Yes! Thank you @bendyarm, I will close that PR in favour of this one.

toml.md
```diff
@@ -405,9 +404,10 @@ apos15 = "Here are fifteen apostrophes: '''''''''''''''"
 str = ''''That,' she said, 'is still pointless.''''
 ```
 
-Control characters other than tab are not permitted in a literal string. Thus,
+Because most control characters are not permitted even in literal and multi-line
+literal strings, these literal strings are not suited for representing blobs of
+binary data.  It is recommended that you use Base64 or another suitable ASCII or
+UTF-8 encoding. The handling of that encoding will be application-specific.
```
eksortso, line 410: We have an alternative paragraph that expresses these same sentiments in #929.

abravalheri: @eksortso, I believe the version of the paragraph in this PR is more explicit and direct with regards to literal and multi-line literal strings to the point than the one in #929.

arp242:
Closing per discussion in #1064, where people preferred the current language.

PR #876, opened 2022-02-11
Force CRLF line endings in toml.abnf

bendyarm:
…  add .gitattributes so CRLF is added at checkout.  
Fixes issue #873 .  

Although the diff looks like the only change is the one added comment paragraph about UTF-8,  
the .gitattributes will convert the LFs to CRLFs.

### DIFF
--- a/toml.abnf
+++ b/toml.abnf
@@ -1,6 +1,10 @@
 ;; This document describes TOML's syntax, using the ABNF format (defined in
 ;; RFC 5234 -- https://www.ietf.org/rfc/rfc5234.txt).
 ;;
+;; Although a TOML document must be valid UTF-8, this grammar is written to
+;; apply to the string of non-negative integers that are the Unicode Scalar
+;; Values obtained from decoding the UTF-8 input.
+;;
 ;; All valid TOML documents will match this description, however certain
 ;; invalid documents would need to be rejected as per the semantics described
 ;; in the supporting text description.
--- a/toml.abnf
+++ b/toml.abnf
@@ -1,9 +1,8 @@
 ;; This document describes TOML's syntax, using the ABNF format (defined in
 ;; RFC 5234 -- https://www.ietf.org/rfc/rfc5234.txt).
 ;;
-;; Although a TOML document must be valid UTF-8, this grammar is written to
-;; apply to the string of non-negative integers that are the Unicode Scalar
-;; Values obtained from decoding the UTF-8 input.
+;; Although a TOML document must be valid UTF-8, this grammar refers to the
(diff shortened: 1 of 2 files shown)


toml.abnf
```diff
@@ -1,6 +1,10 @@
 ;; This document describes TOML's syntax, using the ABNF format (defined in
 ;; RFC 5234 -- https://www.ietf.org/rfc/rfc5234.txt).
 ;;
+;; Although a TOML document must be valid UTF-8, this grammar is written to
+;; apply to the string of non-negative integers that are the Unicode Scalar
+;; Values obtained from decoding the UTF-8 input.
```
arp242, line 6: Maybe I'm being stupid, but I don't really follow what "non-negative integers that are the Unicode Scalar Values obtained from decoding the UTF-8 input" means.

I *think* you mean to say that this document uses Unicode codepoints, rather than UTF-8 byte-encoding?

marzer: > Maybe I'm being stupid, but I don't really follow what "non-negative integers that are the Unicode Scalar Values obtained from decoding the UTF-8 input" means.

You're not stupid; I came here to make the same observation. No idea what that sentence is trying to convey. Seems like it's supposed to be a disambiguation of sorts? Has the opposite effect O_o

bendyarm:
If the checked-in file has CRs in it, then when you checkout, the `toml.abnf text eol=crlf` in `.gitattributes` will cause git to think the file changed when it did not.  This is inconvenient.  But if you put `toml.abnf -text` in `.gitattributes`, there are no line diffs in `git diff`.  So I think the this PR arrangement, where the CRs are added upon checkout and removed upon commit, is better than having literal CRs in the checked-in file.

toml.abnf
bendyarm: Thanks for the feedback.  It is important to be understandable.  I simplified the language (commit [1f8a780](https://github.com/toml-lang/toml/pull/876/commits/1f8a780805e136e1e10e8abbf272dd245e2843f4)).  
The reason I think a paragraph like this will be helpful to some people is because the spec `toml.md` states "A TOML file must be a valid UTF-8 encoded Unicode document", but the grammar does not refer to UTF-8 "code units" (byte values).  There is a decoding step prior to parsing.  Hopefully the new language is acceptable.

The most important part of this PR, though, is to fix issue #873 by adding CRs on checkout.

marzer: @bendyarm Ah, I see. Thanks for the clarification.

pradyunsg:
Thanks @bendyarm! ^>^

Issue #877, opened 2022-02-12
Documentation is ambiguous about tab `\t` in single-line basic strings


abravalheri:
Hello, I was just reading the [spec](https://toml.io/en/v1.0.0#string) in the website and I noticed that the docs are ambiguous about tab being required or not to be escaped in single-line basic strings. Specifically, it states:

```
Basic strings are surrounded by quotation marks ("). Any Unicode character may be used except those that
must be escaped: quotation mark, backslash, and the control characters other than tab (U+0000 to U+0008,
U+000A to U+001F, U+007F).
```

So according to the text, tab characters don't need to be escaped.

However, the very next example, shows tabs being escaped:

```
str = "I'm a string. \"You can quote me\". Name\tJos\u00E9\nLocation\tSF."
```

Am I correct to assume that the text + example implies that `\t` chars *don't have to* be escaped, but *can*?

I had a quick look in the ABNF and I have the impression that tabs need to be escaped, but I might be wrong interpreting the ABNF...

pradyunsg:
What @frostming said! :)

Tab is allowed both as a tab character, as well as in an escaped format.

PR #878, opened 2022-02-12
Clarify tab can be escaped (or not) in basic strings

abravalheri:
As discussed in #877, the current description of basic strings in the docs say that tab characters are allowed.

However this text does it using a double negative (`except... other than...`), and is immediately followed by an example that uses escaped tabs `\t`. As a consequence it is a bit hard to understand.

The change introduced here explicitly says that tabs can be used either directly or in its escaped form.


pradyunsg (commented):
Thanks for filing this PR! I'm not sure that there's any value in duplicating information that's literally mentioned in a code block above, in prose.

toml.md
```diff
@@ -284,6 +284,8 @@ For convenience, some popular characters have a compact escape sequence.
 \UXXXXXXXX - unicode         (U+XXXXXXXX)
 ```
 
+Tab characters can be used either directly or in their escaped form `\t`.
```
pradyunsg, line 287: This is covered on line 277. I'm not sure what this clarification achieves.

abravalheri: Hi @pradyunsg, thank you very much for the review.

I don't think line 277 covers the same content though.

Line 277 says that `\t` is a escape code for the tab character, but it does not explicitly say tab characters can be used directly also.

That in theory is covered by line 266. However it is hard to understand because of the double negative.

The idea of this addition is to explicitly state it can be used *both* ways, which is not explicitly stated elsewhere.

As a non native speaker, I had a hard time processing this info, so I was hoping to make it clearer.

abravalheri: In other ways the information I am hoping to convey with that phrase is that people should not be afraid of using an "unescaped" tab char (which is not covered in line 277).

abravalheri:
Closed in favour of #875 😄

PR #880, opened 2022-02-17
Clarify trailing commas

mol-lux:
I was trying to find out whether a comma is required behind the last item in array.

The Documentation currently states the following:

> A terminating comma (also called a trailing comma) is permitted after the last value of the array.

This, however, may arouse the suspicion that the comma is, in fact, required or encouraged. This PR changes the wording of that paragraph to make it abundantly clear that both options, with or without comma, are possible and equally valid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## unreleased
 
 * Clarify where and how dotted keys define tables.
+* Clarify that a trailing comma is not required on the last item in an array.
 
 ## 1.0.0 / 2021-01-11
 
--- a/toml.md
+++ b/toml.md
@@ -628,9 +628,10 @@ contributors = [
 ```
 
 Arrays can span multiple lines. A terminating comma (also called a trailing
-comma) is permitted after the last value of the array. Any number of newlines
-and comments may precede values, commas, and the closing bracket. Indentation
-between array values and commas is treated as whitespace and ignored.
+comma) is permitted, but not required, after the last value of the array.
+Any number of newlines and comments may precede values, commas, and the closing
+bracket. Indentation between array values and commas is treated as whitespace and
+ignored.
 
 ```toml
 integers2 = [


ChristianSi:
What about "permitted" is it you don't understand?

pradyunsg:
Hiya! 

> This, however, may arouse the suspicion that the comma is, in fact, required or encouraged.

I don't think such a suspicion comes up -- permitted means "to allow the doing of (something)", which is what is the case here. You're allowed to have a trailing comma. I don't think this clarification is particularly useful.

> What about "permitted" is it you don't understand?

Hey @ChristianSi, while I appreciate your response here and likely have the same position as you do, I want to call out that your tone in this response is harsh and slightly rude -- and not particularly welcoming.

pradyunsg:
Closing this, since I don't think this is clarification is necessary. Thanks @mol-lux for filing this PR none the less! :)

redbar0n:
@pradyunsg I initially read it as you and @ChristianSi did, and was sympathetic to the "'permitted' should be sufficient argument" for closing this PR. But on second thought, I changed my mind.

I think @mol-lux might have read it as:

> A terminating comma (also called a trailing comma) is what's permitted after the last value of the array.

Injecting `what's` or similar. That's an honest mistake to make when scanning the text. People often fill in implied words, based on their cultural and linguistical context (e.g. some places suggestions are _not_ simpy suggestions, and other places you cannot do almost anything without permission). In communication, it's also quite common for receivers to assume the sender had a stronger intent than what was communicated. I'm sure you, as well as I, can relate to real life situations where our accuracy and careful wording has been entirely overlooked by the listener.

While I agree that the simple word "permitted" _should be_ sufficiently clear. It obviously wasn't, hence the existence of this issue. If only 1 % of people briefly scan the text as @mol-lux did, and get the wrong idea, then this PR is justified, imho. After all, it only includes 3 simple words `, but not required,` for a bit of redundancy in the clarification. Who hasn't appreciated a bit of helpful redundancy once in a while when learning?

ChristianSi:
@redbar0n: What difference would it make? Without or without "what's", the trailing comma is merely permitted – nothing more, nor less.

redbar0n:
@ChristianSi `what's` indicates exclusivity.

marzer:
>  is what's permitted

Eh. This reads very informal to me. Almost sounds like cockney slang or something.

> is permitted

This is exclusive enough. I doubt anyone is going to interpret that and think "oh but it didn't explicitly say the only thing was a trailing comma! Ergo everything must be fine there!"

redbar0n:
>Also consistency is important; if we clarify the use of "permitted" here, we'd have to be careful to apply the same clarification anywhere we use the word in the spec, both now and in the future. That level of invariant is quite an undertaking for a living document.

We're talking about adding `, but not required,`, exactly 19 characters, to help someone who indicated it was an issue for them. I have argued the case for it as well as I can, so I can only refer back to my post.

Consistency is important, but clarity is also. I would be pragmatic about it: if you receive word that the misunderstanding has cropped up again elsewhere, you could address it then. As far as I can see `is permitted` appears exactly once in `toml.md`.

> At what point is something "sufficiently clear", by your estimation? You'll never reach 100% reading it correctly on the first try.

At the point people do not longer feel enough pain to actually take the time to submit an issue/PR to (help to) fix it.
But it's your repo, so you're free to do as you please.

marzer:
> But it's your repo, so you're free to do as you please.

Eh. It's not my repo. I'm just a guy that has a TOML implementation. I've personally had a whole bunch of spec mis-reads, but it has (in all but a few cases) turned out I was just being careless and read the spec badly.

I don't doubt the motivation. I sympathize. I just don't empathise in this specific case

ChristianSi:
Considering that this is the only case where "is/are permitted" is used in a positive sense (otherwise we only state which things are NOT permitted, i.e. prohibited), one could change

> A terminating comma (also called a trailing comma) is permitted after the last value of the array.

to something like

> You may add a trailing comma after the last value of the array (this comma will be ignored).

(I'm not sure why the term "terminating comma" is used here – it's not terminating anything.)

This would be more in line with RFC 2119 (which standardizes terms such as MAY and MUST) and with the language used in the rest of the spec, which often has "may", but never "is/are permitted".

ChristianSi:
I could prepare a little PR (or someone else could), but let's wait first the see what the maintainers think.

PR #887, opened 2022-03-06
Fix `\e` CHANGELOG note being in the wrong place

marzer:
The `\e` PR, #790, was made before v1.0.0 was released, so the change it makes to CHANGELOG put the change note in the wrong place.


pradyunsg:
Nice catch! :)

PR #889, opened 2022-03-11
Replace "UTF-8 characters" with "Unicode characters"

paulehoffman:
UTF-8 is an encoding, not a character set, so "UTF-8 character" is not correct.

I picked the "should be implemented as" wording from floats, but am not sure this is what you want. It feels like there should be some consistent way to say "TOML interpreters should output $type as $format" for strings, integers, and floats (and maybe others...).

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -260,7 +260,6 @@ String
 
 There are four ways to express strings: basic, multi-line basic, literal, and
 multi-line literal. All strings must contain only valid Unicode characters.
-Strings should be implemented as UTF-8 byte sequences.
 
 **Basic strings** are surrounded by quotation marks (`"`). Any Unicode character
 may be used except those that must be escaped: quotation mark, backslash, and
--- a/toml.md
+++ b/toml.md
@@ -259,7 +259,8 @@ String
 ------
 
 There are four ways to express strings: basic, multi-line basic, literal, and
-multi-line literal. All strings must contain only valid UTF-8 characters.
+multi-line literal. All strings must contain only valid Unicode characters.
+Strings should be implemented as UTF-8 byte sequences.
 
 **Basic strings** are surrounded by quotation marks (`"`). Any Unicode character
 may be used except those that must be escaped: quotation mark, backslash, and


toml.md
```diff
@@ -259,7 +259,8 @@ String
 ------
 
 There are four ways to express strings: basic, multi-line basic, literal, and
-multi-line literal. All strings must contain only valid UTF-8 characters.
+multi-line literal. All strings must contain only valid Unicode characters.
+Strings should be implemented as UTF-8 byte sequences.
```
pradyunsg, line 263:
```suggestion
multi-line literal. All strings must contain only valid UTF-8 encoded
Unicode characters.
```

paulehoffman: The change from "Unicode characters" to "UTF-8 encoded Unicode characters" is unnecessary because the top of the spec says "A TOML file must be a valid UTF-8 encoded Unicode document."

Can you explain why you removed "Strings should be implemented as UTF-8 byte sequences."? As noted in my original comment, other parts of the spec say how output should be implemented.

ChristianSi:
The think the first part of the patch is good.

> All strings must contain only valid UTF-8 characters.

should certainly become

> All strings must contain only valid Unicode characters.

since that's clearly the intended meaning.

As for "Strings should be implemented as UTF-8 byte sequences." No, that
sentence is confusing and unnecessary. It suggests that strings are somehow
encoded separately from the rest of the document, which of course is not
the case. And the spec already says:

> A TOML file must be a valid UTF-8 encoded Unicode document.

That's clear and sufficient. We don't describe comments as "UTF-8 byte sequences" either. No, they are just (Unicode) text, like the whole TOML document.

paulehoffman:
I'm OK with removing the "should be implemented as" part and will instead open this as a general issue.

toml.md
pradyunsg: Fair. I've editted to only remove the second suggestion then! :)

See https://github.com/toml-lang/toml/issues/890 for why I'm opposed to using "byte sequences" as the language here.

pradyunsg (changes requested):
I don't want to use the "byte sequences" language. Let's remove that from this PR.

arp242:
What is a "valid Unicode character"? Is something in the private use plane a "valid Unicode character"? Is something that's unassigned in Unicode 14 (but assigned in maybe the next unreleased unicode version) "valid"? Is something like "U+0000 U+200B U+0000" (NUL zero-width joiner NUL) "valid"?

Depending on context and how your application uses the TOML, you can answer either "yes" or "no" to all of those questions (and some others too).

I'd change it to "All strings must be encoded as UTF-8." and just not talk about "valid characters" at all. That this needs to be valid UTF-8 is implied since invalid UTF-8 is not UTF-8.

arp242:
Alrighty, fair enough; in my reading the "valid" would be obvious/implied; same if I say "all dates must be in RFC 3339 format" the "valid" is implied.

paulehoffman:
Assuming "valid" is incredibly unsafe in a spec like this, and thus should be stated explicitly. Decoders should know that they can drop invalid inputs, not be forced to try to interpret them.

ChristianSi:
I would suggest to simply delete the sentence "All strings must contain only valid UTF-8 characters." from the spec. The "Spec" section already says

> A TOML file must be a valid UTF-8 encoded Unicode document.

That's sufficient.

In fact I'm surprised that the "UTF-8 characters" language is still in the spec. It's true that it makes no sense, so this must be an oversight.

ChristianSi:
For strings, one could also write:

> Strings are sequences of Unicode characters.

It's true there is no need to use "valid" here, since an "invalid Unicode character" is presumably not a "Unicode character" at all.

pradyunsg (approved):

PR #891, opened 2022-03-15
Allow non-English scripts for unquoted keys

### LINKED ISSUE #687: Relax bare key restrictions to allow additional unicode letters and numbers
# Issue
TOML's "bare key" syntax is too restrictive. People who regularly use characters from languages other than English should be able to do so in TOML keys without additional gymnastics. 

I know there's already been [a](https://github.com/toml-lang/toml/issues/337) [lot](https://github.com/toml-lang/toml/issues/220) [of](https://github.com/toml-lang/toml/issues/65) [discussion](https://github.com/toml-lang/toml/pull/67) about this but much of it was from when TOML was less established and I think it warrants revisiting.

# Proposed change
Expand the set of accepted characters allowed in bare keys to include letters and numbers from the entire Unicode space, similar to how identifiers are handled in other Unicode-compliant contexts (e.g. python, javascript, etc.).  Specifically:
- Allow codepoints from categories [Ll](https://www.compart.com/en/unicode/category/Ll), [Lm](https://www.compart.com/en/unicode/category/Lm), [Lo](https://www.compart.com/en/unicode/category/Lo), [Lt](https://www.compart.com/en/unicode/category/Lt), [Lu](https://www.compart.com/en/unicode/category/Lu), [Nd](https://www.compart.com/en/unicode/category/Nd) and [Nl](https://www.compart.com/en/unicode/category/Nl) anywhere in a bare key
- Allow codepoints from categories [Mc](https://www.compart.com/en/unicode/category/Mc) and [Mn](https://www.compart.com/en/unicode/category/Mn) anywhere in a bare key _except as the first character_

# Rationale
After reading much of the existing discussion on the issue, I've identified the points below as being the main objections. I've written a counterpoint for each.

## "ASCII-only is easy to understand"
Allowing Unicode letters and numbers wouldn't change the understandability of the written word in "mostly-ASCII" contexts, excepting maybe people from English-centric countries encountering characters they otherwise rarely see and being unsure how to pronounce them. I'm one of those people and my brain seems to consume them just fine. And it's _almost certainly_ going to improve the understandability of bare keys to people for whom an ASCII environment is _not_ their regular one.

It also wouldn't change the semantic/syntatic understandability of the language; I'm only advocating relaxing the spec to allow letter and number characters, not anything that might be confused for a language construct (no math symbols, for instance).

## "Guides users to choose simple key names"
See above. I'd argue that the keys would be no less simple with this change. I live and work in a European country and a number of my friends and colleagues have non-ASCII letters in their name (e.g. `ä`). I doubt they consider their names to be complex; I certainly don't. If anything, by forcing people to jump through hoops just to type in their language, we're actually making the key names _more_ complex w.r.t. cognitive load.

## "Eliminate any weirdness that could come from having to deal with undelimited Unicode"
The TOML spec dictates UTF-8, not UTF-8-ish. UTF-8 is a solved problem at this point. If a parser doesn't correctly detect and handle malformed UTF-8, I'd argue that the parser needs fixing, not that we should bend over to accommodate users who are using crap tools and libraries. It's such a solved problem that you can even portably consume it [using a state machine](http://bjoern.hoehrmann.de/utf-8/decoder/dfa/) and [validate it using vector intrinsics](https://lemire.me/blog/2018/10/19/validating-utf-8-bytes-using-only-0-45-cycles-per-byte-avx-edition/).

## "Keys should be identifier-like"
Despite the fact that the concept of an "identifier" isn't a thing in TOML, I'll concede that in some situations this might be a concern. A reasonable example is using TOML in code generation contexts; if you used TOML keys to inform variable names historically you'd run into issues in many languages with non-ASCII characters, though this is no longer true. Even good old C++ supports unicode characters in identifiers [on modern compilers](https://godbolt.org/z/N8Bgki).

..._all of which is rendered moot_ by the fact that TOML supports hyphens in bare keys which are often invalid in identifier contexts, so this objection is a non-starter anyway.

## "It complicates implementation"
It really doesn't. Many implementations will be able to leverage built-in helper functions or libraries for working with Unicode. For those that can't, I've put my money where my mouth is and implemented this as a proof-of-concept in my own TOML parser and I'm happy for my code to be used as a starting point:
- [Unicode utils](https://github.com/marzer/tomlplusplus/blob/master/include/toml%2B%2B/toml_utf8.h), with `is_unicode_XXXXX()` codepoint identity functions (generated by a [script](https://github.com/marzer/muu/blob/master/python/generate_unicode_functions.py))
- [Parser](https://github.com/marzer/tomlplusplus/blob/master/include/toml%2B%2B/toml_parser.hpp)

Of course you might argue that simply accepting UTF-8 bytes from a TOML implementation is not an option for everyone, and you'd be right; there will always be situations where only ASCII makes sense (e.g. legacy codebases). I'd respond by pointing out that detecting non-ASCII characters in a character stream [is laughably trivial](https://stackoverflow.com/questions/48212992/how-to-find-out-if-there-is-any-non-ascii-character-in-a-string-with-a-file-path). Applications requiring ASCII-only can easily enforce this themselves.

abelbraaksma:
Fixes: #687

Support all international scripts, except for certain non-symbols, arrows and joins. This PR largely follows the recommendation in my comment (https://github.com/toml-lang/toml/issues/687#issuecomment-567490186), except where that conflicted with backward compat issues (i.e., we allow keys to start with a digit or dash, and that should stay that way).

The excluded ranges are mostly those that are really not "letterlike" (arrows, line drawing chars, spaces, punctuation). If someone feels that some of the excluded ranges should be included, by all means, let me know.

ABNF verified with https://tools.ietf.org/tools/bap/.

Here are some examples that will work once this is allowed. Tested with http://instaparse.mojombo.com/.

```toml
# Examples that are now valid unquoted keys
zeeën = "'seas' - Dutch"
Cuántos-años = "'How old' - Spanish"
الساعة = "arabic"
汉语大字典 = "cjk ideographs"
辭源 = "Ciyuan"
பெண்டிரேம் = "'we are women' - in Tamil"
गंगा = "'Ganges' - in Devanagari, Hindi"
העברית = "'Academy' - Hebrew"
Тіні-забутих-предків = "'Shadows of forgotten ancestors' - movie title in Ukrainian"
VäinöLinna = "finnish author"
బడికి = "'School' - in Telugu"
Người = "'person' - Vietnamese"
😂 = "x1F602 - smiley"
ᚠ = "'Cattle' - Rune"
𓆣𓆤𓆥𓆦 = "Egyptian hieroglyphs"
Fuß = "'Foot' - German"
français  = "'French' - French"

# These are still valid
- = "single dash"
3 = "digit"
_ = "underscore"
```


toml.md
```diff
@@ -104,27 +104,33 @@ Keys
 
 A key may be either bare, quoted, or dotted.
 
-**Bare keys** may only contain ASCII letters, ASCII digits, underscores, and
-dashes (`A-Za-z0-9_-`). Note that bare keys are allowed to be composed of only
-ASCII digits, e.g. `1234`, but are always interpreted as strings.
+**Bare keys** may contain any letter-like Unicode character from any Unicode script,
+may contain ASCII digits, dashes and underscores. Punctuation, spaces, arrows, box drawing 
+and private use characters are not allowed.  Note that bare keys are allowed to be
```
hukkin, line 109: Could we explicitly list the allowed Unicode codepoint ranges similarly as we already do for comments and strings? I think that is valuable to implementers.

arp242: Personally I'd rather see this defined in Unicode categories rather than codepoints (both in the ABNF and in here – you'll still have to copy the codepoints to the ABNF as I don't think you can list Unicode categories, but that seems like a better reference point than a series of characters).

abelbraaksma: @hukkin, I wouldn't mind doing that, but it makes the paragraph a little wordier and arguably a bit harder to read. I strove to keep it simple ;).

hukkin: I understand, but my argument is that for an implementer it'd be a lot _easier_ to read. Compare that to Google searching the meaning the vague concepts like "letter-like Unicode character", "Unicode arrows" and so on. I think a technical spec shouldn't leave room for interpretation.

marzer: I tend to agree with @hukkin here, though would also be satisfied with an explicit note directing implementers to the ABNF, if the full list is too verbose.

abelbraaksma: @hukkin & @marzer, let me think of some useful wording and see how it pans out.

toml.abnf
```diff
@@ -48,17 +49,29 @@ comment = comment-start-symbol *non-eol
...
+val = string / boolean / array / inline-table / date-time / float / integer
 simple-key = quoted-key / unquoted-key
 
-unquoted-key = 1*( ALPHA / DIGIT / %x2D / %x5F ) ; A-Z / a-z / 0-9 / - / _
+;; Unquoted key
+
+unquoted-key = 1*unquoted-key-char
+unquoted-key-char = ALPHA / DIGIT / %x2D / %x5F         ; a-z A-Z 0-9 - _ 
+unquoted-key-char =/ %xC0-D6 / %xD8-F6 / %xF8-FF        ; non-symbol chars in Latin block
+unquoted-key-char =/ %x100-2FF / %x300-37D / %x37F-1FFF ; this excludes GREEK QUESTION MARK, which is basically a semi-colon
```
ChristianSi, line 60: There seems to be a trailing space after "semi-colon", which could be trimmed.

ChristianSi (changes requested):
Good work!

Regarding the text in the written spec, I think the wording with "any letter-like Unicode character from any Unicode script" is good, but I also think the spec should explain what exactly that means. As I see it, the written spec should be self-contained and one should not have to read the ABNF to find out what it actually means. So, we could leave the paragraph as suggested and then add a sentence such as: "More specifically, the allowed characters are ASCII letters and digits (`A-Za-z0-9-`), the underscore `_`  and dash `-`, a well as the Unicode characters U+00C0 to U+00D6, U+00D8 to U+00F6, U+00F8 to U+00FF, U+0010 to U+002FF, U+0300 U+037D" (etc.), this translating the "unquoted-key-char" rule from the ABNF into readable text. It's just 8 lines in the ABNF, so it should be OK as a sentence (maybe set as its own paragraph).

toml.md
abelbraaksma: @hukkin, @marzer, done. I chose to use wording like "all of this, except ...", as that tended to be more readable in my opinion. The ABNF doesn't have except-syntax (EBNF does, btw), so the ABNF looks a bit more convoluted than it needs to be, given that certain individual chars need to be excluded.

The prose is as follows in the `toml.md` now:

* From the ASCII characters, only A-Z, a-z, 0-9, _ and - are allowed
* From the rest of the first 256 characters, only 0080-00BF, "×" (00D7) and "÷" (00F7) are disallowed
* All of 0100-1FFF are allowed, except ";" (037E)
* Characters 200C, 200D, 203F, 2040, 2070-218F, 2C00-2FEF are allowed
* All characters from 3001, except surrogates (D800-D999), Private Use (E000-F8FF, F0000-100000) and process-internal use (FDD0-FDEF)

toml.md
```diff
@@ -109,6 +109,12 @@ may contain ASCII digits, dashes and underscores. Punctuation, spaces, arrows, b
 and private use characters are not allowed.  Note that bare keys are allowed to be
 composed of only ASCII digits, e.g. `1234`, but are always interpreted as strings.
 
+* From the ASCII characters, only A-Z, a-z, 0-9, _ and - are allowed
+* From the rest of the first 256 characters, only 0080-00BF, "×" (00D7) and "÷" (00F7) are disallowed
```
ChristianSi, line 113: I would use the U+ syntax here, as elsewhere in the spec ("U+0080 to U+00BF" etc.)

toml.md
```diff
@@ -109,6 +109,12 @@ may contain ASCII digits, dashes and underscores. Punctuation, spaces, arrows, b
...
+* All of 0100-1FFF are allowed, except ";" (037E)
+* Characters 200C, 200D, 203F, 2040, 2070-218F, 2C00-2FEF are allowed
+* All characters from 3001, except surrogates (D800-D999), Private Use (E000-F8FF, F0000-100000) and process-internal use (FDD0-FDEF)
```
ChristianSi, line 116: That means something like "U+3001 and higher", right? It's not quite clear.

ChristianSi (changes requested):

toml.md
abelbraaksma: was thinking of that too, will do ;)

toml.md
abelbraaksma: I'll update

toml.md
```diff
@@ -104,27 +104,39 @@ Keys
 
 A key may be either bare, quoted, or dotted.
 
-**Bare keys** may only contain ASCII letters, ASCII digits, underscores, and
-dashes (`A-Za-z0-9_-`). Note that bare keys are allowed to be composed of only
-ASCII digits, e.g. `1234`, but are always interpreted as strings.
+**Bare keys** may contain any letter-like Unicode character from any Unicode script,
```
marzer, line 107: Should we take this opportunity to audit the spec for "character" vs "code point"? My personal feeling is that we should avoid the word "character" when talking about Unicode, and prefer the more technically-correct "code point", since "character" tends to mean lots of different things to different people (it even means different things in Unicode itself!).

toml.md
abelbraaksma: done

toml.md
hukkin: Thanks! I'm happy with this, but I think I'd slightly prefer the type of language ChristianSi used in this comment https://github.com/toml-lang/toml/pull/891#pullrequestreview-911367703

- It is more in line with the language used defining allowed comment and string characters
- You already specify the allowed characters in "human language" at the start of the paragraph. This listing should be targeted at developers/implementers and for them the raw codepoint ranges with no additional context are IMO the most convenient.

toml.md
abelbraaksma: Done

toml.md
```diff
@@ -104,27 +104,33 @@ Keys
...
+may contain ASCII digits, dashes and underscores. Punctuation, spaces, arrows, box drawing
```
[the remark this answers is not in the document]
abelbraaksma: Done

toml.abnf
abelbraaksma: Done, removed

abelbraaksma:
>  It's just 8 lines in the ABNF, so it should be OK as a sentence (maybe set as its own paragraph).

@ChristianSi, I updated with a paragraph as bulleted list, with I think is now quite readable (other comments/suggestions have been addressed).

abelbraaksma:
The error is in the build is a bit odd, not sure what to do about that:
image

toml.md
abelbraaksma: @hukkin, I'm ok either way. I've meanwhile updated the text with the comments from other reviewers. If there's consensus I can change it if need be.

toml.md
abelbraaksma: @marzer, the word "character" is used 21 times in the document. I agree that "code-point" is better, at least in most cases. In Unicode, a character can be, say `é`, but it can be either one or two code-points. For characters in some other langs, like Arabic, there can be many more codepoints that comprise a single character.

hukkin:
> The error is in the build, which says "All checks have failed" is a bit odd, not sure what to do about that:

The problem is that pre-commit.ci has been activated for this repository, but there is no configuration file for it. Not related to your changes at all.

toml.md
marzer: >The only argument against using "code-point" is that a large swatch of people won't know what that means...

Yeah, though implementers absolutely must know what it means, so for contexts where the prose is largely implementer-focused (e.g. listing acceptable code point ranges) I'd say that's not too much of an issue. 

I don't consider myself anything close to a unicode expert, though, so I don't really have any suggestions about wording changes. Happy to defer to your judgement here.

arp242: My 2¢ is that this seems a bit of a "technically correct" vs. "perhaps not entirely technically correct, but everyone knows what you mean and it reads a bit better"-discussions. AFAIK no one has ever been confused by "character", so I think it's fine to just keep it.

toml.abnf
```diff
@@ -46,19 +46,31 @@ comment = comment-start-symbol *non-eol
...
+
+unquoted-key = 1*unquoted-key-char
+unquoted-key-char = ALPHA / DIGIT / %x2D / %x5F         ; a-z A-Z 0-9 - _
+unquoted-key-char =/ %xC0-D6 / %xD8-F6 / %xF8-37D       ; non-symbol chars in Latin block
+unquoted-key-char =/ %x37F-1FFF                         ; exclude GREEK QUESTION MARK, which is basically a semi-colon
+unquoted-key-char =/ %x200C-200D / %x203F-2040          ; include combining chars used in some languages
+unquoted-key-char =/ %x2070-218F / %x2C00-2FEF          ; this excludes arrows, blocks and the like
+unquoted-key-char =/ %x3001-D7FF                        ; skip 2FF0-3000 ideographic up/down markers and spaces
+unquoted-key-char =/ %xF900-FDCF / %xFDF0-FFFD          ; skip D800-D999 surrogate block, E000-F8FF Private Use area, FDD0-FDEF intended for process-internal use (unicode)
+unquoted-key-char =/ %x10000-EFFFF                      ; all chars outside BMP range, excluding Private Use planes
```
arp242, line 64: These ranges currently seem include a number of codepoints that are not yet defined; for example `37F-1FFF` includes U+0380-U+0383, which are not defined, as well as others.

These are in the "Greek and Coptic" block, and it's not like they can get defined as just $anything in the future, but this could lead to some undesirable characters/codepoints being included with future Unicode updates, especially with that last `10000-EFFFF` line (which includes unassigned blocks).

There's quite a few of these; I have [a little CLI tool](https://github.com/arp242/uni/) to print it:

    % uni p C0-D6 D8-F F8-37D  37F-1FFF 200C-200D 203F-2040 2070-218F 2C00-2FEF 3001-D7FF  F900-FDCF FDF0-FFFD | grep 'NOT IN UNICODE' | wc -l
    1392

    % uni p 10000-EFFFF | grep 'NOT IN UNICODE' | wc -l
    833334

This ties in a bit with my other comment about using Unicode categories rather than a list of codepoints.

abelbraaksma:
@arp242, (moving to the main thread so that our discussion remains visible in the future) as I read your comments, you basically have two concerns. One is about whether using Unicode Categories would be better suited and the second about why I included unassigned code-points.

### Including unassigned code-points
This is very much "by design". We're trying to be as inclusive as possible. Since nobody can predict what code-points get assigned in the future, nor in what categories they get, the best we can do is include unassigned code-points, unless the block is excluded as "not a character" in Unicode (a few such ranges are in there).

This argument is precisely what large standards bodies like W3C use when they need to design specs. You want your spec to be future-proof and you cannot do that when you exclude ranges that may be assigned in the future.

### Using categories vs fixed ranges
Categories are in flux and code-points get reassigned. This is not only a problem between Unicode versions, but also in libraries. For instance, certain characters in [.NET Framework are in a different category than they are for .NET 5](https://docs.microsoft.com/en-us/dotnet/core/compatibility/globalization/5.0/unicode-categories-for-latin1-chars). Other libraries rely on whatever is the system default, which means running code on one system may fail on another.

From a parsing perspective, one could hard-code the code-points of the categories instead of naming them, which would probably be the way to go. But that is an arduous task and leads to unwieldy ABNF that is hard to analyze. Plus many code-points that are inside `Lo` category shouldn't be included and conversely, many code-points that aren't in one of the `L` categories should be added. 

Finally, if you don't hard-code the categories, you'll need to fix the Unicode version. But suppose you'd pin it to version 5 (most widely implemented), then you miss out on all the good stuff since version 5. Suppose you'd pin it to version 10 (much less widely implemented). Now you're asking implementers that want to conform to TOML to implement all the ranges for categories of version 10. Plus that users may find on their system that character "X" is in `Ll`, but TOML doesn't allow it because it used a version in the past. This breaks the "be liberal in what you accept" paradigm.

### Conclusion
In the end, there are pros and cons for either approach and I hear what you are saying. I've been in exactly that spot when we discussed certain specs in the W3C technical committees. In the end we decided to be inclusive rather than exclusive and it proved to be the easier path. I can kindly repeat my quote from the original thread here:

> Almost all characters are permitted in names, except those which either are or reasonably could be used as delimiters. The intention is to be inclusive rather than exclusive, so that writing systems not yet encoded in Unicode can be used in XML names.

and:

> The ASCII symbols and punctuation marks, along with a fairly large group of Unicode symbol characters, are excluded from names because they are more useful as delimiters in contexts where XML names are used outside XML documents; providing this group gives those contexts hard guarantees about what cannot be part of an XML name.

abelbraaksma:
>> Plus, many Unicode libraries assign different codepoints to categories.

> Character classes are defined in Unicode; perhaps some libraries get them wrong, but I never encountered problems with this. AFAIK this is not a wide-spread problem(?)

Actually, it is. Even the latest version (v14.0) added a bunch of new code-points to the the `L` category. So categories are in constant flux and each version expands on that. See https://www.unicode.org/reports/tr44/tr44-28.html#Unicode_14.0.0, specifically:

> _"Newly encoded modifier letters in the range U+10780..U+107BA were assigned the value Other_Lowercase, for consistency with other, similar modifier letters."_

Bottom line is, if we want to be stable, we'd best use a predefined, as wide as possible set that explicitly includes unassigned code-points.

arp242:
> You want your spec to be future-proof and you cannot do that when you exclude ranges that may be assigned in the future.

I don't really follow this to be honest. The goal of this PR is to allow only letter and digit-like symbols, but with such an inclusive range in the future it might include punctuation, new arrows, or other things now explicitly excluded. So you risk ending up with something very inconsistent. To me this seems much *less* 

Personally I'd rather erroneously exclude some symbols from the allowed list than accidentally include some we didn't intend, since this can be updated with a simple backwards-compatible spec change, whereas the reverse can't be fixed and we'll be stuck with it. That seems more future-proof to me. In the issue about this you mentioned "It follows the well-known adage: be liberal in what you accept", and I actually don't agree with that principle at all. There's been tons of discussions about this in the last few decades and no need to repeat it here, but this seems to be at the core of the issue.

We can also just say "include everything except `=`, `[`, `.`, `]`, some (or all?) spaces", or something to that effect. Do we really need to exclude things like arrows or block drawing characters? It's perhaps kind of silly to use it, but then again so is using Linear-A or various emojis, and that's allowed now too. That would make this entire discussion moot. I can't really think of any serious downsides right now.

toml.md
marzer: @arp242 I'm content with that reasoning too. If "character" is unambiguous in the general case, it's probably fine. Just thought I'd flag it as specs tend to be more precise with Unicode terminology.

toml.md
```diff
@@ -104,27 +104,39 @@ Keys
...
 
-**Bare keys** may only contain ASCII letters, ASCII digits, underscores, and
-dashes (`A-Za-z0-9_-`). Note that bare keys are allowed to be composed of only
-ASCII digits, e.g. `1234`, but are always interpreted as strings.
+**Bare keys** may contain any letter-like Unicode character from any Unicode script,
+as well as ASCII digits, dashes and underscores. Punctuation, spaces, arrows, box drawing 
+and private use characters are not allowed.  Note that bare keys are allowed to be
+composed of only ASCII digits, e.g. `1234`, but are always interpreted as strings.
+
+* From the ASCII characters, only A-Z, a-z, 0-9, _ and - are allowed
```
ChristianSi, line 112: * From the ASCII characters, only `A-Z`, `a-z`, `0-9`, `_` and `-` are allowed

ChristianSi (changes requested):
This looks very good to me now, except for tiny formatting change I would suggest.

ChristianSi:
@arp242 I don't quite understand the point you're trying to make. In one paragraph you suggest to be more restrictive, in the next one to be *less* so.

My own feeling, and your wavering maybe confirms it, is that this PR has already managed to strike a good balance. Also, as I understand it, this is very close to what XML/HTML do, and I'd say that that if it works for those giants, it can work for TOML too!

abelbraaksma:
> We can also just say "include everything except =, [, ., ], some (or all?) spaces", or something to that effect. Do we really need to exclude things like arrows or block drawing characters?

Technically, there's nothing that stops us from doing that, I agree. There's however the long-standing consensus that "idents" need to follow certain rules and generally those rules have been to exclude characters or code-points that aren't letterlike. Fragment identifiers, perhaps the most widely used targets for URIs, follow the exact same logic as I've laid out here and it has been in use for over 25 years without issues.

> It's perhaps kind of silly to use it, but then again so is using Linear-A or various emojis, and that's allowed now too.

Yes, emojis are silly, as is Linear-A. But even in the most recent Unicode version, characters have been added for oft-used languages like Telugu, Latin, CJK Ideographs, Mongolian. obviously, they won't be immediately used by everyone, but that's not the point here. Since version 5.0 (Windows 7) or 5.1 (Windows 10) there have been many thousands of characters added.

I wouldn't oppose specifically excluding emojis. The `xml:id` spec is from 2005, the XML and related HTML specs that define `Name` and `NCName` are from 1998 and about. It was with great foresight that they decided not to lock in the character ranges, yet a small byproduct of that foresight lead to weird things like emojis now being allowed in international domain names, identifiers, tagnames and html fragments. Nobody's ever really gotten worried about that.

If such massive standards (HTML, XML, URI) managed to deal with this for a quarter of a century, without change or backward/future compat issues, I think TOML can do to and should follow good example.

> I don't really follow this to be honest.

I'm sorry about that. Apparently I lack the right words to explain myself clearly. If you look closely to the ranges, you'll notice that in the ASCII range a lot of characters are excluded, which can be used for future use if TOML needs it (i.e., allow `foo + bar` of `foo : bar` to mean something).

With "future proof" I mean that it still allows TOML to change over time. It also means that there won't be new versions of TOML just to cater for changes in categories (which happen just about every year in Unicode). 

Beside that, implementers that have to deal with libraries that do not have strong Unicode support (this includes .NET w.r.t. up-to-date categories), need to implement all this by hand. Keeping things simple helps in adoption.

Anyway, I value your input and point of view, and this is just my take on this feature change. While I do feel strongly about doing it the "standard, well-trodden way", I'm not married to any particular method. If the TOML powers that be disagree with what I'd call my "simple, yet inclusive" approach, I've no problem to switch gears and go with another flow ;).

ChristianSi:
As for the "character" vs. "code point" issue, I don't have a strong preference, but I think we can continue to use the term "character", since it's convenient and widely understood. Serious confusion is unlikely, especially as we use the U+xxxx syntax to precisely define the exact ranges of code points (~ characters) allowed.

arp242:
> There's however the long-standing consensus that "idents" need to follow certain rules and generally those rules have been to exclude characters or code-points that aren't letterlike.

What is the motivation for this? And does this also apply to TOML?

I looked a bit at the xml:id specification, but I couldn't really find anything. I think it's important to look at the why of it rather than just copy things "because its how it's done". TOML is not a programming language or XML, so what does and doesn't work in one area can actually work just fine here.

For example, almost all programming languages require identifiers to start with a number; otherwise `1 + 1` could refer to both digits or variables and you need type information during parsing to figure that out. `-` is typically excluded as well, to prevent confusion with the minus operator, etc. But that doesn't apply to TOML, and we already allow keys to start with numbers and contain `-`, as well as some other more relaxed rules.

abelbraaksma:
> What is the motivation for this? And does this also apply to TOML?

If you don't, it becomes a mess. The 'M' in TOML stands for 'Minimal'. Using this fairly simple approach allows it to stay minimal. Using categories or hand waiving w.r.t. Unicode versions doesn't.

> I looked a bit at the xml:id specification, but I couldn't really find anything. 

The [`xml:id` spec](https://www.w3.org/TR/xml-id/) is a very simple spec that [defines it as an `NCName`](https://www.w3.org/TR/xml-id/#processing), which is a [`Name` without the `:`](https://www.w3.org/TR/REC-xml/#NT-Name). Conversely, such ids are used as [URI fragment identifiers](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax), which is most typically used to reference an ID or Name in an HTML, XML or like document (for which purpose they are restricted to the `NCName` production).

> just copy things "because its how it's done". 

We don't just copy things, we pick what's useful if it serves our purpose. We've had months of discussion before we came to this PR. Have a look at the original issue please.

> TOML is not a programming language or XML,

Nobody says TOML is a programming language. That doesn't mean we cannot borrow useful ideas elsewhere, whether it's from a supermarket, the Hitchhiker's Guide or ALGOL really doesn't matter. In this case we looked at past experiences, discussions and best practices in like fields.

> so what does and doesn't work in one area can actually work just fine here.

Exactly! That's why we came up with this simple, unobtrusive approach.

> For example, almost all programming languages require identifiers to start with a number

They don't, unless you mean labels.

> - is typically excluded as well,

It isn't. Like it isn't in DNS, HTML tag names, XPath and XQuery identifiers, UNC file names, paths in URLs etc, etc.

> But that doesn't apply to TOML, and we already allow keys to start with numbers and contain `-`, as well as some other more relaxed rules.

Exactly, which is precisely why we made the rules the way they are. Nobody is suggesting to change that.

To give a counter example that follows your idea of using Unicode categories: MS SQL Server. You cannot create a database name with characters defined since 2002, as once they chose this approach, it became impossible to ever let go. Since then, MS SQL Server has been limited to Unicode 3.2. If they want to change this, they'll end up in version hell (old scripts won't be compatible with new scripts anymore and v.v.).

abelbraaksma:
But tbh, I feel like we’re going a bit in circles. I’m at fault too, for repeating my arguments. 

Basically, there are two choices. We go for a simple approach that’s trivial to implement, is widely adopted  and is easy to explain and understand but has a remote chance of some day including character assignments that weren’t anticipated; or we go for an approach that at face value seems sensible, but requires significant expansion of the ABNF, is hard to read, hard to maintain, asks a lot of implementers (one already tried and called it a major pita) and locks us in with a Unicode version.

To me, that’s not a hard choice to make.

arp242:
I think there's some confusion here now. In my previous comment I was writing about the idea to allow *all* characters *except* a short list of symbols that would make parsing harder (`=`, `[`, `]`, etc.) or be greatly confusing (e.g. various space symbols). That is, I was continuing from my previous paragraph a few comments up:

> We can also just say "include everything except =, [, ., ], some (or all?) spaces", or something to that effect. Do we really need to exclude things like arrows or block drawing characters? It's perhaps kind of silly to use it, but then again so is using Linear-A or various emojis, and that's allowed now too. That would make this entire discussion moot. I can't really think of any serious downsides right now.

My previous comment wasn't talking about using Unicode categories – that is a done discussion at this point as far as I'm concerned, because I don't actually see a reason to restrict bare keys to just digits and letters. I probably didn't make that too clear, sorry. 

In TOML we already allow almost everything in quoted keys, so the only reason I can see to not allow something in bare keys if it makes parsing harder or if it has the potential to be greatly confusing (such as various forms of spaces, and *maybe* some hompglyps; not sure if that's worth the effort though).

The ABNF notation for this shouldn't be a huge deal, and implementing it shouldn't be either. Basically, it's just a modified table from what you wrote in this patch. Actually both the ABNF and implementation might be shorter and simpler.

> > What is the motivation for this? And does this also apply to TOML?
> If you don't, it becomes a mess. The 'M' in TOML stands for 'Minimal'. Using this fairly simple approach allows it to stay minimal. Using categories or hand waiving w.r.t. Unicode versions doesn't.
>
> > I looked a bit at the [xml:id spec](https://www.w3.org/TR/xml-id/)ification, but I couldn't really find anything.
> The xml:id spec is a very simple spec that [defines it as an NCName](https://www.w3.org/TR/xml-id/#processing), which is a [Name without the :](https://www.w3.org/TR/REC-xml/#NT-Name). Conversely, such ids are used as [URI fragment identifiers](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax), which is most typically used to reference an ID or Name in an HTML, XML or like document (for which purpose they are restricted to the NCName production).

Yeah, I looked at that, but unless I missed it (didn't read from cover to cover), it doesn't explain why these specific character ranges were chosen. Why are some characters excluded? For example it goes out of its way to forbid `×` and `÷`, but it's not clear to me why? I guess the intent was to exclude math symbols (for whatever reason), but does include various math symbols in those ranges, so I'm left wondering: what's the problem with people using `hello÷world = "value"` in TOML? Why would we want to forbid that? And if there is a good reason, then why is `hello₋world = "value"` allowed? It seems rather inconsistent to me, and should either allow or forbid both (but I see no reason to forbid either).

> > For example, almost all programming languages require identifiers to start with a number
>
> They don't, unless you mean labels.

I meant "forbid" instead of "require" 😅 Sorry, that was a confusing mixup.

abelbraaksma:
> My previous comment wasn't talking about using Unicode categories – that is a done discussion at this point 

Aha, I didn't get that, sorry.

> it doesn't explain why these specific character ranges were chosen. Why are some characters excluded?

That's explained in the XML specs, I quoted that earlier, see https://github.com/toml-lang/toml/issues/687#issuecomment-567490186.

> For example it goes out of its way to forbid × and ÷, but it's not clear to me why? I guess the intent was to exclude math symbols (for whatever reason), but does include various math symbols in those ranges

That's historical. All Math Symbols from Unicode 2.0 and 3.2 were excluded. This part of the spec was written in 2001 and then went into CR/PR status, which froze this definition. The symbols you talk about were added at a later date. We could certainly exclude those other ranges, but before you know it, you'll end up with a much more complicated production. Most well-known and oft-used symbols/separators etc are now excluded for the (imo) right reasons and the rest we should just ignore.

The discussion whether or not to include "everything" except a very small set should probably be done in a separate discussion/issue, as that wasn't the goal of the accepted change discussed in #687. Personally, I don't think there's much merit in including many well-known symbols, but ultimately that's for the larger community to decide.

arp242:
> That's explained in the XML specs, I quoted that earlier, see https://github.com/toml-lang/toml/issues/687#issuecomment-567490186.

I guess I didn't read it as such 😅

Anyway, I think they key thing here is that you're not really "restricting identifiers" as it's already unrestricted with quoted keys, so it seems to me it that this makes it a parsing/lexing issue only: you already need to deal with all of Unicode (including for syntax highlighters etc.) and implementation-wise it shouldn't be an issue anywhere I think (or at least: not more than it already is).

ChristianSi:
It might be helpful to revisit the start of #687 – the issue this PR is meant to fix. @marzer has put it perfectly:

> TOML's "bare key" syntax is too restrictive. People who regularly use characters from languages other than English should be able to do so in TOML keys without additional gymnastics.

So let's not get distracted. This is the point of this PR – the only point. It was never after allowing arbitrary symbols in bare keys – after all, we have quoted keys for that already. If you want to use spaces, punctuation or similar stuff in a key, put quotes around it. That's how it has always been, and that's perfectly fine.

But the problem is that right now, "bare keys can be made up of words written in any language, provided it's English" (or another language that uses just the basic Latin alphabet of 2x26 letters – there are a few, but not so many). That's not good, since TOML *should not discriminate* against other languages. So this is what is PR needs to fix: Allow arbitrary words from any language in bare keys. Allowing non-word characters was never the goal, and isn't needed (again: use quoted keys if you want that).

So let's keep on that track, and don't get distracted.

abelbraaksma:
> So let's keep on that track, and don't get distracted.

Agreed. On that note, can we all agree on the character vs code-point discussion above? I'm a little confused on whether or not I should tackle that in this PR or that we can leave it as is for now. I've no problem either way, but wouldn't like the issue to dangle just because if that. @ChristianSi, any thoughts on that? 

Other than that, I addressed all concerns mentioned and this may be ready to merge, but a final review might be good ;).

marzer:
> On that note, can we all agree on the character vs code-point discussion above?

Others have made the case that it's probably fine as-is, and I'm OK with that. Let's handle it separately if it becomes necessary.

toml.md
ChristianSi: This little suggestion is still open.

abelbraaksma: @ChristianSi, tx, I just noticed. Fixed now :).

toml.abnf
abelbraaksma: As discussed below, this is deliberate, we _want to_ include ranges that aren't yet defined so we don't need to tie in with any Unicode version and other arguments mentioned there. Per that discussion, I'll resolve this.

toml.md
abelbraaksma: Per the comment below by @marzer, we'll take this discussion to a new issue, if necessary, and leave the current status quo w.r.t. to the use of the word "character" as-is.

toml.md
abelbraaksma: This has now been reviewed again by @ChristianSi (who made a few suggestions w.r.t. formatting, which has been committed), so I'm closing this discussion.

abelbraaksma:
Latest commit fixes an oversight. When I was going over the allowed ranges, I realized that some number-like characters were excluded, while TOML explicitly allows these. This was also mentioned in the earlier ranges shown in the original request. This allows `①` and `⑩`. Note that `🄑`, `Ⅷ` were already allowed allowed now.

This commit also includes a few extra characters from the low ISO-8859-1 block. Reason is simple: while most characters have been given a new home elsewhere in Unicode, some haven't, mainly the superscript digits 1, 2 and 3, and the currency symbols (which we already allowed in higher blocks). Why allow `⁹` but not `¹`? Compare the [Unicode Block](https://en.wikipedia.org/wiki/Superscripts_and_Subscripts_(Unicode_block)), where these positions are empty.

The commented lines are now allowed, the rest already was allowed:
```toml
𝟘 = "zero"
𞥖 = "alam digit six"
𑁩 = "brahmi digit three"
８ = "fullwidth digit"
꧓ = "javanese three"
༩ = "tibetan nine"
๐๑๒๓ = "thai zero, one, two, three"
௫ = "tamil five"
৶ = "bengali currency denominator three"
୶ = "oriya fraction"
௲ = "tamil one thousand"
൝ = "mayalam three twentieths"
ᶨ = "phonetic superscript"
⁹ = "super nine"
₁ = "sub one"
⅚ = "5/6"
#¹ = "super one"
#³ = "three"
#① = "1"
#⑩ = "10"
#⑴ = "1"
↉ = "bla"
🄑 = "b"
Ⅷ = "8"
ↂ  = "Roman ten thousand"
```

toml.md
```diff
@@ -104,27 +104,39 @@ Keys
...
+**Bare keys** may contain any letter-like or number-like Unicode character from any Unicode script,
+as well as ASCII digits, dashes and underscores. Punctuation, spaces, arrows, box drawing 
+and private use characters are not allowed.  Note that bare keys are allowed to be
+composed of only ASCII digits, e.g. `1234`, but are always interpreted as strings.
+
+* From the ASCII characters, only `A-Z`, `a-z`, `0-9`, `_` and `-` are allowed.
+* From the rest of the ISO-8859-1 characters, allowed are U+B1 to U+B3, U+B9, U+BC to U+BE, U+C0 to U+D6, U+D8 to U+F6 and U+F8 to U+FF
+* All of U+0100-1FFF are allowed, except `;` (U+037E).
+* Characters U+200C, U+200D, U+203F, U+2040, U+2070 to U+218F, U+2460 to U+24FF, and U+2C00 to U+2FEF are allowed.
+* All characters from U+3001 and higher, except surrogates (U+D800 to U+D999), Private Use (U+E000 to U+F8FF, U+F0000 to U+100000) and process-internal use (U+FDD0 to U+FDEF).
```
ChristianSi, line 116: Sorry for nitpicking, but this list and the paragraph before it should be reflowed to correspond to the 80 character line length used in the TOML spec.

toml.md
```diff
@@ -104,27 +104,39 @@ Keys
...
+😂 = "value"
+汉语大字典 = "value"
+辭源 = "value"
+பெண்டிரேம் = "value"
 ```
 
 **Quoted keys** follow the exact same rules as either basic strings or literal
-strings and allow you to use a much broader set of key names. Best practice is
-to use bare keys except when absolutely necessary.
+strings and allow you to use any Unicode character in a key name, including spaces.
```
ChristianSi, line 131: This needs to be reflowed too.

ChristianSi (changes requested):

toml.md
abelbraaksma: I wasn’t aware, will fix, tx

abelbraaksma: Done

toml.md
abelbraaksma: Done, tx

abelbraaksma:
Unless anybody else has suggestions, I think this is good to be merged (but I admit that I don’t fully know what the policy is for merging into master).

marzer:
> Unless anybody else has suggestions, I think this is good to be merged (but I admit that I don’t fully know what the policy is for merging into master).

Looks good to me!

toml.md
```diff
@@ -104,27 +104,44 @@ Keys
...
+
+* From the ASCII characters, only `A-Z`, `a-z`, `0-9`, `_` and `-` are allowed.
+* From the rest of the ISO-8859-1 characters, allowed are U+B1 to U+B3, U+B9, 
+  U+BC to U+BE, U+C0 to U+D6, U+D8 to U+F6 and U+F8 to U+FF
+* All of U+0100-1FFF are allowed, except `;` (U+037E).
+* Characters U+200C, U+200D, U+203F, U+2040, U+2070 to U+218F, U+2460 to 
+  U+24FF, and U+2C00 to U+2FEF are allowed.
+* All characters from U+3001 and higher, except surrogates (U+D800 to U+D999),
+  Private Use (U+E000 to U+F8FF, U+F0000 to U+100000) and process-internal use
+  (U+FDD0 to U+FDEF).
```
ChristianSi, line 121: What's with U+FFFE (65534) and U+FFFF (65535)? They are excluded by the ABNF but, as far as I can see, allowed by the text.

Also, the upper limit of the private use area are different in the text (U+100000) vs. the ABNF comment (x10FFFF). I suppose it doesn't matter because Unicode effectively ends at x10FFFF, but it's still a bit confusing. (One could just say "and above" instead.)

ChristianSi (changes requested):

toml.md
abelbraaksma: @ChristianSi, the part about U+FFFE/FFFF is done. Let me know if you want to change the sentences to be inclusive all the way, except for exclusive in some cases as it is now.

abelbraaksma: Ah, I see now what you mean. Yes, I seem to have been a bit hand-wavy to make the comments not to cumbersome and the ranges easier understood. Apparently that leads to confusion. Private use are, officially, U+F0000..U+FFFFD and U+100000..U+10FFFD (codes ending on FFFE/FFFF are forbidden and considered non-characters here as well).

I think these are little known facts and we can leave it as is, right?

The range in the text is clearly a mistake, 100000 should be 10FFFF, thanks. Sorry for misreading that in your first comment.

abelbraaksma: Fixed. Please reopen if you think I should be more precise in the PUP ranges.

ChristianSi:
I'm a bit surprised that U+B1, the plus–minus sign (±) is included in the ISO-8859-1 range. Aren't mathematical symbols otherwise largely excluded?

abelbraaksma:
> I'm a bit surprised that U+B1, the plus–minus sign (±) is included in the ISO-8859-1 range. Aren't mathematical symbols otherwise largely excluded?

@ChristianSi Absolutely. That must have been a typo, the intention was to include the ISO-8859-1 (or Latin-1 Sup) superscript digits and fractions. Fixed now, thanks!

abelbraaksma:
@ChristianSi, I did double check those ranges, which were copied from the definitions linked to in the original issue, expanded with the circled digits, fractions etc, to be in line with what TOML users would expect. Thanks for reviewing!

marzer (changes requested):

toml.md
```diff
@@ -104,27 +104,44 @@ Keys
...
-ASCII digits, e.g. `1234`, but are always interpreted as strings.
+**Bare keys** may contain any letter-like or number-like Unicode character from
+any Unicode script, as well as ASCII digits, dashes and underscores. 
+Punctuation, spaces, arrows, box drawing and private use characters are not
+allowed.  Note that bare keys are allowed to be composed of only ASCII digits,
+e.g. `1234`, but are always interpreted as strings.
+
+* From the ASCII characters, only `A-Z`, `a-z`, `0-9`, `_` and `-` are allowed.
+* From the rest of the ISO-8859-1 characters, allowed are U+B2, U+B3, U+B9, 
+  U+BC to U+BE, U+C0 to U+D6, U+D8 to U+F6 and U+F8 to U+FF
```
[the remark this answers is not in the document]
abelbraaksma: Agreed!

toml.md
```diff
@@ -104,27 +104,44 @@ Keys
...
+* All of U+0100-1FFF are allowed, except `;` (U+037E).
+* Characters U+200C, U+200D, U+203F, U+2040, U+2070 to U+218F, U+2460 to 
+  U+24FF, and U+2C00 to U+2FEF are allowed.
+* All characters from U+3001 and higher, except surrogates (U+D800 to U+D999),
+  Private Use (U+E000 to U+F8FF, U+F0000 to U+10FFFF), non-characters (U+FFFE 
+  and U+FFFF) and process-internal use (U+FDD0 to U+FDEF).
```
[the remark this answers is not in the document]
abelbraaksma: Good point! I think I go for the current listing style, with your suggestion applied, but I'm not against a different wording of the paragraph.

abelbraaksma:
Fixed @marzer's suggestions, tx!

marzer:
@abelbraaksma I've just updated toml++'s suppoort for this feature to use this PR's version of the proposal, rather than the old category-based one, and you might be pleased to hear that it meant I could excise roughly 1000 lines of code solely devoted to unicode categorization, heh. Much simpler.

abelbraaksma:
> I could excise roughly 1000 lines of code solely devoted to unicode categorization, heh. Much simpler.

Awesome! Though I'm a little bit sad that all that work is now in vain... but good to hear that others don't have to write such logic ;).

marzer:
>Though I'm a little bit sad that all that work is now in vain...

"Deleted code is debugged code"

toml.abnf
```diff
@@ -46,19 +46,32 @@ comment = comment-start-symbol *non-eol
...
+
+unquoted-key = 1*unquoted-key-char
+unquoted-key-char = ALPHA / DIGIT / %x2D / %x5F         ; a-z A-Z 0-9 - _
+unquoted-key-char =/ %xB2 / %xB3 / %xB9 / %xBC-BE       ; superscript digits, fractions
+unquoted-key-char =/ %xC0-D6 / %xD8-F6 / %xF8-37D       ; non-symbol chars in Latin block
+unquoted-key-char =/ %x37F-1FFF                         ; exclude GREEK QUESTION MARK, which is basically a semi-colon
+unquoted-key-char =/ %x200C-200D / %x203F-2040          ; from General Punctuation Block, include the two tie symbols and ZWNJ, ZWJ
+unquoted-key-char =/ %x2070-218F / %x2460-24FF          ; include super-/subscripts, letterlike/numberlike forms, enclosed alphanumerics
+unquoted-key-char =/ %x2C00-2FEF / %x3001-D7FF          ; skip arrows, math, box drawing etc, skip 2FF0-3000 ideographic up/down markers and spaces
+unquoted-key-char =/ %xF900-FDCF / %xFDF0-FFFD          ; skip D800-D999 surrogate block, E000-F8FF Private Use area, FDD0-FDEF intended for process-internal use (unicode)
```
eksortso, line 64: The first part of the comment ought to be `skip D800-DFFF surrogate block`.

toml.md
```diff
@@ -104,27 +104,44 @@ Keys
...
+allowed.  Note that bare keys are allowed to be composed of only ASCII digits,
+e.g. `1234`, but are always interpreted as strings.
+
+* Of the ASCII characters, only `A-Z`, `a-z`, `0-9`, `_` and `-` are allowed.
+* Of the rest of the ISO-8859-1 characters, allowed are U+B2, U+B3, U+B9, 
+  U+BC to U+BE, U+C0 to U+D6, U+D8 to U+F6 and U+F8 to U+FF
+* All of U+0100-1FFF are allowed, except `;` (U+037E).
+* Characters U+200C, U+200D, U+203F, U+2040, U+2070 to U+218F, U+2460 to 
+  U+24FF, and U+2C00 to U+2FEF are allowed.
+* All characters from U+3001 and higher are allowed, except surrogates (U+D800 to U+D999),
```
eksortso, line 119: The end of the surrogates range is wrong. Can you change it to: `except surrogates (U+D800 to U+DFFF),`

eksortso (changes requested):
There are spots in `toml.md` and in comments in `toml.abnf` that misidentify the Unicode surrogates range. The range needs to be updated to U+D800 to U+DFFF in both places.

toml.abnf
abelbraaksma: Well spotted, fixed.

toml.md
abelbraaksma: @eksortso Well spotted, fixed.

abelbraaksma:
@eksortso, tx, a clear memory lapse between hex and dec counting. Fixed now ;).

eksortso:
@abelbraaksma Thanks. If it helps, I found myself typing 999 at one point when writing comments. Something deep there! :D

ChristianSi:
What's the status of this? I think it's ready for merge. @pradyunsg Care to take a look?

abelbraaksma:
@ChristianSi, afaik, I've addressed all review comments, so I think it's good to merge.

ChristianSi:
Has anyone seen a maintainer somewhere around here? @pradyunsg @mojombo

pradyunsg:
I'm here, and I can take a look at this over the weekend.

abelbraaksma:
> I'm here, and I can take a look at this over the weekend.

Great @pradyunsg. How did it go, did you get a chance to check? I'm back here too if you have questions on this PR ;).

ChristianSi:
Still no progress? I'm sad...

@mojombo What became of your initiative to widen the maintainership? Can we hope for any news on that in the foreseeable future?

toml.md
```diff
@@ -104,27 +104,44 @@ Keys
...
+
+* Of the ASCII characters, only `A-Z`, `a-z`, `0-9`, `_` and `-` are allowed.
+* Of the rest of the ISO-8859-1 characters, allowed are U+B2, U+B3, U+B9, 
+  U+BC to U+BE, U+C0 to U+D6, U+D8 to U+F6 and U+F8 to U+FF
+* All of U+0100-1FFF are allowed, except `;` (U+037E).
+* Characters U+200C, U+200D, U+203F, U+2040, U+2070 to U+218F, U+2460 to 
+  U+24FF, and U+2C00 to U+2FEF are allowed.
+* All characters from U+3001 and higher are allowed, except surrogates (U+D800 to U+DFFF),
+  Private Use (U+E000 to U+F8FF, U+F0000 to U+10FFFF), non-characters (U+FFFE 
+  and U+FFFF) and process-internal use (U+FDD0 to U+FDEF).
```
pradyunsg, line 121: I don't think we should be listing this content here. I think having plain digestable language here is much more important than having the spec mirror the details that the ABNF would then unambigously declare.

I notice that this was added since @ChristianSi said:

> As I see it, the written spec should be self-contained and one should not have to read the ABNF to find out what it actually means. 

And... I disagree with that. The whole point of the spec is to be human-readable and including such a dense description with unicode codepoints right at the start is... uhm... suboptimal.

abelbraaksma: @pradyunsg i can see your point. I’m not necessarily in favour of either approach. I could certainly replace the whole bulleted list with something that says “all natural letters and letterlike characters, all digits in all languages”.

Maybe you have something specific in mind?

abelbraaksma:
_(weird, just used GitHub online to merge solved conflicts and it needed a new branch name, and now the source branch name is a little weird, I thought such names could not be changed. Ah well…)

eksortso:
_(The branch name thing may be due to `main` now being the default branch. Nice how smooth that transition was.)_

abelbraaksma:
> (The branch name thing may be due to `main` now being the default branch. Nice how smooth that transition was.)

@eksortso any idea how to rename the branch name again? It’s now ‘update-changelog’, which doesn’t sit right with me. Or should I just not care? (the only way I know of is creating a new branch and deleting the old, but my prev mistake suggests renaming and sticking with PR is possible).

eksortso:
@abelbraaksma When I saw that `toml-lang/toml` changed its default branch, I verified that my fork had no other open branches or PRs, and on GitHub I renamed my `master` to `main` (which evidently just created `main` as a new branch) and then I set `main` to default. When done, a brief explanation of how to adjust my local repo was presented to me. It's a four-step process and I only understood three of the steps. To their credit, the GitHub developers anticipated this sort of change pretty well.

But it turns out I didn't need to follow those steps, because I was able to just clone my fork to my local machine anew, which was my plan anyway.

abelbraaksma:
Got it, tx. Prior to your comment, I didn't even know this change was made, I merged through the website and then it asked for a new branch name (I thought it would be just for the merge itself, going into my branch).

So yeah, I screwed that one up. Not sure I can change it back (to a more sensible name), but I'll check.

toml.md
marzer: Would it be sufficient to draw explicit attention to the ABNF here, as an inset or similar? Something like:

> ℹ&#xFE0F; _The exact ranges of allowed characters can be found in the [ABNF grammar file](https://github.com/toml-lang/toml/blob/main/toml.abnf)._

As it stands the ABNF is only mentioned as an afterthought wayyyyyy at the bottom of the spec document. If it is to be used as an authoritative complement to the spec, we probably should mention it earlier/more prominently.

abelbraaksma:
> _I don't think we should be listing this content here. I think having plain digestable language here is much more important_

@pradyunsg, perhaps just removing that part and, as suggested by @marzer above, adding a link to the ABNF for the curious?

I'd be fine with just this:

> **Bare keys** may contain any letter-like or number-like Unicode character from
> any Unicode script, as well as ASCII digits, dashes and underscores.
> Punctuation, spaces, arrows, box drawing and private use characters are not
> allowed. Note that bare keys are allowed to be composed of only ASCII digits,
> e.g. `1234`, but are always interpreted as strings.
>
> ℹ️ The exact ranges of allowed code points can be found in the [ABNF grammar file](https://github.com/toml-lang/toml/blob/main/toml.abnf).

Perhaps we can add a line something like the following (just a suggestion atm, I haven't really ironed out the wording here yet and am not completely happy with it):

> ℹ️ TOML allows any letter-like and number-like character is allowed in bare keys, including
> characters for almost all living and dead known languages, to embrace internationalization.
> However, some consoles or systems may not properly render certain esoteric scripts, showing
> little boxes or question marks instead. When deciding upon naming bare keys, using only the 
> ASCII ranges or lower Unicode ranges is preferred.

ChristianSi:
I like the first part of the proposed wording, if we don't wait to list the exact ranges in the spec.

I'd avoid the "suggestion", however. For one thing, the spec rarely suggests things or gives advice like that, and also it reads somewhat paternalistic to me. Supposedly format designers will only use languages they and their readers are familiar with, so font support won't be a problem for them – or, if it is, they'll know that already. Either way, the advice is unneeded.

abelbraaksma:
I am ok with either. I understand your point and you’re right, many languages support any Unicode character, or a subset, but I can’t recall they have this kind of advisory either. 

I’d like to hear @pradyunsg’s ideas, in particular to the first para in my proposal above. Then we can round this up ;).

pradyunsg:
I agree with @ChristianSi. The first paragraph works well enough, and the advisory can be omitted. We can change things once if we see people getting confused.

The only thing I'll say is that you should link to the ABNF file via the `[link text goes here][abnf]` syntax, since the release automation has some specific handling to ensure that link works on toml.io. :)

abelbraaksma:
@pradyunsg, thanks. Done. (PS: I assume you mean `(link text)[actual link]` format, right? Using `[link text][actual link]` doesn't seem to work).

toml.md
```diff
@@ -103,10 +103,9 @@ Punctuation, spaces, arrows, box drawing and private use characters are not
 allowed. Note that bare keys are allowed to be composed of only ASCII digits,
 e.g. 1234, but are always interpreted as strings.
 
-ℹ️ The exact ranges of allowed code points can be found in the 
+ℹ️ The exact ranges of allowed code points can be found in the
```
marzer, line 106: @abelbraaksma FYI if you want the 'info' emoji to show up consistently as the actual emoji and not a crap outline version
you need to follow it with `&#xfe0f;` version

Ref: https://stackoverflow.com/questions/72048059/how-to-specify-emoji-version-of-a-unicode-character-in-html

abelbraaksma: Ah, yes, that makes sense. Fixed.

marzer:
@abelbraaksma 

> Using [link text][actual link] doesn't seem to work).

That form is correct; it's the "reference form" ([ref](https://daringfireball.net/projects/markdown/syntax#link)). It requires you to explicitly list the link target in a separate reference list somewhere else in the document (typically right at the top or bottom). Example:

```md

# Hello!

[this is a link][SECRET]

[SECRET]: https://www.youtube.com/watch?v=dQw4w9WgXcQ

```

Produces this:

image

abelbraaksma:
@marzer, of course! 🤦‍♂️. How did I not get that 😆! Fixed.

ChristianSi:
More than 3 weeks and still no news? Is the maintainership bottleneck raising its ugly head again? :sob:

pradyunsg (approved):

marzer:
Hype hype hype, thanks everyone for driving this (particularly @abelbraaksma who actually Did The Thing)

abelbraaksma:
Thanks @pradyunsg for merging, thanks @marzer and @ChristianSi (and others!) for all the helpful comments. Glad we’re now going international!

epage:
Just wanting to catch up on this for when I need to implement this for my toml parser.  How does the selected ranges compare to [unicode's identifier definition](https://www.unicode.org/reports/tr31/) (Id_Start, Id_Continue, Xid_Start, Xid_Continue)?

If keeping the idea of "identifier" I believe this is what programming languages usually use and ideally we wouldn't be re-inventing the wheel.

abelbraaksma:
@epage, Unicode identifiers assume that identifiers don’t start with a digit (of any kind), which is usually correct, but not in TOML, which allows numerical identifiers.

While I did look at those definitions, they didn’t suit our needs. We settled on the simplest possible ranges that are easy to implement. The ones from Unicode are typically way more complex and depend on character properties.

ChristianSi:
TOML's rules for bare keys are pretty close to those used in XML for identifiers, IIRC.

PR #904, opened 2022-05-18
Allow newlines and trailing comma in inline tables

### LINKED ISSUE #516: Proposal: Allow newlines and trailing commas in inline tables
Overall I really like toml and its syntax feels very obvious to me for the most part. The only thing that doesn't is the expclicit cripling of inline tables, i.e. inline tables cannot have newlines or trailing commas. I've read the reasoning behind this in the existing [issue](https://github.com/toml-lang/toml/issues/219#issuecomment-48067853) and [PR](https://github.com/toml-lang/toml/pull/235). However, I don't think that the reason given (discouraging people from using big inline tables instead of sections) weighs up against the downsides. That's why I would like to open up a discussion about this.

There's three main downsides I see:

1. It's unexpected for most people using the language. Most popular languages that have `{}` style mappings allow newlines in them (JSON, Python, Javascript, Go). Also newlines and trailing commas are allowed in lists in the toml spec, so it is inconsistent in this regard.
2. To me even small inline tables are much more readable at first glance when split over multiple lines:

```toml
# Single line
person = { name = "Tom", geography = { lat = 1.0, lon = 2.0 } }

# Multi line
person = { 
    name = "Tom", 
    geography = { 
        lat = 1.0, 
        lon = 2.0,
    },
}
```
3. A deeply nested list of tables are forced to have a lot of repeated keys in the section headers. Compare this version with list of tables: 

```toml
[main_app.general_settings.logging]
log-lib = "logrus"

[[main_app.general_settings.logging.handlers]]
  name = "default"
  output = "stdout"
  level = "info"

[[main_app.general_settings.logging.handlers]]
  name = "stderr"
  output = "stderr"
  level = "error"

[[main_app.general_settings.logging.handlers]]
  name = "access"
  output = "/var/log/access.log"
  level = "info"
```

To the one with inline tables with newlines:
```toml
[main_app.general_settings.logging]
log-lib = "logrus"

handlers = [
    {
        name = "default",
        output =  "stdout",
        level = "info",
    }, {
        name = "stderr",
        output =  "stderr",
        level = "error",
    }, {
        name = "access",
        output =  "/var/log/access.log",
        level = "info",
    },
]
```

Finally, extending current toml parsers to support this is usually really easy, so that also shouldn't be an argument against it. I changed the the https://github.com/pelletier/go-toml implementation to support newlines in tables and I only had to change 5 lines to do it (3 of which I simply had to delete).

arp242:
Fixes #516

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -797,7 +797,9 @@ Inline Table
 ------------
 
 Inline tables provide a more compact syntax for expressing tables. They are
-especially useful for grouped data that can otherwise quickly become verbose.
+especially useful for grouped data that can otherwise quickly become verbose or
+(deeply) nested data.
+
 Inline tables are defined within curly braces: `{` and `}`. Within the braces,
 zero or more comma-separated key/value pairs may appear. Key/value pairs take
 the same form as key/value pairs in standard tables. All value types are
@@ -808,12 +810,19 @@ put on different lines. A terminating comma (also called trailing comma) is
 permitted after the last key/value pair.
 
 ```toml
-name = {
-  first = "Tom",
-  last = "Preston-Werner",
-}
-point = { x = 1, y = 2 }
+name = { first = "Tom", last = "Preston-Werner" }
+point = {x=1, y=2}
 animal = { type.name = "pug" }
+contact = {
+    personal = {
+        name = "Donald Duck",
+        email = "donald@duckburg.com",
+    },
+    work = {
+        name = "Coin cleaner",
+        email = "donald@ScroogeCorp.com",
+    },
+}
 ```
 
 The inline tables above are identical to the following standard table
@@ -830,6 +839,14 @@ y = 2
 
 [animal]
 type.name = "pug"
+
+[contact.personal]
+name = "Donald Duck"
+email = "donald@duckburg.com"
+
+[contact.work]
+name = "Coin cleaner"
+email = "donald@ScroogeCorp.com"
 ```
 
 Inline tables are fully self-contained and define all keys and sub-tables within
--- a/toml.md
+++ b/toml.md
@@ -798,20 +798,20 @@ Inline Table
 
 Inline tables provide a more compact syntax for expressing tables. They are
 especially useful for grouped data that can otherwise quickly become verbose.
-Inline tables are fully defined within curly braces: `{` and `}`. Within the
-braces, zero or more comma-separated key/value pairs may appear. Key/value pairs
-take the same form as key/value pairs in standard tables. All value types are
+Inline tables are defined within curly braces: `{` and `}`. Within the braces,
+zero or more comma-separated key/value pairs may appear. Key/value pairs take
+the same form as key/value pairs in standard tables. All value types are
 allowed, including inline tables.
 
-Inline tables are intended to appear on a single line. A terminating comma (also
-called trailing comma) is not permitted after the last key/value pair in an
-inline table. No newlines are allowed between the curly braces unless they are
-valid within a value. Even so, it is strongly discouraged to break an inline
-table onto multiples lines. If you find yourself gripped with this desire, it
-means you should be using standard tables.
+Inline tables can have multiple key/value pairs on the same line, or they can be
+put on different lines. A terminating comma (also called trailing comma) is
+permitted after the last key/value pair.
 
 ```toml
-name = { first = "Tom", last = "Preston-Werner" }
+name = {
+  first = "Tom",
+  last = "Preston-Werner",
+}
 point = { x = 1, y = 2 }
 animal = { type.name = "pug" }
 ```
--- a/toml.md
+++ b/toml.md
@@ -797,8 +797,8 @@ Inline Table
 ------------
 
 Inline tables provide a more compact syntax for expressing tables. They are
-especially useful for grouped data that can otherwise quickly become verbose or
-(deeply) nested data.
+especially useful for grouped (deeply) nested data that can otherwise quickly
+become verbose.
 
 Inline tables are defined within curly braces: `{` and `}`. Within the braces,
 zero or more comma-separated key/value pairs may appear. Key/value pairs take
--- a/toml.md
+++ b/toml.md
@@ -797,12 +797,12 @@ Inline Table
 ------------
 
 Inline tables provide a more compact syntax for expressing tables. They are
-especially useful for grouped (deeply) nested data that can otherwise quickly
-become verbose.
+especially useful for grouped nested data that can otherwise quickly become
+verbose.
 
-Inline tables are defined within curly braces: `{` and `}`. Within the braces,
-zero or more comma-separated key/value pairs may appear. Key/value pairs take
-the same form as key/value pairs in standard tables. All value types are
+Inline tables are fully defined within curly braces: `{` and `}`. Within the
+braces, zero or more comma-separated key/value pairs may appear. Key/value pairs
+take the same form as key/value pairs in standard tables. All value types are
 allowed, including inline tables.
 
 Inline tables can have multiple key/value pairs on the same line, or they can be
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## unreleased
 
+* Allow newline after key/values in inline tables.
+* Allow trailing comma in inline tables.
 * Clarify where and how dotted keys define tables.
 * Add new `\e` shorthand for the escape character.
 
--- a/toml.abnf
+++ b/toml.abnf
@@ -222,13 +222,14 @@ std-table-close = ws %x5D     ; ] Right square bracket
 
 ;; Inline Table
 
-inline-table = inline-table-open [ inline-table-keyvals ] inline-table-close
+inline-table = inline-table-open [ inline-table-keyvals ] ws-comment-newline inline-table-close
 
-inline-table-open  = %x7B ws     ; {
-inline-table-close = ws %x7D     ; }
-inline-table-sep   = ws %x2C ws  ; , Comma
+inline-table-open  = %x7B  ; {
+inline-table-close = %x7D  ; }
+inline-table-sep   = %x2C  ; , Comma
 
-inline-table-keyvals = keyval [ inline-table-sep inline-table-keyvals ]
+inline-table-keyvals =  ws-comment-newline keyval ws-comment-newline inline-table-sep inline-table-keyvals
+inline-table-keyvals =/ ws-comment-newline keyval ws-comment-newline [ inline-table-sep ]
 
 ;; Array Table
 


toml.md
```diff
@@ -798,20 +798,20 @@ Inline Table
...
+Inline tables can have multiple key/value pairs on the same line, or they can be
+put on different lines. A terminating comma (also called trailing comma) is
+permitted after the last key/value pair.
 
 ```toml
-name = { first = "Tom", last = "Preston-Werner" }
+name = {
+  first = "Tom",
+  last = "Preston-Werner",
+}
```
ChristianSi, line 814: I would suggest to leave the existing single-line example as is and add a second one, using multiple lines and a trailing comma, after it.

ChristianSi (changes requested):
I would not delete the existing example. Looks good otherwise!

eksortso:
Great work, @arp242. Thanks for stepping up!

Would it be worthwhile to mention, within the context of the spec as a whole, that the inline table's purpose is to make nesting simpler when needed? We threw out a chunk of formal editorializing, but I'd think just adding this one bit keeps inline tables in their proper perspective.

arp242:
I restored the previous example and added a new one; I spent a good few minutes staring at my screen trying to think of an example of nested data, and this was the best I could come up with right now 😅

I also added a mention of "nested" in the description.

toml.md
```diff
@@ -797,23 +797,32 @@ Inline Table
...
-table onto multiples lines. If you find yourself gripped with this desire, it
-means you should be using standard tables.
+Inline tables can have multiple key/value pairs on the same line, or they can be
+put on different lines. A terminating comma (also called trailing comma) is
+permitted after the last key/value pair.
 
 ```toml
 name = { first = "Tom", last = "Preston-Werner" }
-point = { x = 1, y = 2 }
+point = {x=1, y=2}
```
arp242, line 814: I intentionally removed the spaces here; I think it clarifies a bit that the spacing is all optional.

ChristianSi (approved):
Looks good to me!

@pradyunsg Care to take a look and merge, if you agree?

toml.md
```diff
@@ -797,7 +797,9 @@ Inline Table
 ------------
 
 Inline tables provide a more compact syntax for expressing tables. They are
-especially useful for grouped data that can otherwise quickly become verbose.
+especially useful for grouped data that can otherwise quickly become verbose or
+(deeply) nested data.
```
eksortso, line 801: This is pretty good. Could you make one subtle change to the last line? If you remove the word "data", it would then describe the types of grouped data that inline tables can help to manage.

arp242: I tweaked the wording now to:

>  They are especially useful for grouped (deeply) nested data that can otherwise quickly
become verbose.

which I felt was a better phrasing and should address your comment too.

eksortso (approved):
LGTM, you got this!

eksortso:
@pradyunsg @mojombo Please review this. It's ready to merge if it's acceptable to you.

toml.md
ChristianSi: I must admit that I liked @eksortso 's wording better. "grouped (deeply) nested data" just feels like too many adjectives in a row.

arp242: I preferred it exactly because it put all adjectives in one place, instead of "amending" it later.

I don't strongly care either way though; I'll leave it as is for now but I'll happily change it if eksortso or other people agree.

eksortso: Arguably, allowing newlines makes inline tables look less crowded, even if they aren't nested. So even with all the adjectives, it gets across the point that this change is about.

eksortso: Not to ignore the "grouped" part. That's also a feature, because the inline syntax puts subtables in the middle of a table block without breaking the block apart. Not being able to do so was an issue that visitors have expressed in the past.

marzer (approved):

eksortso:
@pradyunsg @mojombo Please respond. It's been over two months. If you would merge this PR, it would resolve issue #516, and it would address the deep-nesting issue that's been boiling for years. At the very least, please say something.

pradyunsg:
I’ve been swamped by a few IRL things, so I’ve not been looking to closely at this repository. I’ll make some time this week and catch up. 😅

eksortso:
Thanks, @pradyunsg. One way or another, this PR will be a big hit with us.

And speaking as a guy with IRL problems of his own, I will tell you now, I am grateful to you.

ChristianSi:
We are very understanding ... but only as long as you don't run away again! (insert threatening face)

toml.md
```diff
@@ -797,23 +797,32 @@ Inline Table
 ------------
 
 Inline tables provide a more compact syntax for expressing tables. They are
-especially useful for grouped data that can otherwise quickly become verbose.
-Inline tables are fully defined within curly braces: `{` and `}`. Within the
-braces, zero or more comma-separated key/value pairs may appear. Key/value pairs
-take the same form as key/value pairs in standard tables. All value types are
+especially useful for grouped (deeply) nested data that can otherwise quickly
+become verbose.
```
pradyunsg, line 801: I'd leave this as-is. The `deeply` isn't particularly valuable IMO.

toml.md
```diff
@@ -797,23 +797,32 @@ Inline Table
...
+
+Inline tables are defined within curly braces: `{` and `}`. Within the braces,
```
pradyunsg, line 803:
```suggestion
Inline tables are fully defined within curly braces: `{` and `}`. Within the braces,
```

The fully here is important, since it notes that you can't "inject" values into it.

pradyunsg (commented):
LGTM, with A couple of language nit picks.

pradyunsg:
> as long as you don't run away again! (insert threatening face)

Fair enough. :P

toml.md
arp242: Alrighty; I updated the PR.

pradyunsg: Oh, wow, thanks for the super quick response on this! :)

pradyunsg (approved):

toml.md
arp242: Happened to be on GitHub waiting for the CI to finish on another project :-)

pradyunsg:
Thanks for the PR @arp242 and for the discussions here everyone! :)

dmbaturin:
That's great to see! Two questions, though:

* Will there be a TOML 2.0 spec now to accomodate the change? When can we expect it?
* What do you folks think about best practices for tools that _output_ TOML? Should they offer a version switch and default to 1.0 so that older library/tool versions can read their output?

arp242:
> Will there be a TOML 2.0 spec now to accomodate the change? When can we expect it?

One thing that we need to do is write test cases in toml-test for all the new behaviours *before* we tag/release a new version. Once it's decided "yup, we've done everything we want to do for a new release" someone will need to spend some time on that (if someone hasn't already). There's a `-toml` flag now to set the version, so we can add as many tests for new behaviour as we want without breaking the 1.0 tests.

(Also: I assume that with "TOML 2.0" you mean "a new version"; I don't there will ever be an incompatible 2.0; the next version will most likely be 1.1).

> What do you folks think about best practices for tools that output TOML? Should they offer a version switch and default to 1.0 so that older library/tool versions can read their output?

That's up to you. I bet that >95% of TOML files are read by one application only.

dmbaturin:
>That's up to you. I bet that >95% of TOML files are read by one application only.

I'm talking about tools like linters and formatters. Their output is by definition meant to be read by other applications. However, the situation with newer TOML versions support in the ecosystem is dire: most libraries listed in the wiki still don't even support 1.0! I believe some written best practice advice is needed.

marzer:
> However, the situation with newer TOML versions support in the ecosystem is dire:

That's not a good-faith read of the situation, methinks. Practically all of the actively-maintained implementations are v1.0.0-compliant. The implementations listed on the wiki under v0.5.0 and below are basically all dead/abandoned projects.

arp242:
> most libraries listed in the wiki still don't even support 1.0

I'm not sure how up-to-date the wiki is, plus many libraries listed there are unmaintained. For example for C♯ it lists:

	v1.0.0 compliant

		7 months - 130 stars - https://github.com/dezhidki/Tommy
		6 days   -  57 stars - https://github.com/SamboyCoding/Tomlet
		1 month  - 218 stars - https://github.com/xoofx/Tomlyn 

	v0.5.0 compliant

		2 years -   4 stars - https://github.com/DanielArgent/Tomen
		2 years - 222 stars - https://github.com/paiden/Nett (archived)

	v0.4.0 compliant

		7 years - 7 stars - https://github.com/azyobuzin/HyperTomlProcessor 

	Unknown (or pre-v0.1.0) compliance

		 6 years -  16 stars - https://github.com/LBreedlove/Toml.net
		10 years -   6 stars - https://github.com/RichardVasquez/TomlDotNet
		 9 years -  43 stars - https://github.com/rossipedia/toml-net 

So basically, all the v0.5.0-compliant ones are unmaintained.

I just listed the C♯ ones because that happened to be the top one, but I bet the situation is similar for many other languages.

Someone should probably clean up the wiki a bit.

Other than that, I would say "use your best judgement". Different projects can make different reasonable decisions here: some might want to update to TOML 1.1 as soon as possible and that's completely fine, others might want to stick with 1.0 (or even 0.5) for a long time, and that's fine too. I don't think there is "one right way" to go about this.

eksortso:
For what it's worth, my take is that the changes don't merit a v2.0 because they don't break valid v1.0.0 configs. There are still significant open PRs like #891 that are backwards compatible, so perhaps a v1.1.0 is worth considering?

Output tools should write configs that the input tools can read. That's the be-all and end-all of it. Revisit the second question when parsers start to accept newlines in inline tables.

pradyunsg:
This will be in a 1.1.0. I'm not gonna commit to a timeline for that tho.

> What do you folks think about best practices for tools that _output_ TOML? Should they offer a version switch and default to 1.0 so that older library/tool versions can read their output?

Well, it depends. If they're style-preserving, sure -- preserve it. If they're regular dumps, and they can preserve 1.0 compat, it's reasonable too.

KaKi87:
> This will be in a 1.1.0. I'm not gonna commit to a timeline for that tho.

Hello, it's been more than a year since you said that, any news ? Thanks

eksortso:
Well this PR was merged, but we're discussing TOML 1.1.0 over at #928.

czy-29:
I have a question about this. Since we can already wrap lines in inline tables, is the comma at the end of each line still necessary?

ChristianSi:
@czy-29 For inline tables, commas between key-value pairs are still required, yes. Linebreaks are allowed, but ignored; they don't act as delimiters.

czy-29:
@ChristianSi What I want to say is not about the current actual implementation, but about the rationality and consistency of the design.
For regular key-value pairs and tables, the newline character can serve as a delimiter. However, when it comes to inline tables, the newline character cannot be used as a delimiter and is simply ignored. I believe this introduces inconsistency into the specification and increases user confusion and cognitive load.
I believe that multi-line inline tables should be consistent with multi-line key-value pairs in terms of specification.

ChristianSi:
@czy-29: However, inline tables and block-level tables will never use exactly the same syntax, unless you want to allow the latter to be separated by commas instead of newlines as well.

PR #796, opened 2020-12-19
Add \xHH Unicode escape code to basic strings

eksortso:
Per a conversation in #622, this PR adds a two-character \x escape code sequence for the first 256 Unicode code points. This provides a short explicit sequence for, among other code points: ASCII characters, certain control characters, and whitespace characters such as tab (\x09) and space (\x20) that need visible representation without resorting to longer \u and \U codes. Such sequences are commonly found in languages that also support the \u and \U codes for Unicode.

This PR replaces #709 and is intended for a post-v1.0.0 version of the TOML standard.

PR #910, opened 2022-06-18
Add a pre-commit configuration file and adopt Prettier

pradyunsg:
This makes it easier to set up linters that check for various things
within the repository.


pradyunsg:
Well, I've been procrastinating on this for a while. :)

pradyunsg:
I'm gonna leave this up for a bit, to let folks chime in with thoughts on this. I'm assuming no one's gonna be strongly opposed, but I'm open to being surprised. :)

marzer:
FYI this has conflicts with these PRs:
- #796: Add \xHH Unicode escape code to basic strings ([diff](https://github.com/toml-lang/toml/compare/manual-pre-commit-fixes...eksortso:unicode_escapes))
- #891: Allow non-English scripts for unquoted keys ([diff](https://github.com/toml-lang/toml/compare/manual-pre-commit-fixes...abelbraaksma:main))
- #894: Make seconds optional in times ([diff](https://github.com/toml-lang/toml/compare/manual-pre-commit-fixes...eksortso:i671_seconds_optional))
- #904: Allow newlines and trailing comma in inline tables ([diff](https://github.com/toml-lang/toml/compare/manual-pre-commit-fixes...arp242:multiline-itbl))

They've all been open quite a while at this stage. Can they be actioned first? Adding additional conflicts to them is only going to add (potentially many) more months worth of iteration to them.

ChristianSi:
I agree, let's finish the unfinished business first.

Also, to avoid the noise, I'd suggest to configure the pretty printer so as to make its output similar to the current Markdown style used, in order to just remove inconsistency rather than changing everything. In particular, this seems to mean:

* List items start with `*`.
* Underlining (`===`, `---`) is used for headers.

Otherwise this looks good.

eksortso:
> I agree, let's finish the unfinished business first.

Agreed.

> Also, to avoid the noise, I'd suggest to configure the pretty printer so as to make its output similar to the current Markdown style used
**_[[...]]_**

In the long run, it would make sense to use the default Markdown style. One less thing to worry about.

However, I think there's a way to configure Prettier with a TOML file. Correct me if I'm wrong, but the [Prettier docs](https://prettier.io/docs/en/configuration.html) say that `.prettierrc.toml` in the project's root folder can enforce our current style. I'm a big fan of dogfooding, so if we can get a `.prettierrc.toml` file set up, then let's do what @ChristianSi said.

ChristianSi:
IMHO there is no such thing as a "default Markdown style", at least not when it comes to stuff such `*` vs. `-` for lists or underlining vs. `#` for headers. Either is equally fine.

However, while Prettier is configurable to some degree, I cannot see any options that control how Markdown output is formatted. Maybe I'm just overlooking something?

If not, I'd suggest to look for another tool that's flexible enough to do a good job. This one, as far as I can tell, was mostly developed for JavaScript code, the Markdown support seems just a minor add-on and (I suspect) not all that well developed.

abelbraaksma:
I’ve no problem with having a standard style for markdown. I’d prefer `*` over `-` for bullets, but that’s just a personal preference. 

@pradyunsg, as others have already mentioned, there are quite a few completed, verified, agreed upon PRs open. Before we merge this one in, those others should be merged. If there’s anything we can do to help to make this easier, please let us know (including adding new maintainers).

pradyunsg:
> Can they be actioned first?

Done. ;)

#891 is gonna need more discussion, and spec language change -- so I don't wanna block on that.

> [`*` vs. `-` for lists, underlines vs. `#` for headers, keeping consistency with status quo]

As far as I am concerned, having an automated tool do the formatting for me is super useful, because -- as long as the output is understandable and "sane" -- it's very useful to have the tool enforce a consistent style across the project wherein no one needs to debate about style choices.

The defaults of prettier are workable, with the line length enforcement being useful for easier-to-review diffs on the GitHub UI (which doesn't do too well with long lines, but highlights changed words).

arp242:
Would be nice to document what the max line length is now; it seems shorter than 80 based on this patch; it's now 79? 78? 76?

Can maybe also add .editorconfig:

    root = true
    [*.md]
    max_line_length = 78  # Or whatever it is.
    indent_style = space  # Might as well add this too.

pradyunsg:
It's 80, which is the default -- https://prettier.io/docs/en/options.html#print-width

abelbraaksma:
> https://github.com/toml-lang/toml/pull/891 is gonna need more discussion, and spec language change -- so I don't wanna block on that.

No worries, I’ll be able to deal with some reformatting there.

Issue #870, opened 2022-01-24
TOML as an RFC?


mnot:
Have you considered publishing v1.0.0 as an [RFC](https://rfc-editor.org/)?

I can see two immediate benefits:

* The format would be easier to use / reference from other RFCs and by other standards efforts.
* You could get `application/toml` [registered](https://www.iana.org/assignments/media-types/media-types.xhtml)[^1].

If you'd consider it, I'd be happy to help - you don't have to go through the full standards process, the independent stream would be very appropriate for this and comparatively lightweight. 

[^1]: You don't need an RFC for the full format to get the registration -- but it does require some sort of standards publication.

paulehoffman:
+1 to the "happy to help". @mnot and I both have written dozens of RFCs (although none together, yet).

pradyunsg:
> Have you considered publishing v1.0.0 as an [RFC](https://rfc-editor.org/)?

Yes! That's the motivation behind keeping the language grammer in ANBF.

IIUC, the blocker here is one of the authors figuring out the "paperwork" of the process and doing it. Help would very much be welcome! :)

paulehoffman:
If one of the main contributors wants to step up as the primary author, I'm happy to walk you through the paperwork. @mnot will probably do so as well, and may agree with me on some of it. :-)

mnot:
Definitely. Consider us at your disposal. From our side, we just need buy-in from the TOML project and someone to work with; we can handle the process issues, etc.

eksortso:
I'm just a bit player here. But with a number of serious enhancements in the pipeline, TOML v1.1.0 would be a better candidate for an RFC once it's released. Is there an immediate demand for an RFC to be filed now, or can we take the time to propose one with TOML's new features?

Deferring until the v1.1.0 release would also give us an opportunity to move toward a set of documents that would line up more with the requirements of an RFC, where it would make sense to further that process in the project as it stands.

ChristianSi:
Yes, I agree. The changes made since TOML 1.0 are significant and certainly justify the release of 1.1 in the foreseeable future. So, to be really future-proof, that's a better target for standardization.

mcarans:
Is there a plan for releasing 1.1? I think it would be a good prompt for library implementers to support the significant changes that have been made since 1.0. That in turn will hopefully prompt some developers who've been reticent to use TOML (and specifically pyproject.toml) to reconsider since I think the changes overcome concerns or limitations that they had raised.

eksortso:
[...]

Issue #538, opened 2018-05-18
Define appropriate implementation behaviors for extreme values


jongiddy:
One thing that I would like to see defined before 1.0 is the behavior when implementation limits are reached. This would consist of an explicit statement that TOML itself does not have a limit but that parsers may.

For example:
> Strings can have any length. However, parsers may produce an error for strings that are too long for them to safely handle.

> Integers can have any length. Parsers must produce an error if the value cannot be represented in an internal type. Parsers must not produce an integer that does not equal the full value represented in the TOML and must not produce a string representation of the value.

And perhaps most debatable, what happens to very large or very small floats (e.g. 1e500 and 1e-500). Programming languages often convert these to `inf` or 0. For configuration, I'd consider it more appropriate to produce an error. For example, a very small float may represent an epsilon value or error allowance, and should produce an error rather than be rounded down to zero.

> Numeric floats do not convert to `±inf`. The parser should produce an error if the float is too large to be represented as a non-infinite float.
> Floats that contain a non-zero mantissa do not convert to zero. The parser should produce an error if the float is too small to be represented as a non-zero float.

jongiddy:
@eksortso I think it would be useful to say that an integer outside the supported ranges should give an error, rather than return a different integer, which may happen for languages that have error-free wrapping, or return a truncated integer calculated up until a parse error occurs.

Instead of producing an error, I'd support the parser returning the correct integer where the language can support it (e.g. Python).

If that latter behavior is allowed, and it is clear that an error must be produced otherwise, then the 64-bit signed value could be defined as the guaranteed minimum that a parser should support.

Defining these things specifically encourages parser developers to test for these cases and also ensures parser implementations are more interchangeable

jongiddy:
Both the Python TOML parsers listed on the Wiki ("specs-conforming and strict" `pytoml` and "passes the TOML test suite" `toml`) support integers outside the 64-bit signed range, so parser developers already interpret the statement about integers as a minimum range.

This also emphasises that the word "expected" has almost no meaning in a standard.

For floats, both Python libraries convert `1e500` to `inf` and `1e-500` to `0.0`. Other languages may return the maximum non-infinite float or produce an error. For simplicity of building a parser, it may be best to leave the behavior of floats undefined.

pradyunsg:
Are there existing real world problems/issues caused by the lack of strict language on the behavior at these extreme values?

pradyunsg:
Some notes:
[...]

PR #924, opened 2022-10-11
Permit more control characters in comments

eksortso:
In reconsideration of the limitation of control codes within comments discussed in #567, a case was made to simplify comment parsing. This PR intends to replace those previously imposed restrictions with a simplified approach that permits anything except newlines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 
-- Relax comment parsing; control characters are again permitted.
+- Relax comment parsing; most control characters are again permitted.
 - Allow newline after key/values in inline tables.
 - Allow trailing comma in inline tables.
 - Clarify where and how dotted keys define tables.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## unreleased
 
+- Relax comment parsing; control characters are again permitted.
 - Allow newline after key/values in inline tables.
 - Allow trailing comma in inline tables.
 - Clarify where and how dotted keys define tables.
--- a/toml.abnf
+++ b/toml.abnf
@@ -83,8 +83,8 @@ quotation-mark = %x22            ; "
 
 basic-char = basic-unescaped / escaped
 basic-unescaped = wschar / %x21 / %x23-5B / %x5D-7E / non-ascii
-
 escaped = escape escape-seq-char
+
 escape = %x5C                   ; \
 escape-seq-char =  %x22         ; "    quotation mark  U+0022
 escape-seq-char =/ %x5C         ; \    reverse solidus U+005C
--- a/toml.abnf
+++ b/toml.abnf
@@ -39,7 +39,8 @@ newline =/ %x0D.0A  ; CRLF
 
 comment = comment-start-symbol *allowed-comment-char
 comment-start-symbol = %x23 ; #
-allowed-comment-char = %x01-%09 / %0E-10FFFF
+allowed-comment-char = %x01-09 / %x0E-7F / non-ascii
+non-ascii = %x80-D7FF / %xE000-10FFFF
 
 ;; Key-Value pairs
 
@@ -82,7 +83,6 @@ quotation-mark = %x22            ; "
 
 basic-char = basic-unescaped / escaped
 basic-unescaped = wschar / %x21 / %x23-5B / %x5D-7E / non-ascii
-non-ascii = %x80-D7FF / %xE000-10FFFF
 
 escaped = escape escape-seq-char
 escape = %x5C                   ; \
--- a/toml.abnf
+++ b/toml.abnf
@@ -39,7 +39,7 @@ newline =/ %x0D.0A  ; CRLF
 
 comment-start-symbol = %x23 ; #
 non-ascii = %x80-D7FF / %xE000-10FFFF
-non-linefeed = %x00-09 / %x0B-7F / non-ascii
+non-linefeed = %x00-09 / %x0B-0C / %x0E-7F / non-ascii
 
 comment = comment-start-symbol *non-linefeed
 
--- a/toml.abnf
+++ b/toml.abnf
@@ -39,9 +39,9 @@ newline =/ %x0D.0A  ; CRLF
 
 comment-start-symbol = %x23 ; #
 non-ascii = %x80-D7FF / %xE000-10FFFF
-non-eol = %x00-09 / %x0B-7F / non-ascii
+non-linefeed = %x00-09 / %x0B-7F / non-ascii
 
-comment = comment-start-symbol *non-eol
+comment = comment-start-symbol *non-linefeed
 
 ;; Key-Value pairs
 
--- a/toml.abnf
+++ b/toml.abnf
@@ -39,7 +39,7 @@ newline =/ %x0D.0A  ; CRLF
 
 comment-start-symbol = %x23 ; #
 non-ascii = %x80-D7FF / %xE000-10FFFF
-non-eol = %x09 / %x20-7E / non-ascii
+non-eol = %x00-09 / %x0B-7F / non-ascii
 
 comment = comment-start-symbol *non-eol
 
--- a/toml.abnf
+++ b/toml.abnf
@@ -37,11 +37,9 @@ newline =/ %x0D.0A  ; CRLF
 
 ;; Comment
 
+comment = comment-start-symbol *allowed-comment-char
 comment-start-symbol = %x23 ; #
-non-ascii = %x80-D7FF / %xE000-10FFFF
-non-linefeed = %x00-09 / %x0B-0C / %x0E-7F / non-ascii
-
-comment = comment-start-symbol *non-linefeed
+allowed-comment-char = %x01-%09 / %0E-10FFFF
 
 ;; Key-Value pairs
 
@@ -84,8 +82,9 @@ quotation-mark = %x22            ; "
 
 basic-char = basic-unescaped / escaped
 basic-unescaped = wschar / %x21 / %x23-5B / %x5D-7E / non-ascii
-escaped = escape escape-seq-char
+non-ascii = %x80-D7FF / %xE000-10FFFF
 
+escaped = escape escape-seq-char
 escape = %x5C                   ; \
 escape-seq-char =  %x22         ; "    quotation mark  U+0022
 escape-seq-char =/ %x5C         ; \    reverse solidus U+005C
--- a/toml.md
+++ b/toml.md
@@ -38,14 +38,13 @@ should be easy to parse into data structures in a wide variety of languages.
 
 - TOML is case-sensitive.
 - A TOML file must be a valid UTF-8 encoded Unicode document.
-- Whitespace means tab (0x09) or space (0x20).
-- Newline means LF (0x0A) or CRLF (0x0D 0x0A).
+- Whitespace means tab (U+0009) or space (U+0020).
+- Newline means LF (U+000A) or CRLF (U+000D U+0x0A).
 
 ## Comment
 
 A hash symbol marks the rest of the line as a comment, except when inside a
-string. All valid code points are permitted in a comment, but keep in mind that
-a LF or CRLF terminates the comment.
+string. All valid code points except CR (U+000D) are permitted in a comment.
 
 ```toml
 # This is a full-line comment
--- a/toml.md
+++ b/toml.md
@@ -36,15 +36,10 @@ should be easy to parse into data structures in a wide variety of languages.
 
 ## Spec
 
-A TOML file must be a valid UTF-8 encoded Unicode document. Specifically this
-means that, should a file as a whole not form a
-[well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence),
-the file must be rejected (preferably) or ill-formed byte sequences must be
-replaced with U+FFFD as per the Unicode spec.
-
 - TOML is case-sensitive.
-- Whitespace means tab (U+0009) or space (U+0020).
-- Newline means LF (U+000A) or CRLF (U+000D U+000A).
+- A TOML file must be a valid UTF-8 encoded Unicode document.
+- Whitespace means tab (0x09) or space (0x20).
+- Newline means LF (0x0A) or CRLF (0x0D 0x0A).
 
 ## Comment
 
@@ -269,7 +264,7 @@ The above TOML maps to the following JSON.
 ## String
 
 There are four ways to express strings: basic, multi-line basic, literal, and
-multi-line literal. All strings must contain only Unicode characters.
+multi-line literal. All strings must contain only valid Unicode characters.
 
 **Basic strings** are surrounded by quotation marks (`"`). Any Unicode character
 may be used except those that must be escaped: quotation mark, backslash, and
@@ -297,7 +292,7 @@ For convenience, some popular characters have a compact escape sequence.
 ```
 
 Any Unicode character may be escaped with the `\xHH`, `\uHHHH`, or `\UHHHHHHHH`
-forms. The escape codes must be Unicode
+forms. The escape codes must be valid Unicode
 [scalar values](https://unicode.org/glossary/#unicode_scalar_value).
 
 All other escape sequences not listed above are reserved; if they are used, TOML
@@ -421,9 +416,8 @@ str = ''''That,' she said, 'is still pointless.''''
 ```
 
 Control characters other than tab are not permitted in a literal string. Thus,
-for binary data, it is recommended that you use Base64 or another suitable
-binary-to-text encoding. The handling of that encoding will be
-application-specific.
+for binary data, it is recommended that you use Base64 or another suitable ASCII
+or UTF-8 encoding. The handling of that encoding will be application-specific.
 
 ## Integer
 
--- a/toml.md
+++ b/toml.md
@@ -52,8 +52,9 @@ key = "value"  # This is a comment at the end of a line
 another = "# This is not a comment"
 ```
 
-Control characters other than tab (U+0000 to U+0008, U+000A to U+001F, U+007F)
-are not permitted in comments.
+All characters except line feeds (0x0A) are permitted in comments and may be
+ignored. At their discretion, parsers that read comments may exclude a final
+carriage return (0x0D) appearing before a terminating line feed.
 
 ## Key/Value Pair
 
--- a/toml.md
+++ b/toml.md
@@ -37,10 +37,10 @@ should be easy to parse into data structures in a wide variety of languages.
 ## Spec
 
 A TOML file must be a valid UTF-8 encoded Unicode document. Specifically this
-means that, should a file as a whole not form a [well-formed code-unit
-sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence), the
-file must be rejected (preferably) or ill-formed byte sequences must be replaced
-with U+FFFD as per the Unicode spec.
+means that, should a file as a whole not form a
+[well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence),
+the file must be rejected (preferably) or ill-formed byte sequences must be
+replaced with U+FFFD as per the Unicode spec.
 
 - TOML is case-sensitive.
 - Whitespace means tab (U+0009) or space (U+0020).
--- a/toml.md
+++ b/toml.md
@@ -36,16 +36,18 @@ should be easy to parse into data structures in a wide variety of languages.
 
 ## Spec
 
+A TOML file must be a valid UTF-8 encoded Unicode document.
+
 - TOML is case-sensitive.
-- A TOML file must be a valid UTF-8 encoded Unicode document.
 - Whitespace means tab (U+0009) or space (U+0020).
 - Newline means LF (U+000A) or CRLF (U+000D U+000A).
 
 ## Comment
 
 A hash symbol marks the rest of the line as a comment, except when inside a
-string. All valid code points except NUL (U+0000) and historic line-breaking
-codes (U+000A through U+000D) are permitted in a comment.
+string. Comments may contain any valid code points except a limited subset of
+ASCII control codes that could cause problems during editing or processing
+(specifically, U+0000, and U+000A to U+000D).
 
 ```toml
 # This is a full-line comment
--- a/toml.md
+++ b/toml.md
@@ -37,9 +37,10 @@ should be easy to parse into data structures in a wide variety of languages.
 ## Spec
 
 A TOML file must be a valid UTF-8 encoded Unicode document. Specifically this
-means that, should a file as a whole not form a well-formed code-unit sequence,
-the file must be rejected (preferably) or ill-formed byte sequences must be
-replaced with U+FFFD as per the Unicode spec.
+means that, should a file as a whole not form a [well-formed code-unit
+sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence), the
+file must be rejected (preferably) or ill-formed byte sequences must be replaced
+with U+FFFD as per the Unicode spec.
 
 - TOML is case-sensitive.
 - Whitespace means tab (U+0009) or space (U+0020).
@@ -420,8 +421,9 @@ str = ''''That,' she said, 'is still pointless.''''
 ```
 
 Control characters other than tab are not permitted in a literal string. Thus,
-for binary data, it is recommended that you use Base64 or another suitable text
-encoding. The handling of that encoding will be application-specific.
+for binary data, it is recommended that you use Base64 or another suitable
+binary-to-text encoding. The handling of that encoding will be
+application-specific.
 
 ## Integer
 
--- a/toml.md
+++ b/toml.md
@@ -44,7 +44,8 @@ should be easy to parse into data structures in a wide variety of languages.
 ## Comment
 
 A hash symbol marks the rest of the line as a comment, except when inside a
-string. All valid code points except CR (U+000D) are permitted in a comment.
+string. All valid code points except NUL (U+0000) and historic line-breaking
+codes (U+000A through U+000D) are permitted in a comment.
 
 ```toml
 # This is a full-line comment
--- a/toml.md
+++ b/toml.md
@@ -39,7 +39,7 @@ should be easy to parse into data structures in a wide variety of languages.
 - TOML is case-sensitive.
 - A TOML file must be a valid UTF-8 encoded Unicode document.
 - Whitespace means tab (U+0009) or space (U+0020).
-- Newline means LF (U+000A) or CRLF (U+000D U+0x0A).
+- Newline means LF (U+000A) or CRLF (U+000D U+000A).
 
 ## Comment
 
--- a/toml.md
+++ b/toml.md
@@ -36,7 +36,10 @@ should be easy to parse into data structures in a wide variety of languages.
 
 ## Spec
 
-A TOML file must be a valid UTF-8 encoded Unicode document.
+A TOML file must be a valid UTF-8 encoded Unicode document. Specifically this
+means that, should a file as a whole not form a well-formed code-unit sequence,
+the file must be rejected (preferably) or ill-formed byte sequences must be
+replaced with U+FFFD as per the Unicode spec.
 
 - TOML is case-sensitive.
 - Whitespace means tab (U+0009) or space (U+0020).
@@ -45,9 +48,9 @@ A TOML file must be a valid UTF-8 encoded Unicode document.
 ## Comment
 
 A hash symbol marks the rest of the line as a comment, except when inside a
-string. Comments may contain any valid code points except a limited subset of
-ASCII control codes that could cause problems during editing or processing
-(specifically, U+0000, and U+000A to U+000D).
+string. Comments may contain any Unicode code points except the following
+control codes that could cause problems during editing or processing: U+0000,
+and U+000A to U+000D.
 
 ```toml
 # This is a full-line comment
@@ -265,7 +268,7 @@ The above TOML maps to the following JSON.
 ## String
 
 There are four ways to express strings: basic, multi-line basic, literal, and
-multi-line literal. All strings must contain only valid UTF-8 characters.
+multi-line literal. All strings must contain only Unicode characters.
 
 **Basic strings** are surrounded by quotation marks (`"`). Any Unicode character
 may be used except those that must be escaped: quotation mark, backslash, and
@@ -293,8 +296,8 @@ For convenience, some popular characters have a compact escape sequence.
 ```
 
 Any Unicode character may be escaped with the `\xHH`, `\uHHHH`, or `\UHHHHHHHH`
-forms. The escape codes must be valid Unicode
-[scalar values](https://unicode.org/glossary/#unicode_scalar_value).
+forms. The escape codes must be Unicode [scalar
+values](https://unicode.org/glossary/#unicode_scalar_value).
 
 All other escape sequences not listed above are reserved; if they are used, TOML
 should produce an error.
@@ -417,8 +420,8 @@ str = ''''That,' she said, 'is still pointless.''''
 ```
 
 Control characters other than tab are not permitted in a literal string. Thus,
-for binary data, it is recommended that you use Base64 or another suitable ASCII
-or UTF-8 encoding. The handling of that encoding will be application-specific.
+for binary data, it is recommended that you use Base64 or another suitable text
+encoding. The handling of that encoding will be application-specific.
 
 ## Integer
 
--- a/toml.md
+++ b/toml.md
@@ -52,9 +52,10 @@ key = "value"  # This is a comment at the end of a line
 another = "# This is not a comment"
 ```
 
-Comments may contain any Unicode code points except the following
-control codes that could cause problems during editing or processing: U+0000,
-and U+000A to U+000D.
+Comments may contain any Unicode code points except the following control codes
+that could cause problems during editing or processing: U+0000, and U+000A to
+U+000D.
+
 ## Key/Value Pair
 
 The primary building block of a TOML document is the key/value pair.
--- a/toml.md
+++ b/toml.md
@@ -44,7 +44,8 @@ should be easy to parse into data structures in a wide variety of languages.
 ## Comment
 
 A hash symbol marks the rest of the line as a comment, except when inside a
-string. All code points except line feeds (U+000A) are permitted in comments.
+string. All valid code points are permitted in a comment, but keep in mind that
+a LF or CRLF terminates the comment.
 
 ```toml
 # This is a full-line comment
--- a/toml.md
+++ b/toml.md
@@ -296,8 +296,8 @@ For convenience, some popular characters have a compact escape sequence.
 ```
 
 Any Unicode character may be escaped with the `\xHH`, `\uHHHH`, or `\UHHHHHHHH`
-forms. The escape codes must be Unicode [scalar
-values](https://unicode.org/glossary/#unicode_scalar_value).
+forms. The escape codes must be Unicode
+[scalar values](https://unicode.org/glossary/#unicode_scalar_value).
 
 All other escape sequences not listed above are reserved; if they are used, TOML
 should produce an error.
--- a/toml.md
+++ b/toml.md
@@ -44,17 +44,13 @@ should be easy to parse into data structures in a wide variety of languages.
 ## Comment
 
 A hash symbol marks the rest of the line as a comment, except when inside a
-string.
+string. All code points except line feeds (U+000A) are permitted in comments.
 
 ```toml
 # This is a full-line comment
 key = "value"  # This is a comment at the end of a line
 another = "# This is not a comment"
 ```
-
-All characters except line feeds (0x0A) are permitted in comments and may be
-ignored. At their discretion, parsers that read comments may exclude a final
-carriage return (0x0D) appearing before a terminating line feed.
 
 ## Key/Value Pair
 
--- a/toml.md
+++ b/toml.md
@@ -44,9 +44,7 @@ should be easy to parse into data structures in a wide variety of languages.
 ## Comment
 
 A hash symbol marks the rest of the line as a comment, except when inside a
-string. Comments may contain any Unicode code points except the following
-control codes that could cause problems during editing or processing: U+0000,
-and U+000A to U+000D.
+string.
 
 ```toml
 # This is a full-line comment
@@ -54,6 +52,9 @@ key = "value"  # This is a comment at the end of a line
 another = "# This is not a comment"
 ```
 
+Comments may contain any Unicode code points except the following
+control codes that could cause problems during editing or processing: U+0000,
+and U+000A to U+000D.
 ## Key/Value Pair
 
 The primary building block of a TOML document is the key/value pair.


ChristianSi:
So I would suggest to change the wording to:

> All characters except line feeds (0x0A) and carriage return (0x0D) are permitted in comments and will be ignored.

The ABNF would need to be adjusted as well.

arp242:
Also should to use U+ syntax instead of 0x for codepoints.

The way it's phrased now is kinda confusing; U+0A is the newline; this is already allowed in comments, because every line ends with a comment.

In general, I don't really see the point of this though; no one raised any practical concerns over this ("I want to do X because I can't because the specification limits me") and it just seems to introduce needless churn for implementers who will need to update the implementations, tests, etc. I would not be in favour of merging this.

arp242:
> @arp242 This PR addresses a number of points raised by @abelbraaksma in #567, specifically getting rid of (almost) all restrictions on characters permitted within comments, because we do not want a document rejected just because a control character shows up in a comment.
> 
> And if you've ever copy/pasted Excel cells into a text widget before, you do come across stray CR characters far too often.

Yes, I read that; but I don't see any practical problems. And both LF and CRLF end of lines are already allowed; if the ABNF excludes that then that's a bug in the ABNF.

eksortso:
> And both LF and CRLF end of lines are already allowed; if the ABNF excludes that then that's a bug in the ABNF.

A CR LF line ending is still permitted, because if you follow the ABNF to the letter, the CR in the newline is considered part of the content of the comment; it's all in `comment` and gets picked up in `*non-eol`. What remains is the LF, which is still a `newline`. So the ABNF is not ambiguous. If you are preserving comments, you can exclude the final CR in the `comment` if the `newline` immediately follows.

eksortso:
@ChristianSi You said:
> Maybe it would be better to express this as something like:
> 
> > All valid code points are permitted in comments (but keep in mind that a linebreak – LF or CRLF – terminates the comment).

That makes sense. In fact, let's remove the parentheses. The definition of a newline immediately precedes this section, so it ought to be clear why we mention LF and CRLF. Here's my reworked version.

> All valid code points are permitted in a comment, but keep in mind that a LF
> or CRLF terminates the comment.

I weighed the value of keeping the word "valid" in there. I was presuming that the document was already verified as being valid UTF-8, meaning no illegal bytes and no byte streams representing surrogate code points (among others) would be present. If a TOML document is read into a string for processing, then validation is performed before the hash symbol is even processed as the start of a comment. But you are coming from a perspective in which the validity of the document's byte stream is yet to be verified. For languages that live close to the bare metal, that could be the case. But surrogates are always invalid in UTF-8, and if they are present in any text document, then that's a bigger problem than a TOML parser ought to handle. Certainly bigger than isolated CRs from Excel.

Invalid byte streams, including surrogate code points, should be rejected outright. Parsers should _not_ tolerate them in comments, as they should not tolerate invalid UTF-8 documents in general. But we can still say "valid" code points, so there's no ambiguity whatsoever.

I will make one modification to the ABNF: rename `non-eol` to `non-linefeed`. It isn't used anywhere else, and the new name doesn't presume the line ending. I tested it in [Instaparse Live](http://instaparse.mojombo.com/), which was smart enough not to suck in the CR in the CRLF into the comment.

hukkin:
As an implementation maintainer I'm slightly opposed to this change simply because it seems this isn't something that users are asking for but rather something that `toml-lang` contributors _think_ would be an improvement. Seems like potential busywork for implementation maintainers.

I'm also concerned about allowing lone CR in comments. Assuming that text editors render lone CR as a line ending, I think it creates an opportunity to craft valid TOML documents that look different to human eye than a compliant parser. For example:

```toml
# All line endings in this document are lone CRs
a = "hello"
b = "world"
```
Assuming that all line endings in the document are CRs, the document is still valid. It is a single line TOML document consisting of one comment. I.e. the content, when read by a compliant parser is an empty mapping `{}`. To human eye it looks like `{"a": "hello", "b": "world"}` though.

eksortso:
> Assuming that all line endings in the document are CRs, the document is still valid

The only places where I've seen single CRs as line endings are in text documents on pre-OSX MacOS systems and in text cells in Excel. Since we are not creating the TOML standard for twenty-year-old Macs, uncoupled CRs created by text editors ought to be a rare occurrence.

When comments were first defined in TOML, there were no restrictions on control codes in comments. That was changed a few years ago when all control codes except horizontal tabs were excluded.

Was this a burden for parser writers to handle when it was introduced? A little, though I'd have to check if they were doing it right. Was it a limitation that users were asking for? It was not, based on what I've read here.

Will existing parsers have to relax these limitations if this PR is merged? Well yes, but they would also be making changes to implement other new features in the future TOML v1.1.0.

Do I have a stake in this? Just in how users copy whole Excel cells into text widgets and the pain I've experienced in stripping loose CRs out of database columns when they are not properly removed or converted. Naive users won't be flummoxed by CRs in comments any longer. This PR doesn't address single CRs showing up anywhere else though.

hukkin:
> The only places where I've seen single CRs as line endings are in text documents on pre-OSX MacOS systems and in text cells in Excel. Since we are not creating the TOML standard for twenty-year-old Macs, uncoupled CRs created by text editors ought to be a rare occurrence.

I agree lone CRs are a rare occurrence. My concern is someone with malicious intent creating a TOML document that looks different to human eye compared to a compliant parser.

ChristianSi:
I think @hukkin's concern is valid *if* editors tend to interpret lone CR's as newlines. What you see should be what you get, so a TOML file that has different linebreaks in a text editor than what a parser sees it not a good idea.

So what do text editors do? I saved @hukkin's CR-only example from above as a text file:

```python
with open('test.txt', 'w') as f:
    f.write('# All line endings in this document are lone CRs\ra = "hello"\rb = "world"\r')
```

and opened it in a few editors, with the following results:

* vim/gvim renders them as `^M`, not linebreak
* libreoffice ignores them altogether
* nano renders them as linebreaks and shows "Converted from Mac format"
* gedit renders them as linebreaks and doesn't show any warning, as far as I could see.

You are all invited to make the same experiment with the editors you commonly use or can think of, and report back.

For me I must say that two our of four is too close for comfort. Gedit's behavior is especially problematic, and since it's the default GUI text editor in Ubuntu (the one that opens when you double-click a text file), I assume it's pretty widely used.

Therefore I return to my earlier position that **lone CRs should remain forbidden in comments**, even if all other control characters are allowed.

eksortso:
@ChristianSi I tried basically the same thing in several Windows 10 text editors. I altered your code slightly, writing the same file in binary mode:

```py
with open('testCR.txt', 'wb') as f:
    f.write(b'# All line endings in this document are lone CRs\ra = "hello"\rb = "world"\r')
```

When I opened them, these were the results:

* Windows 10 Notepad renders them as linebreaks and shows "Macintosh (CR)" as the line ending.
* [Notepad++](https://notepad-plus-plus.org/) does the same thing. The line ending status allows you to modify the line ending to "Windows (CR LF)" and "Unix (LF)" and back to "Macintosh (CR)".
* [VS Code](https://code.visualstudio.com/), out of the box, renders the CRs as linebreaks but incorrectly identifies the line ending as "CRLF". The hex editor does show the 0D codes correctly though. The line ending status will allow you to switch to LF; switching to CRLF does nothing immediately. There's no option to return to CR.
* [Notepad2](https://www.flos-freeware.ch/notepad2.html) (an old reliable workhorse) renders the CRs as linebreaks and identifies the line ending as "CR".
* [micro](https://micro-editor.github.io/) (a new reliable workhorse) incorrectly renders the CRs as whitespace and also misidentifies the line ending as "unix". Among these editors, micro is the newest, so the disregard for CRs can't be unexpected.

I also ran `type testCR.txt` in both `cmd` and PowerShell 7. In `cmd`, typing out the document caused each subsequent line to overwrite the beginning of the previous line, adhering closely to CR's original meaning as a carriage return without a line feed. In PowerShell 7, the CR's were treated as linebreaks.

The mixed behaviors are discomforting. So despite my misgivings expressed previously, I'm changing my mind. Lone CRs ought to be forbidden in comments.

But @hukkin's position is to retain the status quo. Keep all control codes except tab out of comments.

Let me run one more acid test. What happens when the DEL control code, U+007F, is present?

```py
with open('testDEL.txt', 'w') as f:
    f.write("""\
# All lines in this document end with a DEL character.\x7f
a = "hello"\x7f
b = "world"\x7f
""")
```

Notepad and Notepad2 show the DEL character as an unknown character, with a rectangular placeholder. VS Code and Notepad++ show the character as a "DEL" symbol, the former even displaying the symbol in red. micro shows them as whitespace characters. `type testDEL.txt` showed DEL as a pentagon-shaped placeholder when run in both `cmd` and PowerShell 7.

@ChristianSi @abelbraaksma You may want to test how these appear in *nix environment text editors, and how they appear when you `cat` or `bat` them. (I ought to install WSL2 someday.)

The similarities in DEL's appearances, though less glaring, still suggest that allowing other control characters in comments may still be sensible. But I am starting to think that, per @hukkin, keeping the status quo is the best route for us to take. I'm not about to close this PR though; in fact, I will disallow CR characters and replace the "0x"s with "U+00"s in `toml.md` where appropriate.

@abelbraaksma I'm beginning to consider closing this PR, so if you want to continue making your case, now's the time to do it.

As for anyone else with a stake in this, including @pradyunsg @BurntSushi @mojombo, what are your thoughts?

marzer:
I haven't really been following this discourse, but FWIW my position is this:
- having to care about control characters in comments is annoying from a parsing perspective
- lone CR bad

abelbraaksma:
> @abelbraaksma I'm beginning to consider closing this PR, so if you want to continue making your case, now's the time to do it.

It's been a while since I made my original opinion. But from your comments above, I think we have basically two things to consider:

* Line ending normalization
* Any Unicode character allowed in comments

### Thoughts on line endings

I think the first point above belongs to editors. However, if you compare it to some W3C standards, _they are liberal in what they accept, but require normalization_. That is, [they allow `CRLF`, lone `CR`, lone `LF`](https://www.w3.org/TR/REC-xml/#sec-line-ends). Not all programming platforms do this automatically, though (I mean, in general, not specific to XML or anything W3 related). 

However, the algorithm is ridiculously simple:

* All line endings must be normalized before parsing
* `CRLF` is interpreted as single `LF`
* any `CR` not followed by an `LF` is interpreted as a single `LF`

Henceforth, line endings are then normalized, and what we call a `new line` or `carriage return` in Unicode is interpreted as such and _always_ start a new line.

In practice, this would mean removing the following from the ABNF:

```
newline =  %x0A     ; LF
newline =/ %x0D.0A  ; CRLF
```

and making newlines implicit (i.e., part of the prose, not the ABNF). Or, conversely, since _alternatives_ in parsing are ordered, this could work, I think (which, btw, already suggests that the current code is actually incorrect?):

```
newline =/ %x0D.0A  ; CRLF
newline =/ %x0D     ; CR
newline =  %x0A     ; LF
```


### Thoughts on control characters

Currently, it is clear to me that comments are overly restrictive (i.e., control characters are absent), but also too lenient (illegal unicode characters are allowed). The range, as it currently stands is:

```
comment-start-symbol = %x23 ; #
non-ascii = %x80-D7FF / %xE000-10FFFF
non-eol = %x09 / %x20-7E / non-ascii

comment = comment-start-symbol *non-eol
```

If we expand this, we get:

```
allowed-comment-char =%x09 / %x20-7E / %x80-D7FF / %xE000-10FFFF
```

Let's break this down:

1. Control characters (the current discussion) are disallowed, but my point in the original issue was: "encounder `#`, then read to EOL" is easier to understand, and at least parsers won't need to raise to a user something like: "hey, you wrote a comment, and we know you just want to ignore whatever is in there, but we cannot do that, there are invisible characters that we don't like, please remove them".

2. `%7F` is the `DEL` character. This is just a valid Unicode character, albeit without a glyph, but that's true for so many. I'd just allow it.

3. `%D800 - %xDFFF`, the so-called "surrogate block". These have special meaning, _but only in UTF-16_, when encountered as literal bytes in the stream. However, they themselves can be encoded just fine, both in UTF-16 and UTF-8, so there is no specific need to disallow them.

4. `%FFFE` and `%FFFF`. From all the _noncharacters_ in Unicode, these are special, as they are used as BOM (byte order mark). It is certainly discouraged to use these in any text document. Even in UTF-8, it often appears in the start to signal that the document should be interpreted as UTF-8 and not ASCII or Latin-1 or something. But, if appearing at the start of a document, they may be special, there is nothing that disallows UTF-8 to encode this, even though the characters themselves have no meaning (unless software assigns special meaning). This used to be different, but has changed long ago, see section 9.3 in these minutes: https://www.unicode.org/wg2/docs/n2403.pdf.

5. Other _noncharacters_: basically anything that contains `FFFE` or `FFFF` is considered a non-character in Unicode. I.e. `%2FFFE` and `%10FFFE` are non-characters. Again, there use is discouraged, because both UTF-16 and UTF-32 use these for endianness detection (i.e., `FFFE` means big endian and `FEFF` means little endian). But again, in UTF-8, there's nothing wrong in encoding these, it's just _bad practice_ to use them.

6. `%FFFD` the Replacement Character. This one is currently allowed, but problematic in many ways. A lot of editors, when encountering invalid encodings (not to be confused with invalid code points), replace these with this character. It is used for cases where the parser basically says "I don't know what you mean". Since it's already allowed, it should stay that way, I guess.

7. `%00`. The `NUL` character is allowed in Unicode, but many standards disallow it explicitly. The reason is simple: a lot of programming languages represent a string in memory using a `NUL` delimiter, like C/C++. While this may not be the case for many other languages, there have been recorded cases where DOS and buffer overflow attacks where successful using `NUL` in Unicode strings. TLDR: I'd vote for certainly still disallowing this character anywhere in the document.

8. `%85`, `%2028` and `%2029` (see https://en.wikipedia.org/wiki/Newline#Unicode). Different standards give these different meanings. `%85` is `NEL` and that code point is special as it maps to some old EBCDIC newline character. The others were "invented" for Unicode and I've rarely seen them in practice. Since we currently don't consider these newlines, I don't think we can change that (or should even consider).

### Conclusion?

Sorry for the long post (it didn't set out that way!). Basically, we have this situation:

* We already allow characters that are strongly discouraged to appear in Unicode
* We disallow characters that Unicode has absolutely no problem with

My vote: let's take the union of these two and simplify parsing. A comment can just contain any Unicode character, with just two exceptions: line endings, and `NUL`. I.e., it would become this:

```
allowed-comment-char =%x01-%09 / %0B-%0C / %0E-10FFFF
```

Meanwhile, I do think we should correct the newline handling.

I realize this may be a little too lenient for some, but I see no harm in doing so. Any existing correct TOML file out there still parses just fine, and once this is adopted, parsers won't have to treat comments any different than "ignore from `#` to EOL`.

ChristianSi:
Hmm, I don't think we should allow lone CR's as linebreaks – most specifically, since we had discussed that in the past and decided against it. I'm not sure where that was, but anyone willing to search should find it. And just reversing one's position every few years without really new arguments having come up is not a good thing.

More specifically, there are currently no TOML fines that use CR's as linebreaks, since this linebreak convention isn't used anywhere anymore, except maybe on some computers that have long ago ended up in museums, there is just no reason why they should ever come into existence.

As for @abelbraaksma's suggestions on allowed-comment-char, the first variant seems otherwise fine to me. So I would keep FF and VT forbidden, and add CR to the forbidden characters too.

abelbraaksma:
@ChristianSi thanis for clarifying. I’m not aware of the old decisions, I wholeheartedly agree to staying consistent. 

I have seen such file in the wild, usually as a result of bad switching between platforms or editors mistreating LFs (try opening such files in Notepad, then manually adding newlines on top of those).

But bad editing is no reason for allowing such things, I agree. And the times that DNX had LFCR (as opposed to CRLF) is indeed far in the past.

eksortso:
I'll use @abelbraaksma's first `allowed-comment-char` variant. So we will be disallowing NUL, CR, FF, VT, and LF in comments (though the last one won't generate an error because it would simply end the line).

The matter of normalizing line endings (just LF and CRLF in TOML's case) is appealing to me, especially since we're going to someday form an RFC based on the TOML spec, per #870. But seeing as text editors can unintentionally mix line endings, I won't push it.

eksortso:
A few more notes. I do not like surrogate code points being encoded in UTF-8, because it is in fact _illegal_ to have such byte streams in UTF-8. And in fact, we banned surrogates from TOML documents in #619 three years ago with the introduction of `non-ascii` in the ABNF.

But to simplify comment scanning while disallowing the historic line-breaking ASCII codes (_**please help:** what's a better name for the collected set of codes CR, FF, VT, and LF?_), I'll use the new `allowed-comment-char = %x01-%09 / %0E-10FFFF`, and relocate `non-ascii` elsewhere closer to where it is first used.

Expect a little rearrangement.

toml.md
```diff
@@ -38,23 +38,21 @@ should be easy to parse into data structures in a wide variety of languages.
 
 - TOML is case-sensitive.
 - A TOML file must be a valid UTF-8 encoded Unicode document.
-- Whitespace means tab (0x09) or space (0x20).
-- Newline means LF (0x0A) or CRLF (0x0D 0x0A).
+- Whitespace means tab (U+0009) or space (U+0020).
+- Newline means LF (U+000A) or CRLF (U+000D U+0x0A).
```
hukkin, line 42:
```suggestion
- Newline means LF (U+000A) or CRLF (U+000D U+000A).
```
A very minor error here! :)

(IMO best would be having these unrelated cosmetic changes in a separate PR)

eksortso: Since it's germane to what we're discussing here, let's roll it into this PR.

abelbraaksma:
I wrote this about _surrogate codepoints_:

> However, they themselves can be encoded just fine, both in UTF-16 and UTF-8, so there is no specific need to disallow them.

This isn't (entirely) correct, apologies. I merely checked Wikipedia and a few other sources, but I should've checked the actual source, the Unicode standard. The short summary of a long reading of specs:

* In the past, surrogate characters where encoded as two pairs of three UTF-8 code units (bytes). There is still software out there that does this, but it is considered _invalid_. A historical Unicode encoding, CESU-8, supports this, though. UTF-8 readers encountering this should either signal with an error, or display the `U+FFFD` (replacement char, or `�`).
* Proper encoding of single surrogate characters (not being in pairs) _does not exist_ (unless historically, CESU-8), and when encountered, should be replaced, again, by `U+FFFD`.
* Proper encoding of proper pairs of surrogates, encountered as transcoding from a UTF-16 stream, _must be encoded as a single 4-unit (byte) UTF-8 stream_. It is illegal to do otherwise (illegal from a standards perspective, I doubt you'll be persecuted).
* Encoding of surrogate code-points _as themselves_ in UTF-16 is not possible. They have special meaning.
* Lone surrogate code code-points in UTF-16 are, again, illegal. They can either be signaled, or replaced by `�`.

### What does this mean for us?

Not much, really. TOML currently _requires_, encoding as UTF-8. What this means is, that _it is impossible to encounter surrogate characters_.

If you do encounter them, you'll encounter them _prior_ to reading them as UTF-8. By the time your stream turns from a byte stream into a proper UTF-8 stream, those (illegal) surrogate pairs are no longer there, they _must_ be replaced by `�`, or, alternatively, have raised an error: "Illegal file encoding detected", or something like that.

### So, what does this _really_ mean for us?

We should not treat surrogate characters as special. They don't exist. The only place where we should disallow them, is in escape sequences. I.e., you should not be allowed to write `\D800` in a string.

If we still have remnants of ABNF dealing with surrogates, we should really (probably?) remove them. Or we should remove the requirement for TOML being valid UTF-8, in which case it makes sense to embed the Unicode rules in the spec.

----
For reference, this FAQ summarizes the above as well: [UTF-8 FAQ](https://unicode.org/faq/utf_bom.html#UTF8)

abelbraaksma:
The above really is an in-depth answer to:

> I do not like surrogate code points being encoded in UTF-8, because it is in fact illegal to have such byte streams in UTF-8. And in fact, we banned surrogates from TOML documents in https://github.com/toml-lang/toml/pull/619 three years ago with the introduction of non-ascii in the ABNF.

The issue you refer to makes the wrong assumption. It conflates "UTF-16 surrogate code-points" and "UTF-8 encoding". Futhermore, UTF-16 cannot represent these code-points either, but as _code unit_ in UTF-16 they have meaning. These code-points are deemed non-representable. [They don't even have a table in Unicode](https://www.unicode.org/charts/PDF/UD800.pdf). Quote:

> Isolated surrogate code points have no interpretation; consequently, no character code charts or names lists are provided for
this range

And:

> For a complete understanding of high-surrogate code units low-surrogate code units, and surrogate pairs used for the
UTF-16 encoding form, see the appropriate sections of the Unicode Standard, online a

Note that this second comment talks about _code units_, not code-points. The latter "have no interpretation", the former do.

We should be rather careful about our wording when it comes to this. And perhaps we should even add a Note to the ABNF (it's quite common to have this type of notes in specs), explaining that any TOML byte stream is to be interpreted as a _valid UTF-8 byte stream_ and that, as such, only valid code units should be considered, and invalid code units should be, as per the Unicode spec, be considered as `U+FFFD`.

This includes the many other, potentially invalid encodings one may encounter in a UTF-8 document. In fact, we should just defer to the Unicode spec itself here.

abelbraaksma:
I could have been briefer in my response. In #619 we did not ban surrogates, they were never allowed to begin with. We merely added a redundancy in the spec, a rule about a condition that didn't exist ;).

abelbraaksma:
> Though I don't think "valid Unicode code point" is a well-defined term? So maybe it's better to talk about "Unicode Scalar Values", which is defined in the Unicode spec?

Yes, terminology is tricky. The surrogate code-points [are certainly _valid code-points_](https://unicode.org/glossary/#code_point), after all, they are defined. However, they have _no representation_, so they cannot and must not exist in a UTF-8 byte stream (as explained at lengths above).

Another way of reading "valid" in this context is that, because they cannot be represented, that they are _invalid_, but I feel that that's requiring a little too much interpretation on the reader.

Note that different libraries use the word "code-point" differently and it depends on contexts. For instance in .NET, when normalizing, it [can throw an error "invalid code-point"](https://stackoverflow.com/a/27084836/111575). However, this is contextual: in the realm of normalizing a string, unassigned code-points may be treated as invalid. The UTF-8 encoding would only throw an exception on encountering invalid byte sequences (which includes the surrogate pairs).

Also invalid could, for instance, [_variation sequences_ be](https://unicode.org/glossary/#variation_sequence). But we should not concern ourselves with that. Even with invalid variations, the string would still be a _valid UTF-8_ byte sequence.

We could link to the term [_Well-formed code-unit sequence_ from the spec](https://unicode.org/glossary/#well_formed_code_unit_sequence), if we want to be precise.

For fun, Unicode itself is quite precise. So much so, that they use the [following definition for _Surrogate character_](https://unicode.org/glossary/#surrogate_character):

> Surrogate character: A misnomer. It would be an encoded character having a surrogate code point, which is impossible. Do not use this term.

Anyway, TLDR etc etc, what about this?

> Specifically this means that, should a file contain byte sequences that cannot be decoded as [valid Unicode Coded Character Sequence](https://unicode.org/glossary/#coded_character_sequence) (such as code units attempting to represent [Surrogate code points](https://unicode.org/glossary/#surrogate_code_point)) the file must be rejected or these byte sequences must be replaced with `�` (U+FFFD), as per the Unicode spec.

abelbraaksma:
Also for fun, in case anyone thought `CR`, or `LS` was entirely dead, this is what Visual Studio offers when a file has inconsistent line-endings (and yes, this was a TOML file that got warbled from going back and forth between Windows and Mac machines).
image

And no, I'm not suggesting we should suddenly start supporting `LS` or `PS`, which would be rather exotic.

eksortso:
It took me a long time to compose a response to recent messages. And by the time I was able to post my response, another handful of responses came through. So I'm breaking my response up into smaller chunks.

@ChristianSi wrote:
> @abelbraaksma: Thanks for the detailed explanations!

Thanks as well @abelbraaksma! That was a very deep dive into Unicode terminology and usage. Somewhat distracting, but enlightening. We'll have to take what we learned from this and express it as laconically as we can within the spec to our less technically-oriented audience. (I don't do laconic well, but I try!)

I'm using the [Unicode Glossary](https://unicode.org/glossary/#unicode_scalar_value) for my interpretations of the terminology that we're discussing. I got [Joel Spolsky](https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/) to thank for my early Unicode clarity, and I've forgotten most of it, it seems!

@abelbraaksma wrote:
> > And perhaps we should even add a Note to the ABNF (it's quite common to have this type of notes in specs), explaining that any TOML byte stream is to be interpreted as a valid UTF-8 byte stream and that, as such, only valid code units should be considered, and invalid code units should be, as per the Unicode spec, be considered as U+FFFD, or raise an exception.

The ABNF already contains the following notice. Its use of "Unicode Code Point" is still acceptable; all the code points encountered after successful parsing would be scalar values.

```abnf
;; Although a TOML document must be valid UTF-8, this grammar refers to the
;; Unicode Code Points you get after you decode the UTF-8 input.
```

But it doesn't say what to do with invalid code points or incorrect byte streams. So let's add a version of @ChristianSi's response to `toml.md`. I have something written up. But let me review the last few responses first.

eksortso:
From our perspective, injecting U+FFFD to replace bad byte sequences or code points is not a good idea from our perspective. For many years, we have implicitly held, with [notable exceptions](https://github.com/toml-lang/toml/issues/538#issuecomment-397857741), that the proper response to bad TOML is not warnings or substitutions but _errors,_ and as this has taken root, we think that the more helpful and descriptive the error messages are, the better. That ought to be our standard, and we should say so in no uncertain terms.

The spec routinely refers to errors being _produced_ in exceptional situations, but we see the word _throw_ written in a few places. It may be worth making our own terminology consistent, going forward.

eksortso:
We are drifting far away from the original purpose of this PR. Let's get this done, and get back on track.

I took @ChristianSi's recommendation into account and revised the **Spec** section of `toml.md`. I moved the case-sensitive condition down and made the expected format of the document its own paragraph. It also emphasizes that errors must be produced if the document isn't valid UTF-8.

Here is the proposal. What do you think?

---

## Spec

A TOML file must be a valid UTF-8 encoded Unicode document. If a TOML parser
encounters an invalid byte sequence, or any code point not valid in UTF-8 (such
as a surrogate), then it must reject the document and produce an error.

- TOML is case-sensitive.
- Whitespace means tab (U+0009) or space (U+0020).
- Newline means LF (U+000A) or CRLF (U+000D U+000A).

abelbraaksma:
> From our perspective, injecting U+FFFD to replace bad byte sequences or code points is not a good idea. For many years, we have held implicitly with https://github.com/toml-lang/toml/issues/538#issuecomment-397857741, that the proper response to bad TOML is not warnings or substitutions but errors, [...]

Yes, but we are not talking about TOML here. We are talking about UTF-8 encoding. Even if you tried, you could not prevent this behavior if your language of choice just reads a file as a string or UTF-8 stream.

It's the _standard before the standard_. Just like we don't define how TOML is saved on disk or in memory, we don't define how bad UTF-8 is treated, as that is already part of UTF-8 itself.

Not even XML, HTML or any other standard depending on UTF-8 says how the encoding/decoding process should take place. We just say what encoding TOML should be in.

We should just say: it should be valid UTF-8. Period. I tried to capture that in my last paragraph (the suggested wording there). The explicit mention there on surrogate pairs _is only in reference to the Unicode specification on UTF-8_. It doesn't change how, for about two and a half decades, UTF-8 encoding/decoding has been done (well, for conforming implementations, that is).

Try to load invalid UTF-8 in (almost) any editor. You'll get the U+FFFD character. We can try to force different behavior, but I think we should just go with the flow here and stick with accepted practices.

Also: talking about clarity of errors and the like: I don't think an encoding error is going to be clear for anybody. I.e., he or she would open the file in an editor and there's no error anymore. If it comes to clarity, I think just sticking to what Unicode itself says, and referring to that (and possibly repeating it) is really the best we can do.

abelbraaksma:
PS: in the end I'm fine with whatever is going to be decided here. Don't take my point of view as gospel (I'm sure you won't, but just wanted to emphasize that). I tend to be a bit picky when it comes to standards, it's what you get when you spend too long in standardization committees. That's not a recommendation! 😆

eksortso:
If we state that TOML should be (not _must_ be) valid UTF-8, any problems that parsers may face from trying to mess with invalid file formats will be up to their maintainers to correct. But if we said "must," then the ABNF (and its assumption of successfully decoded text) would make a lot more sense. So I will keep the "must" in the valid UTF-8 requirement. I'll remove the rest. But I'll keep that line first.

> Try to load invalid UTF-8 in (almost) any editor. You'll get the U+FFFD character. We can try to force different behavior, but I think we should just go with the flow here and stick with accepted practices.

Editors have their own behaviors when handling bad byte sequences. But if you had a parser that loaded a byte stream, it may not immediately decode it before starting to parse, and try to push their own behaviors if they encounter bad inputs. But on the other hand, I may be underestimating the processing power of modern embedded systems.

> Also: talking about clarity of errors and the like: I don't think an encoding error is going to be clear for anybody. I.e., he or she would open the file in an editor and there's no error anymore. If it comes to clarity, I think just sticking to what Unicode itself says, and referring to that (and possibly repeating it) is really the best we can do.

True.

abelbraaksma:
> If we state that TOML should be (not must be) valid UTF-8

While not really OT, a typical, other way of phrasing this, is saying that the spec is based on valid Unicode and that at a minimum, UTF8 must be supported to be a compliant parser. Whether parsers allow a wider range of encodings is really up to the implementers and “outside the scope f this specification”. This already happens in practice anyway, and is a good thing.

You can still say that a valid, interchangeable file must be encoded in valid UTF-8.

But that’s probably something for another thread and discussion :).

eksortso:
Here are my notes on the most recent commit.

In `toml.abnf`, I moved `non-ascii` back into the definition of `allowed-comment-chars`. This way, we avoid mentioning the surrogate code point range. We make that doubly clear.

In `toml.md`, I shifted up the UTF-8 document requirement up to its own line outside the bullet list in the Spec section. And I reworked the Comment description. Instead of naming the excluded code points, I describe why they're excluded and just list them.

Does the following language make sense?

> A hash symbol marks the rest of the line as a comment, except when inside a
> string. Comments may contain any valid code points except a limited subset of
> ASCII control codes that could cause problems during editing or processing
> (specifically, U+0000, and U+000A to U+000D).

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## unreleased
 
+- Relax comment parsing; control characters are again permitted.
```
ChristianSi, line 5: Change to: ...; most control characters are again permitted.

toml.abnf
```diff
@@ -37,11 +37,10 @@ newline =/ %x0D.0A  ; CRLF
 
 ;; Comment
 
+comment = comment-start-symbol *allowed-comment-char
```
ChristianSi, line 40: This new rule most be moved two lines down, after `allowed-comment-char`.

toml.md
```diff
@@ -36,25 +36,25 @@ should be easy to parse into data structures in a wide variety of languages.
 
 ## Spec
 
+A TOML file must be a valid UTF-8 encoded Unicode document.
```
ChristianSi, line 39: I still think it would be good to add a more specified second sentence here, similar to the excellent wording [proposed](https://github.com/toml-lang/toml/pull/924#issuecomment-1284733670) by @abelbraaksma. However, if everybody else considers this unnecessary, I will not insist.

toml.md
```diff
@@ -36,25 +36,25 @@ should be easy to parse into data structures in a wide variety of languages.
...
-- Whitespace means tab (0x09) or space (0x20).
-- Newline means LF (0x0A) or CRLF (0x0D 0x0A).
+- Whitespace means tab (U+0009) or space (U+0020).
+- Newline means LF (U+000A) or CRLF (U+000D U+000A).
 
 ## Comment
 
 A hash symbol marks the rest of the line as a comment, except when inside a
-string.
+string. Comments may contain any valid code points except a limited subset of
```
ChristianSi, line 48: I would change "valid code points" to "characters". Since we already state that that the whole document must be valid UTF-8, "valid" seems an unnecessary duplication. Any elsewhere in the document we consistently talk about "characters" rather than "code points", so it seem a bit strange to deviate here from the usual terminology.

ChristianSi (changes requested):

ChristianSi:
Another little clean-up proposal while we're at it: the spec talks several times about "Unicode character(s)", but uses the alternative "UTF-8 character(s)" just once. I would propose to change the latter to "Unicode character" as well, as the specific mention of UTF-8 seems rather pointless here.

toml.md
abelbraaksma: There’s, though, one problem with “character”. A single character can consist of multiple code points, and Unicode says nothing about what a valid character is. You can, for instance, make your own by using several combining diacritics.

Furthermore, while we require valid UTF-8 byte streams, it’s possible to use invalid combinations of code points, like a smiley variant selector in the wrong position. Or a BOM in the middle of the file.

toml.md
abelbraaksma: Would it make sense to just add a link to the spec, ideally where it describes UTF-8? Or the FAQ link where it explains what it means to be “valid”? But yeah, I concur with whatever is decided here, though adding the informational sentence wouldn’t hurt, I guess.

toml.md
ChristianSi: OK, let's stick with "code points" then. But is there any good reason for the "valid"?

toml.md
ChristianSi: So what about?

> Specifically this means that, should a file as a whole not form a [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence), the file must be rejected (preferably) or illegal byte sequences must be replaced with `�` (U+FFFD), as per the Unicode spec.

This is your proposal, but shortened somewhat and with a "preferably" as a compromise between the TOML spec authors' general preference for strict error handling and the recognition that a "replacement" strategy is equally Unicode-compatible and that implementers may not always have a convenient choice this regard, depending on which libraries they rely on for Unicode handling.

toml.md
abelbraaksma: I think it is in reference to the fact that, as I showed above, you _could_ encode invalid code points (i.e., code points above `0x10FFFF` for instance), or you _could_ consider encoding surrogate code points, but these aren't _invalid code points_, they are just _invalid code units_ in a UTF-8 byte stream.

Terminology is tricky. By the time you have _valid UTF-8_, you cannot have invalid code points (unless "invalid" means in the context of this specification). Perhaps this would be better?

> Comments may contain any Unicode code points except a limited subset of

toml.md
abelbraaksma: @ChristianSi  Sounds good to me, I like it!

abelbraaksma: I think having this line in, removes the requirement elsewhere to try to define "valid", and we can just drop it, and use the word "Unicode" instead, or leave it out completely.

CHANGELOG.md
eksortso: Will do. You'll see it in a bit.

eksortso:
> Another little clean-up proposal while we're at it: the spec talks several times about "Unicode character(s)", but uses the alternative "UTF-8 character(s)" just once. I would propose to change the latter to "Unicode character" as well, as the specific mention of UTF-8 seems rather pointless here.

I replaced "UTF-8 character" with "Unicode character". And in the literal string section, I replaced the original phrase with "Thus, for binary data, it is recommended that you use Base64 or another suitable _text_ encoding" (my emphasis). The word "text" replaced "ASCII or UTF-8". The context is binary-to-text encoding, so this is appropriate.

toml.md
eksortso: For as long as I've worked with Unicode, I've wondered what an intuitive replacement for the word "character" that wouldn't sound too alien to non-technical people could be. "Code point" is a bit formal, but to the extent that people need to be knowledgeable about these things, they need to know that code points are more fundamental than characters.

That said, I've used the final sentence above. Thanks for your help and insight @ChristianSi @abelbraaksma !

toml.md
eksortso: I added this text (without the literal replacement character) in the spec section.

> A TOML file must be a valid UTF-8 encoded Unicode document. Specifically this
> means that, should a file as a whole not form a well-formed code-unit sequence,
> the file must be rejected (preferably) or ill-formed byte sequences must be
> replaced with U+FFFD as per the Unicode spec.

I also removed the word "valid" in two places in the spec that referred to "valid" Unicode characters or scalar values.

eksortso:
Dang, that bot is good. I had `[scalar values]` split across separate lines, and the bot meshed those lines together. I didn't think there was a difference.

Anyway, what do you think of the current state of the PR?

abelbraaksma (approved):
LGTM, great work.

We might (emph on ‘might’) specify the range of scalar values in an informational sentence, similarly how you did that above. Something like:

> Specifically this means an escape sequence can represent any Unicode code point, assigned or not, with the exception of code points in the surrogate block.

But since we link to the definition, this may be a little TMI.

toml.md
```diff
@@ -36,25 +36,28 @@ should be easy to parse into data structures in a wide variety of languages.
 
 ## Spec
 
+A TOML file must be a valid UTF-8 encoded Unicode document. Specifically this
+means that, should a file as a whole not form a well-formed code-unit sequence,
```
ChristianSi, line 40: Why not add a link to https://unicode.org/glossary/#well_formed_code_unit_sequence here? We also link to "Unicode scalar value" elsewhere in the document, and considering that this is a fairly technical term, it would really be helpful.

toml.md
```diff
@@ -417,8 +420,8 @@ str = ''''That,' she said, 'is still pointless.''''
 ```
 
 Control characters other than tab are not permitted in a literal string. Thus,
+for binary data, it is recommended that you use Base64 or another suitable text
+encoding. The handling of that encoding will be application-specific.
```
ChristianSi, line 424: I would change "text encoding" to "binary-to-text encoding", as Wikipedia [has it](https://en.wikipedia.org/wiki/Binary-to-text_encoding).

ChristianSi (changes requested):

ChristianSi:
@eksortso:

> Dang, that bot is good. I had `[scalar values]` split across separate lines, and the bot meshed those lines together. I didn't think there was a difference.

There shouldn't be a difference, and I have no idea why the bot did that. Markdown allows links to include linebreaks. But if the bot prefers to keep them on a single line, sure why not...

ChristianSi:
@abelbraaksma:

> We might (emph on ‘might’) specify the range of scalar values in an informational sentence, similarly how you did that above.

I don't think that's necessary, considering that the definition is just one mouse click away. But I wouldn't be opposed either.

toml.md
eksortso: Good idea.

eksortso: On second thought, the link to the Unicode glossary could use a suitable replacement. A casual user might follow the link, get run around on the different definitions that they would need to understand, and give up. I'll use that link. But something that directly spells out what constitutes a valid document in Unicode and in UTF-8 in particular would be more helpful than that technical definition.

toml.md
eksortso: Alright.

toml.md
abelbraaksma: Such a link, that points directly to the pdf as a deeplink to the correct section, is above in my comments. Maybe use that?

abelbraaksma: This one: https://www.unicode.org/versions/Unicode15.0.0/ch03.pdf#G7404

eksortso: I'm sorry @abelbraaksma, but jumping into a huge PDF document would make a casual user quit all the faster. The glossary is laconic at least. So we'll stick with the link to that, and anyone who wants to dive into that rabbit hole can do so and find the Unicode specs they need.

Modern text editors know what a valid UTF-8 document should look like, after all.

eksortso:
Thanks @ChristianSi @abelbraaksma for the approvals, and with all of your help!

@pradyunsg @mojombo Can you have a look at this now? Is it good to merge?

pradyunsg:
There's quite a few more changes here than the PR title suggested; I'll try and take a look in the coming weeks.

toml.md
abelbraaksma: I merely responded to:

>  But something that directly spells out what constitutes a valid document in Unicode and in UTF-8 in particular would be more helpful than that technical definition.

Be careful what you wish for! (though, tbh, a link from spec to spec shouldn’t scare somebody who’s capable of interpreting ABNF. Then again, 99 out of 100 will just use a library that does the UTF8 reading, which gives all this for free)

eksortso:
@pradyunsg Could we split the changes into two separate PRs? The Unicode language changed a lot. Suppose we could isolate those changes and address them separately?

toml.md
eksortso: > I merely responded to:
> 
> >  But something that directly spells out what constitutes a valid document in Unicode and in UTF-8 in particular would be more helpful than that technical definition.
> 
> Be careful what you wish for!

In retrospect, I was asking for two diametrically opposed texts. One that was short, sweet, directed at casual readers, and already well accepted by UTF-8 encoder libraries; that's what I actually wanted. The other was the entirety of the Unicode spec detailing UTF-8. Which you directed me to.

@abelbraaksma Sorry for the confusion. Part of me wants to keep the spec as direct as possible. But those days are gone. Frankly, [toml.io](https://toml.io)'s home page is as direct as we'll ever get.

eksortso:
I stripped this PR back down to its essentials, that is, the re-allowance of most control characters inside comments. Anything pertaining to Unicode language changes was already moved over to #929, which can be reviewed separately.

@ChristianSi @abelbraaksma Want to take another look?

@pradyunsg Can you review this PR again?

eksortso:
@pradyunsg This is a lot simpler since the last time you looked at it. We're allowing all but five (UTF-8-valid) code points in comments. Is this good to merge?

eksortso:
@pradyunsg Could you please review this and make a decision? It's been two months since the last reminder.

toml.md
```diff
@@ -44,17 +44,16 @@ should be easy to parse into data structures in a wide variety of languages.
...
-string.
+string. Comments may contain any Unicode code points except the following
+control codes that could cause problems during editing or processing: U+0000,
+and U+000A to U+000D.
 
 ```toml
 # This is a full-line comment
 key = "value"  # This is a comment at the end of a line
 another = "# This is not a comment"
 ```
```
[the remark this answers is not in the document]
eksortso: @pradyunsg Sure thing. I didn't know this was a style guideline, but I know now.

(I didn't know I put that in the commit message. May squash merge forgive me!)

pradyunsg:
Thanks @eksortso! ^.^

abelbraaksma:
Well done, great work!

Issue #165, opened 2013-03-03
Unescaped invisible bytes


uiri:
Is it permissible to stick a literal null byte inside a string? Or must a null byte be expressed as either \0 or \x00 ? If not a null byte, what about a null byte with the least significant digit flipped (literal 0x01)? Is that permissible inside a string or must it be expressed as \x01? It is rather confusing when running something like:

```
ip = "\x0a\x00\x00\x02"
```

through a parser and back out turns into

```
ip = "\n\0\0"
```

With an invisible 0x02 at the end of the string. Is the parser not supposed to turn the escape sequence into the literal byte? Is the emitter (for lack of a better word) supposed to turn any byte which isn't printable ascii into a byte escaped with the \xXX sequence? Strings are supposed to be UTF-8 so I assume throwing in unicode characters is fine but some non-printable ASCII are designated as special escaped characters which muddies the waters a bit.

PR #929, opened 2022-10-27
Clarify Unicode and UTF-8 references

eksortso:
This PR applies the changes to clarify Unicode and UTF-8 references, done originally for #924 while considering relaxations on control code in comments. These are standalone changes, separate from the original scope of #924. Many thanks to @abelbraaksma and @ChristianSi for their work on this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## unreleased
 
+- Clarify Unicode and UTF-8 references.
 - Allow newline after key/values in inline tables.
 - Allow trailing comma in inline tables.
 - Clarify where and how dotted keys define tables.
--- a/toml.md
+++ b/toml.md
@@ -36,10 +36,15 @@ should be easy to parse into data structures in a wide variety of languages.
 
 ## Spec
 
+A TOML file must be a valid UTF-8 encoded Unicode document. Specifically this
+means that, should a file as a whole not form a
+[well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence),
+the file must be rejected (preferably) or ill-formed byte sequences must be
+replaced with U+FFFD as per the Unicode spec.
+
 - TOML is case-sensitive.
-- A TOML file must be a valid UTF-8 encoded Unicode document.
-- Whitespace means tab (0x09) or space (0x20).
-- Newline means LF (0x0A) or CRLF (0x0D 0x0A).
+- Whitespace means tab (U+0009) or space (U+0020).
+- Newline means LF (U+000A) or CRLF (U+000D U+000A).
 
 ## Comment
 
@@ -265,7 +270,7 @@ The above TOML maps to the following JSON.
 ## String
 
 There are four ways to express strings: basic, multi-line basic, literal, and
-multi-line literal. All strings must contain only valid UTF-8 characters.
+multi-line literal. All strings must contain only Unicode characters.
 
 **Basic strings** are surrounded by quotation marks (`"`). Any Unicode character
 may be used except those that must be escaped: quotation mark, backslash, and
@@ -293,7 +298,7 @@ For convenience, some popular characters have a compact escape sequence.
 ```
 
 Any Unicode character may be escaped with the `\xHH`, `\uHHHH`, or `\UHHHHHHHH`
-forms. The escape codes must be valid Unicode
+forms. The escape codes must be Unicode
 [scalar values](https://unicode.org/glossary/#unicode_scalar_value).
 
 All other escape sequences not listed above are reserved; if they are used, TOML
@@ -417,8 +422,9 @@ str = ''''That,' she said, 'is still pointless.''''
 ```
 
 Control characters other than tab are not permitted in a literal string. Thus,
-for binary data, it is recommended that you use Base64 or another suitable ASCII
-or UTF-8 encoding. The handling of that encoding will be application-specific.
+for binary data, it is recommended that you use Base64 or another suitable
+binary-to-text encoding. The handling of that encoding will be
+application-specific.
 
 ## Integer
 
--- a/toml.md
+++ b/toml.md
@@ -299,10 +299,12 @@ For convenience, some popular characters have a compact escape sequence.
 
 Any Unicode character may be escaped with the `\xHH`, `\uHHHH`, or `\UHHHHHHHH`
 forms. The escape codes must be Unicode
-[scalar values](https://unicode.org/glossary/#unicode_scalar_value). Note that
-there is no low-level byte sequence syntax in TOML, and that these escape codes
-must not be made to work in such a way. So for binary data, it is recommended
-that you use Base64 or another binary-to-text encoding.
+[scalar values](https://unicode.org/glossary/#unicode_scalar_value).
+
+Note that there is no low-level byte sequence syntax in TOML, and that the
+character escape codes must not be made to work in such a way. So for binary
+data, it is recommended that you use Base64 or another binary-to-text encoding,
+and that external tools ought to be used to interpret such encodings.
 
 All other escape sequences not listed above are reserved; if they are used, TOML
 should produce an error.
--- a/toml.md
+++ b/toml.md
@@ -301,10 +301,11 @@ Any Unicode character may be escaped with the `\xHH`, `\uHHHH`, or `\UHHHHHHHH`
 forms. The escape codes must be Unicode
 [scalar values](https://unicode.org/glossary/#unicode_scalar_value).
 
-Note that there is no low-level byte sequence syntax in TOML, and that the
-character escape codes must not be made to work in such a way. So for binary
-data, it is recommended that you use Base64 or another binary-to-text encoding,
-and that external tools ought to be used to interpret such encodings.
+Keep in mind that all TOML strings are sequences of Unicode characters, _not_
+byte sequences. For binary data, avoid using these escape codes. Instead,
+external binary-to-text encoding strategies, like hexadecimal sequences or
+[Base64](https://www.base64decode.org/), are recommended for converting between
+bytes and strings.
 
 All other escape sequences not listed above are reserved; if they are used, TOML
 should produce an error.
--- a/toml.md
+++ b/toml.md
@@ -36,15 +36,15 @@ should be easy to parse into data structures in a wide variety of languages.
 
 ## Preliminaries
 
-A TOML file must be a valid UTF-8 encoded Unicode document. Specifically this
-means that, should a file as a whole not form a
-[well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence),
-the file must be rejected (preferably) or ill-formed byte sequences must be
-replaced with U+FFFD as per the Unicode specification.
-
 - TOML is case-sensitive.
 - Whitespace means tab (U+0009) or space (U+0020).
 - Newline means LF (U+000A) or CRLF (U+000D U+000A).
+- A TOML file must be a valid UTF-8 encoded Unicode document.
+
+  Specifically this means that a file _as a whole_ must form a
+  [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence)
+  and will be rejected (preferably) or have ill-formed byte sequences replaced
+  with U+FFFD as per the Unicode specification.
 
 ## Comment
 
--- a/toml.md
+++ b/toml.md
@@ -14,7 +14,7 @@ should be easy to parse into data structures in a wide variety of languages.
 
 ## Table of contents
 
-- [Spec](#user-content-spec)
+- [Preliminaries](#user-content-preliminaries)
 - [Comment](#user-content-comment)
 - [Key/Value Pair](#user-content-keyvalue-pair)
 - [Keys](#user-content-keys)
@@ -34,13 +34,13 @@ should be easy to parse into data structures in a wide variety of languages.
 - [MIME Type](#user-content-mime-type)
 - [ABNF Grammar](#user-content-abnf-grammar)
 
-## Spec
+## Preliminaries
 
 A TOML file must be a valid UTF-8 encoded Unicode document. Specifically this
 means that, should a file as a whole not form a
 [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence),
 the file must be rejected (preferably) or ill-formed byte sequences must be
-replaced with U+FFFD as per the Unicode spec.
+replaced with U+FFFD as per the Unicode specification.
 
 - TOML is case-sensitive.
 - Whitespace means tab (U+0009) or space (U+0020).
--- a/toml.md
+++ b/toml.md
@@ -423,8 +423,7 @@ str = ''''That,' she said, 'is still pointless.''''
 
 Control characters other than tab are not permitted in a literal string. Thus,
 for binary data, it is recommended that you use Base64 or another suitable
-binary-to-text encoding. The handling of that encoding will be
-application-specific.
+binary-to-text encoding.
 
 ## Integer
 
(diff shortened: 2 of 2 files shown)


eksortso:
@pradyunsg Here are the isolated Unicode and UTF-8 language changes.

* Updates the changelog
* Emphasizes the UTF-8 encoding requirement
* Consistently uses U+xxxx notation for characters throughout
* Corrects misapplications of the language for UTF-8 and Unicode

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## unreleased
 
+- Clarify Unicode and UTF-8 references.
```
ChristianSi, line 5: I wonder whether this even needs to be mentioned, considering that it doesn't change anything substantially.

ChristianSi (approved):

CHANGELOG.md
eksortso: Since the body of the text has a substantial number of changes, I thought it was worth including. We could drop that line if it's really unnecessary.

ChristianSi: I could one new sentence (about the "well-formed code-unit sequence"), two deleted words (2x "valid") and the reference to "binary-to-text encoding" (which technically isn't Unicode-related). So, while I wouldn't strongly object, in my opinion this doesn't warrant a changelog entry.

marzer: @ChristianSi Changelog entries aren't a finite resource.

eksortso: If the new changelog entry conflicted with an existing one, I'd change or remove the new one. But since this is not the case, I'm keeping it.

toml.md
```diff
@@ -36,10 +36,15 @@ should be easy to parse into data structures in a wide variety of languages.
 
 ## Spec
 
+A TOML file must be a valid UTF-8 encoded Unicode document. Specifically this
+means that, should a file as a whole not form a
+[well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence),
+the file must be rejected (preferably) or ill-formed byte sequences must be
+replaced with U+FFFD as per the Unicode spec.
```
abelbraaksma, line 43: Nit: I think we should not use abbreviations like these, I’d prefer “As per the Unicode Specification”. We could even link to the specification’s homepage (though, to be fair, 25 years in, most people will just know).

toml.md
```diff
@@ -417,8 +422,9 @@ str = ''''That,' she said, 'is still pointless.''''
 ```
 
 Control characters other than tab are not permitted in a literal string. Thus,
+for binary data, it is recommended that you use Base64 or another suitable
+binary-to-text encoding. The handling of that encoding will be
+application-specific.
```
abelbraaksma, line 427: Nit: I’d prefer the wording “implementation specific”. Not all implementations will be applications (in the common meaning of that word, ie, libraries are not applications, usually).

abelbraaksma (approved):
Looks good! Two minor nits. Also, should we specify a minimum Unicode version? Iirc, UTF-8 was only introduced in version 2. Most public, free or just plain built in Unicode support is typically at 5 (for instance .NET Framework on Win7) or higher.

For most things, version wouldn’t matter. Just the introduction to the higher planes (V2) and surrogates (also V2) are relevant, I guess.

toml.md
eksortso: We're linking deep into unicode.org, and throughout the document, we say "Unicode" thirteen times. We mention no other character mapping mechanism or encoding, except for ASCII, and then only when describing numbers. How on earth will this get confused with any other standard?

The mention of ASCII alone has got my dander up.

toml.md
eksortso: I think we should just strike the entire last sentence. TOML does not handle the topic of binary encodings. Let those who do, figure all that out themselves. They don't need us saying anything.

abelbraaksma: Sounds good

toml.md
abelbraaksma: Lol. Yeah, leave the link. The nit was just about “spec” vs “specification” mainly, really ;).

eksortso: Oh, just that? Sure, we could do that. I have no objections to "Unicode specification" instead of "Unicode spec". But if we change it once in one place, how about we change it in three other places where the abbrev "spec" is used (referring to the first section, which is more "preliminaries" than "spec" anyway)? I'm willing to do that for the sake of formality. But I doubt that is reason enough to make such a change across the board.

So what do you say to one instance of "specification" and three of "preliminaries", all properly cased, to replace the abbreviation "spec"?

toml.md
```diff
@@ -417,8 +422,8 @@ str = ''''That,' she said, 'is still pointless.''''
 ```
 
 Control characters other than tab are not permitted in a literal string. Thus,
+for binary data, it is recommended that you use Base64 or another suitable
+binary-to-text encoding.
```
ChristianSi, line 426: @eksortso Please don't be so rash. If somebody makes a suggestion, wait for one or two days to give others a chance to comment before you actually act on it.

Regarding this sentence, i think it should stay. Since the spec mentions Base64, a quick reader might get the impression that a good TOML parser should support that encoding – this sentence makes it crystal clear that that's not the case.

Also, this change is entirely unrelated to Unicode. IF the sentence should be deleted (and, like I said, I don't think so) that should really be dealt with in a separate PR.

ChristianSi (changes requested):

toml.md
eksortso: @ChristianSi I am getting carried away, sorry. 24 hours after discussion for a change is a good rule of thumb here.

If I had my way, I would not mention binary data at all. Strings are not explicit binary sequences, and the people who are using strings to represent binary data don't rely on any unique features of TOML to use them. Let's just take all that out, and leave what remains: "Control characters other than tab are not permitted in a literal string."

Any new issue we encounter while addressing an old issue really should have its own PR, and it shouldn't drag the original PR into it. That is, unless we cannot separate out the changes. The original paragraph mentions "ASCII or UTF-8", so I'll argue that this change is tangential.

We could get rid of all references to ASCII in a separate PR, for instance. And I'm inclined to do so.

ChristianSi: The change from "ASCII or UTF-8 encoding" to "binary-to-text encoding" is fine (and was, at least in part, suggested by me). It's also clearly Unicode relevant since ASCII and UTF-8 were mentioned. What I object to is merely the deletion of the sentence "The handling of that encoding will be application-specific", which was a later and (IMHO) unwarranted change.

eksortso: That line was from the [original commit](https://github.com/toml-lang/toml/commit/2719d7f2822c63c9802afa32fa551b2927e05a7f) that added the "binary data" statement in the first place. The line about control codes not being allowed in literal strings was added later.

That particular commit does show some interesting history: the `\x` escape code, which was restored later, was taken out in the same commit. [At the time,](https://github.com/toml-lang/toml/issues/165#issuecomment-14471472) it was decided by @mojombo that strings would be authentic Unicode strings, and not byte sequences. When `\x` was brought back, it was within this context. But I can understand now why the language regarding binary data encoding was added.

I no longer want those lines purged. But perhaps, that whole sentiment ought to be reworded in more modern (and more correct) language.

eksortso: The most recent commit restates the binary data encoding language while discussing the escape codes. So this new language emphasizes that `\x`, `\u`, and `\U` are for Unicode, not for bytes. Mention of "application-specific" logic has been removed.

eksortso: Bear with me, @ChristianSi. I misread your objection to deleting the "application-specific" line. And @abelbraaksma you were suggesting we state that binary-to-text encodings are "implementation-specific," rather than application-specific.

Calling it implementation-specific is confusing, because as I have made clear, TOML has nothing to do with binary data. TOML parsers that implement Binary64 support are doing more than just parsing TOML. What consumers do with their strings is their business, not ours.

I'll add another line to my previous commit.

eksortso: What do you think of this replacement?

> Note that there is no low-level byte sequence syntax in TOML, and that the
> character escape codes must not be made to work in such a way. So for binary
> data, it is recommended that you use Base64 or another binary-to-text encoding,
> and that external tools ought to be used to interpret such encodings.

abelbraaksma: That's pretty clear and complete, I think. I like it.

toml.md
```diff
@@ -293,9 +298,14 @@ For convenience, some popular characters have a compact escape sequence.
 ```
 
 Any Unicode character may be escaped with the `\xHH`, `\uHHHH`, or `\UHHHHHHHH`
-forms. The escape codes must be valid Unicode
+forms. The escape codes must be Unicode
 [scalar values](https://unicode.org/glossary/#unicode_scalar_value).
 
+Note that there is no low-level byte sequence syntax in TOML, and that the
+character escape codes must not be made to work in such a way. So for binary
```
ChristianSi, line 305: I'm not quite happy with the sentence: "Note that there is no low-level byte sequence syntax in TOML, and that the character escape codes must not be made to work in such a way."

I get what you're trying to say, but what exactly is the meaning of that "must"? Would be TOML police come if you try to do so anyway? Hardly! If one really wants to do so, one could treat Unicode strings that contain just code points between U+0000 and U+00FF as byte sequences. It wouldn't make much sense to use `\xHH` to represent otherwise unwriteable bytes in such sequences, since it would entail an overhead of up to 300%, while Base64 has an overhead of only about 33%. Still, there is no law against it.

Therefore I would propose to replace these two lines with something like: "Keep in mind that all TOML strings are sequences of Unicode characters, *not* of bytes. For binary" – and then the next lines, recommending "Base64 or another binary-to-text encoding", could neatly follow.

eksortso: > If one really wants to do so, one could treat Unicode strings that contain just code points between U+0000 and U+00FF as byte sequences.

Heh, actually that doesn't work for the byte range 80 to FF, because in UTF-8, `"\x80"` through `"\xFF"` are each two-byte code point.

But @ChristianSi, I like the approach that you are taking. I will keep your first sentence about the same, but it will lead into a discussion on byte sequences instead. This allows the text to flow a little more naturally. The rest is reformulated to emphasize that the methods used to get bytes from strings and vice versa are not a part of TOML, but rather external methods. I also mention hexadecimal strings as such a method, since those show up a lot as hashes nowadays. I linked to a [Binary64](https://www.base64decode.org/) site describing that method.

Here's the most recent version of the paragraph. Would this work?

```md
Keep in mind that all TOML strings are sequences of Unicode characters, _not_
byte sequences. For binary data, avoid using these escape codes. Instead,
external binary-to-text encoding strategies, like hexadecimal sequences or
[Base64](https://www.base64decode.org/), are recommended for converting between
bytes and strings.
```

toml.md
```diff
@@ -293,9 +298,15 @@ For convenience, some popular characters have a compact escape sequence.
 ```
 
 Any Unicode character may be escaped with the `\xHH`, `\uHHHH`, or `\UHHHHHHHH`
-forms. The escape codes must be valid Unicode
+forms. The escape codes must be Unicode
 [scalar values](https://unicode.org/glossary/#unicode_scalar_value).
 
+Keep in mind that all TOML strings are sequences of Unicode characters, _not_
+byte sequences. For binary data, avoid using these escape codes. Instead,
+external binary-to-text encoding strategies, like hexadecimal sequences or
```
ChristianSi, line 306: I'm not quite convinced of the new mention of "hexadecimal sequences". If one wants to mention an alternative to Base64, why not mention [Base85](https://en.wikipedia.org/wiki/Ascii85#RFC_1924_version), which is more rather than less compact than Base64, is used by Git and supported by the Python standard library, and even has its own RFC (if originally an April Fools RFC). Or just stick to Base64, as before.

Otherwise this looks really good now!

ChristianSi (changes requested):

ChristianSi (approved):

toml.md
abelbraaksma: That’s a big improvement, I’d go for it 👍

abelbraaksma (approved):
The new preamble is very clear, and the rest of the improvements remove ambiguity. Language is hard, esp spec language, I really do appreciate you taking the time to get this right.

TLDR: LGTM!

eksortso:
Couldn't have done it without you guys @abelbraaksma @ChristianSi !

@pradyunsg Here's the Unicode and UTF-8 language cleanup that was moved out of #924 and greatly refined. Think these changes will improve the specification a lot. What do you think?

eksortso:
@pradyunsg Please review these changes and make a decision. Our work coalesced two months ago, and we sent a reminder to review this at that time. What do you think?

toml.md
```diff
@@ -34,12 +34,17 @@ should be easy to parse into data structures in a wide variety of languages.
...
 - [ABNF Grammar](#user-content-abnf-grammar)
 
-## Spec
+## Preliminaries
+
+A TOML file must be a valid UTF-8 encoded Unicode document. Specifically this
+means that, should a file as a whole not form a
+[well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence),
+the file must be rejected (preferably) or ill-formed byte sequences must be
+replaced with U+FFFD as per the Unicode specification.
```
pradyunsg, line 43: How about:

```

  Specifically this means that a file _as a whole_ must form a [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence) and will be rejected (preferably) or have ill-formed byte sequences replaced with U+FFFD as per the Unicode specification.
```

Making the above content the final bullet in the list below?

toml.md
```diff
@@ -34,12 +34,17 @@ should be easy to parse into data structures in a wide variety of languages.
 - [MIME Type](#user-content-mime-type)
 - [ABNF Grammar](#user-content-abnf-grammar)
 
-## Spec
+## Preliminaries
```
pradyunsg, line 37: Nice! That we used `Spec` here has always seemed a bit odd to me. 😅

pradyunsg (approved):
LGTM! This is _really_ nice!

One minor nit-pick again. 😅

toml.md
eksortso: @pradyunsg Sure, that's perfectly fine.

pradyunsg:
Thanks @eksortso!

abelbraaksma:
Very happy this got approved, considering the discussions that came with it, glad this is in! 👏

ChristianSi:
Last-minute changes are always dangerous. While it's good this was merged, in the last minute (without anyone being able to re-review it) the sentence

> Specifically this means that, should a file as a whole not form a [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence), the file must be rejected (preferably) or ill-formed byte sequences must be replaced with U+FFFD as per the Unicode specification.

was changed to:

> Specifically this means that a file _as a whole_ must form a [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence) and will be rejected (preferably) or have ill-formed byte sequences replaced with U+FFFD as per the Unicode specification.

Hmm! So now it seems that **every** TOML file "will be rejected (preferably)" which was probably not the intent! 

I'd suggest changing this to:

> Specifically this means that a file _as a whole_ must form a [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence), otherwise it be rejected (preferably) or ill-formed byte sequences will be replaced with U+FFFD as per the Unicode specification.

Or just return to the original wording.

@eksortso: Will you open a new PR for that? Or should I?

eksortso:
@ChristianSi I'll open the new PR. Even though @pradyunsg wrote the change, I accepted it thinking that all it was doing was restating the requirement in positive terms. I take responsibility for missing the wrong conjunction and the subsequent change in meaning.

The new wording will be as follows. I kept the first MUST clause in its own sentence. The consequences of violating this clause are specified in the second MUST ("otherwise") clause. Together, these two sentences are semantically identical to the original wording.

> Specifically this means that a file _as a whole_ must form a [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence). Otherwise, it must be rejected (preferably), or have ill-formed byte sequences replaced with U+FFFD as per the Unicode specification.

Issue #603, opened 2019-03-10
Will custom type syntax be good for TOML health?


LongTengDao:
```
key = (compute) ' 5 * 60 * 60 '

key = (toSecond) ' 5h '

key = (toTable) [
  ['name', 'age', 'sex'],# head
  ['Jack', '10', 'male'],# item 1
  ['Max', '20', 'male'],# item 2
]

key = (toDOM) '''
  <div>
    <span></span>
  </div>
'''

[table] ('and other custom transform type')
```

I don't mean the custom type syntax is a replacement of standard types. I am just wondering, maybe the exploration of de facto standards, will facilitate the development of standard types, with less discussion which hard to decide, and avoid these requirement become a dialect which will conflict with spec in the future?

eksortso:
It's an intriguing idea (definitely post-1.0). May I make a suggestion, though?

Custom types would need to be expected by parsers. Perhaps it would be better to put the custom type after the key's name, to associate that type with the key?
```
[server1]
timeout (seconds) = 300
header (toDOM) = '''
  <div>
    <span></span>
  </div>
'''
guys (csv) = [
  ['name', 'age', 'sex'],# head
  ['Jack', '10', 'male'],# item 1
  ['Max', '20', 'male'],# item 2
]
```

The parser, given the smarts to handle them, would produce converted output and handle specified constraints. For instance, `header` with the `(DOM)` type would validate the embedded string against a particular HTML standard, and `guys` with `(csv)` would hold a string in CSV format equivalent to the 2D array.

One example that intrigues me is the use of units, given this. (See #514.) `timeout (seconds) = 3m`, for instance, could assign a value of 180 to `timeout`, which expects seconds but can convert minutes appropriately.

LongTengDao:
@eksortso Oh, I think that's better and thorough!

eksortso:
@LongTenDao Well, we could have both types of syntax. Consider two different things: keys with custom type, and values with custom type.

Here's an example that might please a few people, and for good reason. The`guys` example above required that all elements be strings. That's a little painful to accept, because ages are numeric. But we can't have numbers in arrays with strings in TOML v0.5.0. But suppose our parser allows an `(m)` array to be heterogenous ("m" for "mixed type"). We could use something like `guy1 = (m)["Jack", 10, "male"]`, and then use the following for a table that will be converted to a CSV string:
```
guys (csv) = [
  ['name', 'age', 'sex'],
  (m)['Jack', 10, 'male'],
[...]

PR #948, opened 2023-01-10
Clarify that comments never affect the tables produced by parsers

eksortso:
In the discussion on #603, it became clear that no restrictions are placed on parsers that could further modify the hash tables that they generate based on included comments. The purpose of this PR is to assure that parsers will never modify a TOML document's resulting hash table due to the presence or contents of the comments contained within the document.

The change to `toml.md` will prevent a certain class of abuses of the TOML syntax, without prohibiting comment-preserving parsers from doing their thing (as those comments are not part of the resulting table but merely reference it from locations in the source document).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## unreleased
 
+- Clarify that comments never affect the tables produced by parsers.
 - Allow newline after key/values in inline tables.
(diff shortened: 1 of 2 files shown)


eksortso:
In light of recent updates, I will close this PR and reopen another.

PR #951, opened 2023-01-12
Correct requirement of well-formed code-unit sequence

eksortso:
As pointed out by @ChristianSi in #929, a last-minute change to the text for the "well-formed code-unit sequence" requirement inadvertantly invalidates all TOML files.

This PR corrects that text.

I ask that, at minimum, both @ChristianSi and @pradyunsg review this correction.

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -43,8 +43,8 @@ should be easy to parse into data structures in a wide variety of languages.
 
   Specifically this means that a file _as a whole_ must form a
   [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence).
-  Otherwise, it must be rejected (preferably), or have ill-formed byte sequences
-  replaced with U+FFFD as per the Unicode specification.
+  Otherwise, it must be rejected (preferably) or have ill-formed byte sequences
+  replaced with U+FFFD, as per the Unicode specification.
 
 ## Comment
 
--- a/toml.md
+++ b/toml.md
@@ -42,9 +42,9 @@ should be easy to parse into data structures in a wide variety of languages.
 - A TOML file must be a valid UTF-8 encoded Unicode document.
 
   Specifically this means that a file _as a whole_ must form a
-  [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence)
-  and will be rejected (preferably) or have ill-formed byte sequences replaced
-  with U+FFFD as per the Unicode specification.
+  [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence).
+  Otherwise, it must be rejected (preferably), or have ill-formed byte sequences
+  replaced with U+FFFD as per the Unicode specification.
 
 ## Comment
 


eksortso:
@pradyunsg Could you please elevate this PR? This was a serious mistake, and it needs our attention.

toml.md
```diff
@@ -42,9 +42,9 @@ should be easy to parse into data structures in a wide variety of languages.
 - A TOML file must be a valid UTF-8 encoded Unicode document.
 
   Specifically this means that a file _as a whole_ must form a
-  [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence)
-  and will be rejected (preferably) or have ill-formed byte sequences replaced
-  with U+FFFD as per the Unicode specification.
+  [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence).
+  Otherwise, it must be rejected (preferably), or have ill-formed byte sequences
```
ChristianSi, line 46: I would delete the comma before "or". It's unnecessary, and to me to the comma seems to suggest that only the U+FFFD replacement corresponds to the Unicode specification. But AFAIK either variant is Unicode-compatible.

ChristianSi (changes requested):
I would tend to change a tiny thing. Otherwise it's fine!

toml.md
eksortso: Good point. Rejecting the document is Unicode-compliant behavior too, isn't it?

The comma that I'm adding is intended to set apart the different options. So I could also put a comma before "as per" for the same effect. I actually prefer doing this, because it would make reading the sentence easier.

marzer: > Rejecting the document is Unicode-compliant behavior too, isn't it?

I hope so, that's what I'm doing in TOML++ 😅

pradyunsg:
Oops? 😅

Thanks for catching this @ChristianSi and for the quick PR on this @eksortso. Much appreciated!

toml.md
eksortso: Hey, it is! Found it [here](https://www.unicode.org/versions/Unicode15.0.0/ch03.pdf). [Unicode 15.0](https://www.unicode.org/versions/Unicode15.0.0/), Section 3.2, Conformance clause C10.

I'd like to add the second comma to the PR as described earlier. If no objections, then I'll add it in tomorrow.

eksortso: Did I lose a comment somewhere?

eksortso: I made some confusing edits to my previous comment. Here's the change that I'd like to make, which involves _adding_ another comma to the second sentence, after U+FFFD is mentioned.

> Specifically this means that a file _as a whole_ must form a [well-formed code-unit sequence](https://unicode.org/glossary/#well_formed_code_unit_sequence). Otherwise, it must be rejected (preferably), or have ill-formed byte sequences replaced with U+FFFD, as per the Unicode specification.

If this is alright, I'm putting it in tomorrow.

ChristianSi: My preference would be to add the second comma, *but still* to remove the first one. The first one reads strange to me, especially since the "must" before it applies to the "have" after it as as well.

But well, if that comma nevertheless stays, I'll survive it (somehow).

eksortso: > My preference would be to add the second comma, _but still_ to remove the first one.

A much better idea! I'll do that.

Obviously I don't want to move too quickly, since I've jumped too many guns in the past. So let me ask while everyone's still here, can I push this change now?

eksortso: Moving the comma now.

Issue #949, opened 2023-01-10
[Proposal] Allow apostrophe escaping in inline literal strings


Keyacom:
## Preface

Allow apostrophe escaping in inline literal strings.

## Why

The YAML specification nicely allows `'` to be placed inside a raw string by preceding it with itself:
```yml
ParsedString: "Emma's boyfriend is nice."
RawString: 'Emma''s boyfriend is nice.'
```
When fed to a YAML parser, both `ParsedString` and `RawString`'s values will be the same.

TOML does not allow that within the current specification. I am proposing this, so you won't need to create a multiline literal string just to include an apostrophe in it.

## Questions

You can ask more below!

### Why not `\'` instead of `''`?

This could be influenced by how Ruby parses raw strings, but I don't want that because of the potential for ambiguous behavior of `\`, which could be a potentially breaking change.

marzer:
> When fed to a YAML parser [...] 

In general I think where YAML chooses to do one thing, it would be to the betterment of TOML health to explicitly _not_ choose to do that thing. There's just [far too much magic](https://noyaml.com/) in YAML. This is a good example; why add this magic when there's already two good alternatives that don't introduce unexpected behaviour? A double-quoted string would accept your example text verbatim without issue. More complex examples can use a ML literal string.

eksortso:
To illustrate @marzer's point, here are two ways to state the string `Emma's boyfriend is nice.` in TOML without having to escape the apostrophe.

```toml
ParsedString = "Emma's boyfriend is nice."
RawString = '''Emma's boyfriend is nice.'''
```

The `RawString` example uses multi-line literal string syntax (three single-quotes on each side). Within them, you can use apostrophes, or even literal pairs of apostrophes, freely.

The website [toml.io](https://toml.io) even mentions this explicitly in its quick tour:

> Since there is no escaping, there is no way to write a single quote inside a literal string enclosed by single quotes. That's where **multi-line literal strings** come in:

though that site doesn't actually use an example like this. I'll run this past them.

arp242:
> Edit: also worth noting that this would be a backwards-incompatible change; merely updating the TOML spec version of a parser would change the way a literal string featuring '' was parsed. There's a lot of TOML out in the world already.

This isn't a concern since `''` is only parsed inside `'..'`-delimited strings; so just `''` is still an empty string. For example in SQL:

```
(1)=# select '', '''';
 ?column? │ ?column?
──────────┼──────────
          │ '
```

[...]

PR #960, opened 2023-01-19
toml.md: Add apostrophe example for multi-line literal strings

eksortso:
When I [pointed to the website](https://github.com/toml-lang/toml/issues/949#issuecomment-1377864878) when discussing how apostrophes are permitted in multi-line literal strings, I was surprised to discover that this permission is described, but it isn't made explicitly clear with an example.

This change adds such an example.

```toml
literal-apostrophes = '''You're free to use apostrophes in here.'''
```

This replaces a [PR on toml-lang/toml.io](https://github.com/toml-lang/toml.io/pull/61), created before I realized that the front page to https://toml.io is generated.

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -407,6 +407,7 @@ modification.
 
 ```toml
 regex2 = '''I [dw]on't need \d{2} apples'''
+literal-apostrophes = '''You're free to use apostrophes in here.'''
 lines  = '''
 The first newline is
 trimmed in raw strings.


arp242:
The regex2 example contains a `'`:

    regex2 = '''I [dw]on't need \d{2} apples'''

And it's a bit more explicit in the next block of examples:

    # 'That,' she said, 'is still pointless.'
    str = ''''That,' she said, 'is still pointless.''''

eksortso:
@arp242 The reason I'm doing this is so that the automatically generated front page to [toml.io](https://toml.io) will show an apostrophe example inside a multi-line literal string where the possibility is mentioned.

ChristianSi:
In what way the front page is generated from the spec? To me that seems astonishing, since both the text and the examples seem to be fairly different.

eksortso:
@ChristianSi I'll confess. I don't know if `index.html` is actually generated; I operated on a [note](https://github.com/toml-lang/toml.io/pull/61#issuecomment-1384596788) from @pradyunsg that the page may be generated. I saw [`toml.io`/`lib`/`spec_to_html.js`](https://github.com/toml-lang/toml.io/blob/main/lib/spec_to_html.js) in the website's code base, and guessed that that script was used, at least in the beginning, to generate `index.html` based on the spec documents.

But I just discovered something that renders this PR moot. We have a decent example of this use case already in the spec:

```toml
regex2 = '''I [dw]on't need \d{2} apples'''
```

So this MLL string has an apostrophe in it.

Do we need to make this more clear? If not, I'll just close this PR.

ChristianSi:
@eksortso I thought you wanted an example added to the *front page*? That would make sense to me, but here is probably not the place to discuss it.

As for the spec, I too think it's fine. @arp242 had already pointed out that example and another one [above](https://github.com/toml-lang/toml/pull/960#issuecomment-1397056741).

eksortso:
I was just wondering if we're making that obvious to our normal mortal human being audience, not just those of us who see regular expressions in their dreams. I never have, but I use them a lot, including inside config files where I'd not have expected to find them.

This PR is gone.

PR #961, opened 2023-02-09
Change wording for literal string in example

eksortso:
This corrects an instance of the wrong terminology being used to describe the nature of multi-line literal strings. The original author wrote "raw strings" in the example, but TOML doesn't have "raw" strings, but rather, "literal" strings.


ChristianSi (approved):
Sure thing!

eksortso:
Thanks @ChristianSi and @pradyunsg

PR #977, opened 2023-05-27
Capitalize the unicodes in `toml.abnf`

KrLite:
I suppose that the `6e` here should be capitalized like other unicodes in the file.
This change unifies the file without affecting any of the syntax.


eksortso:
Good catch!

LGTM!

pradyunsg:
Thanks for this PR! :)

Issue #966, opened 2023-03-09
Clarify that key uniqueness depends only on binary representation, recommend normalization


SnoopJ:
I've just learned about #891 and I'm excited to see that the TOML specification is improving Unicode support.

Do I understand right that this changeset makes no recommendations for implementers when it comes to equivalence of keys? I see [a note on normalization](https://github.com/toml-lang/toml/issues/687#issuecomment-568020667) on the related issue, but if I understand the PR correctly, keys that are "equivalent" under one of the normalization forms of [UAX#15](https://unicode.org/reports/tr15/) will be distinct under the specification unless their binary representations are identical.

That note suggests that a warning/suggestion for implementers might be added to the spec, but it looks like that never happened. This issue is a request that such a note be added to at least make implementers (and users of parsers that don't bother with normalization) aware of the potential confusion of keys, as in the example of `ñaña` (NFC form, 6 bytes in UTF-8) vs. `ñaña` (NFKD form, 8 bytes in UTF-8).

pradyunsg:
See also #954

eksortso:
We did discuss key comparisons and issues regarding normalization in great detail in the discussion on #941. It's a very deep issue! In fact, some languages and tools tend to perform normalization implicitly.

@SnoopJ I personally like the idea that keys be normalized during comparisons, but we shouldn't make normalized or binary comparisons mandatory in all cases. In old-fashioned parlance, that's not a MUST nor a MAY, but a SHOULD.

During that past discussion, I suggested something like the following be added to `toml.md`, though this is different from the original in that I heavily favored NFC in my original draft:

> Because [some keys with different code points look the same](https://en.wikipedia.org/wiki/Unicode_equivalence), parsers should compare keys using a specific [normalized form](https://unicode.org/reports/tr15/#Norm_Forms) of the keys, rather than just using binary comparisons.

What are your thoughts on this text? Could we make the point more clear?

And also, what about including this line for TOML emitters?

> Likewise, encoders should write keys using a specific normalized form.

Thoughts?

ChristianSi:
@eksortso I like the sentence about parsers, but instead of "a specific normalized form" I'd write, "Normalization Form C (NFC) or Normalization Form D (NFD)". The NFK... forms normalize too aggressively (equating `…` with `...`, say), hence they **must not** be used!

I'm less sure the sentence about encoders is needed, but I guess it makes sense.

marzer:
[...]

PR #979, opened 2023-06-02
Backout Unicode bare keys

arp242:
This backs out the unicode bare keys from #891.

This does *not* mean we can't include it in a future 1.2 (or 1.3, or whatever); just that right now there doesn't seem to be a clear consensus regarding to normalisation and which characters to include. It's already the most discussed single issue in the history of TOML.

I kind of hate doing this as it seems a step backwards; in principle I think we *should* have this so I'm not against the idea of the feature as such, but things seem to be at a bit of a stalemate right now, and this will allow TOML to move forward on other fronts.

It hasn't come up *that* often; the issue (#687) wasn't filed until 2019, and has only 11 upvotes. Other than that, the issue was raised only once before in 2015 as far as I can find (#337). I also can't really find anyone asking for it in any of the HN threads on TOML.

Reverting this means we can go forward releasing TOML 1.1, giving people access to the much more frequently requested relaxing of inline tables (#516, with 122 upvotes, and has come up on HN as well) and some other more minor things (e.g. `\e` has 12 upvotes in #715).

Basically, a lot more people are waiting for this, and all things considered this seems a better path forward for now, unless someone comes up with a proposal which addresses all issues (I tried and thus far failed).

I proposed this over here a few months ago, and the responses didn't seem too hostile to the idea:
https://github.com/toml-lang/toml/issues/966#issuecomment-1469883434

### DIFF
--- a/toml.abnf
+++ b/toml.abnf
@@ -50,19 +50,7 @@ key = simple-key / dotted-key
 val = string / boolean / array / inline-table / date-time / float / integer
 
 simple-key = quoted-key / unquoted-key
-
-;; Unquoted key
-
-unquoted-key = 1*unquoted-key-char
-unquoted-key-char = ALPHA / DIGIT / %x2D / %x5F         ; a-z A-Z 0-9 - _
-unquoted-key-char =/ %xB2 / %xB3 / %xB9 / %xBC-BE       ; superscript digits, fractions
-unquoted-key-char =/ %xC0-D6 / %xD8-F6 / %xF8-37D       ; non-symbol chars in Latin block
-unquoted-key-char =/ %x37F-1FFF                         ; exclude GREEK QUESTION MARK, which is basically a semi-colon
-unquoted-key-char =/ %x200C-200D / %x203F-2040          ; from General Punctuation Block, include the two tie symbols and ZWNJ, ZWJ
-unquoted-key-char =/ %x2070-218F / %x2460-24FF          ; include super-/subscripts, letterlike/numberlike forms, enclosed alphanumerics
-unquoted-key-char =/ %x2C00-2FEF / %x3001-D7FF          ; skip arrows, math, box drawing etc, skip 2FF0-3000 ideographic up/down markers and spaces
-unquoted-key-char =/ %xF900-FDCF / %xFDF0-FFFD          ; skip D800-DFFF surrogate block, E000-F8FF Private Use area, FDD0-FDEF intended for process-internal use (unicode)
-unquoted-key-char =/ %x10000-EFFFF                      ; all chars outside BMP range, excluding Private Use planes (F0000-10FFFF)
+unquoted-key = 1*( ALPHA / DIGIT / %x2D / %x5F ) ; A-Z / a-z / 0-9 / - / _
 
 ;; Quoted and dotted key
 
--- a/toml.md
+++ b/toml.md
@@ -106,11 +106,9 @@ first = "Tom" last = "Preston-Werner" # INVALID
 
 A key may be either bare, quoted, or dotted.
 
-**Bare keys** may contain any letter-like or number-like Unicode character from
-any Unicode script, as well as ASCII digits, dashes and underscores.
-Punctuation, spaces, arrows, box drawing and private use characters are not
-allowed. Note that bare keys are allowed to be composed of only ASCII digits,
-e.g. 1234, but are always interpreted as strings.
+**Bare keys** may only contain ASCII letters, ASCII digits, underscores, and
+dashes (`A-Za-z0-9_-`). Note that bare keys are allowed to be composed of only
+ASCII digits, e.g. `1234`, but are always interpreted as strings.
 
 &#x2139;&#xfe0f; The exact ranges of allowed code points can be found in the
 [ABNF grammar file][abnf].
@@ -120,23 +118,18 @@ key = "value"
 bare_key = "value"
 bare-key = "value"
 1234 = "value"
-Fuß = "value"
-😂 = "value"
-汉语大字典 = "value"
-辭源 = "value"
-பெண்டிரேம் = "value"
 ```
 
 **Quoted keys** follow the exact same rules as either basic strings or literal
-strings and allow you to use any Unicode character in a key name, including
-spaces. Best practice is to use bare keys except when absolutely necessary.
+strings and allow you to use a much broader set of key names. Best practice is
+to use bare keys except when absolutely necessary.
 
 ```toml
 "127.0.0.1" = "value"
 "character encoding" = "value"
+"ʎǝʞ" = "value"
+'key2' = "value"
 'quoted "value"' = "value"
-"╠═╣" = "value"
-"⋰∫∬∭⋱" = "value"
 ```
 
 A bare key must be non-empty, but an empty quoted key is allowed (though
@@ -157,7 +150,6 @@ name = "Orange"
 physical.color = "orange"
 physical.shape = "round"
 site."google.com" = true
-பெண்.டிரேம் = "we are women"
 ```
 
 In JSON land, that would give you the following structure:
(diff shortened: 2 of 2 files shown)


eksortso:
Yeah, rolling back all the work done to allow more bare keys is a very bad idea.

Don't confuse bare keys with the issues surrounding potential normalization of key strings. Quoted keys could always raise these issues. And we can wait until after v1.1.0 is released to address normalization issues, should they arise in real life.

@arp242 I think I know where you're coming from by floating this proposal. I once opened a PR, #663, that I hesitated to suggest but seemed to have some support in its favor, in order to advance discussion. Ultimately it was closed, but it certainly advanced discussion, and it prompted a counter proposal, #676, that did have broad consensus.

That said, someone may think to push a PR to move the normalization debate. I would do so if I had more time, but I've left plenty of proposal language in the discussions which anyone is invited to use.

arp242:
> rolling back all the work done

I don't really see it so much as "rolling back" but rather as "delaying". No work is really lost.

I'd just like to move forward with TOML. Right now no one seems 100% happy with how things are with the unicode keys (although the reasons for the unhappiness are very varied, which is part of the problem) and it's been like that for over a year now.

> Don't confuse bare keys with the issues surrounding potential normalization of key strings. Quoted keys could always raise these issues.

Yes, but very few people write quoted keys so it's much less of an issue (almost a non-issue in reality). It's not a surprise this has gone largely unnoticed/undiscussed before the Unicode PR was merged.

> And we can wait until after v1.1.0 is released to address normalization issues, should they arise in real life.

Possibly solutions such as "only normalized forms will be allowed in TOML bare keys" will be closed off due to backwards compatibility, so I don't really agree with that. I have said this many times before, but I very much think we should consider both issues together.

arp242:
> Ultimately it was closed, but it certainly advanced discussion, and it prompted a counter proposal, https://github.com/toml-lang/toml/pull/676, that did have broad consensus.

I did spend quite a bit of time trying to come up with a better proposal which should address all issues by the way, but it's not so easy. By far the best way would be to spec out all problems so they can't occur, but then the Unicode ranges get much larger – I managed to reduce it to about 100, which is still quite a lot.

It's a problem a lot of environments have struggled with. I thought I mentioned that in the commit message but it seems that got lost in editing.

ChristianSi:
I'm 1000% opposed to this proposal and I'm very relieved to see that it doesn't seem to have any realistic chance anyway. As I see it, bare keys written in any natural language will the **the** big new feature and advantage of TOML 1.1 if and when it comes.

As for the million dollar question: I don't think TOML 1.1 is blocked by the normalization issue, rather it's blocked by nobody working on it. I once tried to [triage](https://github.com/toml-lang/toml/issues/928#issuecomment-1294659512) the issues that should be addressed in some way or other for it to reach completion. Possibly one or two of the issues mentioned there have meanwhile been resolved, and probably a few new ones have emerged, but I think that analysis is still largely valid.

In any ideal world, there would be somebody working on these issues, trying to find a suitable solution (which might include closing as "out of scope" or deferring after 1.1) for each of them, unless finally an RC for 1.1 is ready to be published. In a *perfect* world, that person or persons would be TOML maintainers – that world, sadly, is not in sight, but even some other community member(s) should be able to make progress if they focus on this task.

Maybe there are any volunteers here willing to pick it up? (I can't, due to other obligations.) @arp242: Maybe it would be something for you, seeing that you seem very interested in getting 1.1 finished?

arp242:
I don't think we really *need* to do anything for TOML 1.1 @ChristianSi; anything can be in any future version, so any "milestones" are rather arbitrary, IMHO.

pradyunsg:
Honestly, I'm not opposed to this idea as strongly as others seem to be. While I do agree that allowing bare keys across the board would be an improvement, the argument for "this is complicated and no one has the time to work on it, let's get to this later" is a sound one.

> any "milestones" are rather arbitrary, IMHO.

FWIW, I agree with this. Don't use what's been put in the milestones as a hammar to shut down a line of argument. :)

---

I think it comes down to how important we think it is to have this in this release rather than delaying it for a future release. IIUC, @arp242 thinks that it's worth getting a release out sooner, while @eksortso and @ChristianSi think that it's worth resolving the outstanding issues with bare keys.

I don't know yet what my preferences are -- I haven't caught up on the entirety of the bare keys discussion yet, to have an updated sense of the complexities involved -- but I don't think that this "doesn't seem to have any realistic chance". It is very much possible that we go down this route, if the bare keys discussion doesn't reach a resolution / have a feasible solution that there's sufficient agreement on.

Two arguments against deferring this for a new release are that:

- It's unlikely that implementers will pick a new TOML version to just support this additional complexity, so it makes sense to couple the functionality that people want with complicated-but-valuable functionality that enables non-english TOML configuration files to be easy to write as well.
- As a configuration format specification which would end up serving as a very-low-level piece of language and tooling ecosystems, each release of TOML requires a bunch of work from those downstream ecosystems. Fewer releases will mean less work for those downstream ecosystems.[^1]

[^1]: I'm biased, as I expect that I'll likely end up doing a bunch of work on the Python side of this.

arp242:
> "this is complicated and no one has the time to work on it, let's get to this later" is a sound one.

It's not even "lack of time", it's that I simply can't find a solution which would satisfy everyone. This may be a failure on my part though, but it's not just "lack of time", it's "we can't agree on a good way to proceed on all these issues". Every possible resolution I can see has strong opposition from at least one person.

ChristianSi:
Let's keep in mind that the true problem is not at all this one, but #966. And I think that one *is* solvable. But even if we decide to leave the normalization issue unsolved in 1.1 and address it only later, nothing we would be last.

pradyunsg:
Yea, I'd mentally bundled those together as "bare keys need changes before being good-enough for release".

arp242:
I don't think this entire feature is worth it: the issue doesn't have that many upvotes, the PR has zero, when I did this PR I searched for people asking for this feature in HN threads about TOML and I couldn't find anyone (I saw a lot of people complaining about a lot of stuff, it is HN after all, but not this), YAML allows it but no one seems to be using it, we haven't had people ask about it in issues or anywhere else.

In short, it's not really solving a significant practical problem. I say this as someone who is actually using it right now (my `Sinn_Féin` key that I mentioned the other day). Yes, it will be useful on occasion, but I feel that >95% of this is "adding it for the sake of adding it".

And that would be acceptable if it's a small thing with simple implications. But it's not.

TOML is for configuration files. That's what we're always saying, right? How often do you want non-ASCII in configuration files? Rarely. My `Sinn_Féin` usage is as a "data file" rather than "config file".

The best we can do is "we're going to do a kind of bad half-Unicode implementation, but because not a lot of people will use it, probably, it's not going to be a problem, probably, but if people do it will be a problem."

Meh.

That's basically what we have now. #990 improves on that IMHO, but still doesn't really solve everything. The only way we can solve it is by going full-in Unicode, but people don't want that either, for reasonable reasons, so we just can't win.

We've spent more time on this than it will ever save. This is increasingly just a sunk-cost fallacy.

pradyunsg:
My 2 cents: Let's back this out and kick the can down the road for this.

I still think it's a good idea to permit Unicode in bare keys but, as I see it, the underlying issue here is that ABNF doesn't let us clearly reflect what we want here (which is Unicode character classes + normalisation semantics) which vary by Unicode version as well as the specific normalization approach selected.

I do think that we ought to have something in TOML 1.2.0 in some way in addition to this feature (that's assuming we're able to figure out ABNF + Unicode situation in the future) so that implementors have a reason to move forward for this.

ChristianSi:
For normalization, however, I'm quite sure I remember that for all practical purposes we had already agreed NOT to normalize, so that's pretty much a non-issue.

pradyunsg (approved):
Let's do this.

ChristianSi:
So, any objections to merging this? If not, I'll do it in a few days (unless somebody else beats me to it).

PR #996, opened 2023-10-01
Revert "Permit more control characters in comments (#924)"

### LINKED ISSUE #995: Excluding LF in 'allowed-comment-char' is confusing
I'm adding a test to toml-test for "Relax comment parsing; most control characters are again permitted.", and implementing it in my parser, and I'm rather confused what 0x0a (LF) is excluded. You need to support this if you want CRLF, no?

There's a long thread in #924 about this, but to be honest I stopped reading after a few comments – people shouldn't need to read entire threads to understand the specification.

I think the intent is that "CR" and "CRLF" are allowed, but "LF" isn't? That is what it does, I think, but it's really not very obvious IMHO.

At the very least there should be a comment in the ABNF, and the phrasing in toml.md ("that could cause problems during editing or processing: U+0000, and U+000A to U+000D") could also be improved. Now it's too easy to be misinterpreted as "only CR newlines are allowed and LF is forbidden, full stop", which isn't what was intended.

arp242:
This reverts commit ab7495887ccd0f567d8a220e1a5ab0ca6fdbd16f.

I'm a simple guy. Someone reports a problem, I drink coffee and fix it. No one reports a problem? There is nothing to fix and I go drink beer.

No one really reported this as a problem, but it *does* introduce needless churn for all TOML implementations and the test suite. Do we need to forbid *anything* in comments? Probably not, and in strings we probably only need to forbid \x00. But at least before it was consistent with strings, and more importantly, what everyone wrote code for, which is tested, and already works.

[None of the hypotheticals](https://github.com/toml-lang/toml/issues/567#issuecomment-568269120) on why this is "needed" are practical issues people reported, and most aren't even fixed: a comment can still invalidate the file, you must still parse each character in a comment as some are still forbidden, the performance benefits are very close to zero they might as well be zero, and you still can't "dump whatever you like" in comments.

So it doesn't *actually* change anything, it just changes "disallow this set of control characters" to ... "disallow this set of control characters" (but for a different set). That's not really a substantial or meaningful change. The only (minor) real-world issue that was reported (from the person doing the Java implementation) was that "it's substantially more complicated to parse out control characters in comments and raise an error, and this kind of strictness provides no real advantage to users". And that's not addressed at all with this, so...

---

And while I'm at it, let me have a complaint about how this was merged:

1. Two people, both of whom actually maintain implementations, say they don't like this change.
2. This is basically ignored.
3. Three people (none of whom maintain an implementation, AFAIK) continue written a fairly large number of large comments, so anyone who wasn't already interested in this change unsubscribes and/or goes 🤷
4. "Consensus".

Sometimes I feel TOML attracts people who like to argue things from a mile-high ivory tower with abstract arguments that have only passing familiarity with any actual pragmatic reality.

Fixes #995

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 ## unreleased
 
 - Clarify Unicode and UTF-8 references.
-- Relax comment parsing; most control characters are again permitted.
 - Allow newline after key/values in inline tables.
 - Allow trailing comma in inline tables.
 - Clarify where and how dotted keys define tables.
--- a/toml.abnf
+++ b/toml.abnf
@@ -37,10 +37,11 @@ newline =/ %x0D.0A  ; CRLF
 
 ;; Comment
 
-comment = comment-start-symbol *allowed-comment-char
 comment-start-symbol = %x23 ; #
-allowed-comment-char = %x01-09 / %x0E-7F / non-ascii
 non-ascii = %x80-D7FF / %xE000-10FFFF
+non-eol = %x09 / %x20-7E / non-ascii
+
+comment = comment-start-symbol *non-eol
 
 ;; Key-Value pairs
 
--- a/toml.md
+++ b/toml.md
@@ -57,9 +57,8 @@ key = "value"  # This is a comment at the end of a line
 another = "# This is not a comment"
 ```
 
-Comments may contain any Unicode code points except the following control codes
-that could cause problems during editing or processing: U+0000, and U+000A to
-U+000D.
+Control characters other than tab (U+0000 to U+0008, U+000A to U+001F, U+007F)
+are not permitted in comments.
 
 ## Key/Value Pair
 


ChristianSi:
> No one really reported this as a problem

I'd consider @abelbraaksma [comment](https://github.com/toml-lang/toml/issues/567#issuecomment-568269120) that triggered the whole change indeed a bug report, even though it's not a bug he had run into personally. He got 4 thumps-ups, and in the whole discussion that followed in that issue not a single person argued that the restrictive 1.0 status (where comments must fulfill most string requirements, despite not really being strings) should be kept. Plus, though you say that implementation maintainers didn't like this, the [very next comment](https://github.com/toml-lang/toml/issues/567#issuecomment-1272505041) says:

> As an implementation author, I agree with @abelbraaksma above.

As for the later discussion in #924, I suppose you refer to yourself (https://github.com/toml-lang/toml/pull/924#issuecomment-1276851773) and @hukkin (https://github.com/toml-lang/toml/pull/924#issuecomment-1278831310) expressing doubts? While the latter talked about "potential busywork for implementation maintainers", their main concern seem to have been about "about allowing lone CR in comments" – and this was subsequently addressed by indeed keeping them forbidden. Plus another maintainer, @marzer, explicitly welcomed the change (https://github.com/toml-lang/toml/pull/924#issuecomment-1279967421: "status quo bad").

And you yourself write above:

> Do we need to forbid anything in comments? Probably not...

Arguably, the current main, which you wish to revert, comes closer to that ideal ("forbid as little as possible in strings"), while the old 1.0 standard is to forbid a lot, without anyone seeming to know exactly why.

Also I'm a bit concerned by your suggestion that TOML maintainers are more important than other people here, that is, TOML users. Now admittedly, maintainers are essential, since without them there would be nothing to so. But users are important too, since without them there would be no point in maintaining anything. So I'd suggest not to play them out against each other.

Plus there is the workload of future maintainers, of all the TOML implementations that don't exist yet but will hopefully be written some day. I'd say that while the difference is not large, for them life has become a bit simpler thanks to the accepted change (#924). In 1.0, comments are essentially a fifth string type – not exactly matching any of the four actual string types, but still having similar requirements when it comes to error checking. In the future, error checking in strings is considerably simplified, since you only have to check for CR (if line splitting has already happened) or you have to check for CR or LF to find the end of the comment (if not –with the former, if not followed by LF, signaling an error). That's arguable more simple than having to scan for arbitrary control characters and confusing the user with strange error messages ("illegal comment").

ChristianSi:
A small correction: Rechecking the ABNF, I see that VT (vertical tab) and FF (form feed) are still forbidden too, apparently due to https://github.com/toml-lang/toml/pull/924#issuecomment-1279999132. That wasn't much discussed at the time. If you (or anybody else here) makes the case, that they too should be allowed in comments, to make the code simpler and more consistent, I think I'd be on board with that – if they occur at all, they're unlikely to cause problems, so there doesn't seem to be a particular reason to still forbid them.

ChristianSi:
Somebody has [found out](https://github.com/toml-lang/toml/issues/1013) that 1.0 is inconsistent in its rules regarding comments – specifically, the ABNF allows U+007F (DEL), while the written spec forbids it. So "revert to the old status quo" is simply not an option, since the old status quo is buggy and needs to be fixed.

pradyunsg:
That was already spotted and fixed (as mentioned in #995's closing comment).

ChristianSi:
Spotted and fixed, yes, but only if we _don't_ revert. So let's not – the old status quo, since buggy, is simply no longer an option.

arp242:
That's a simple typographical fix, unrelated to the issues raised here.

eksortso:
@arp242 Since you have started [rejecting user contributions in toml-test](https://github.com/toml-lang/toml-test/pull/159#issuecomment-2169225000) that conform to the current unreleased TOML spec but disagree with what your particular take is, it's time for this issue to be settled.

@pradyunsg In light of the conversation here over the past eight months, and in related feedback on the changes since #924, what is your take on this PR?

arp242:
We can't add tests for it because the TOML library needs updating, and that's not trivial (not super-hard, but not trivial). I don't see the point of spending time on it because this improves or fixes nothing for any users, as explained before.

I'm not doing that, regardless of what ends up in TOML 1.1. I got tons of actually useful stuff to do.

If you have deep philosophical problems with that then find someone else to maintain toml-test. I'm serious. To hell with "but akshually, I have better feels with this" changes that don't objectively improve anything and that don't address practical problems from random people who haven't put in any effort in anything. And no, I'm not going to engage in lengthy discussions endlessly responding to 6 paragraphs of woolly text. I have actual code to write. I spent tons of time on toml-test and associated things like the text-matrix, including on some things I don't really agree with, but there are limits.

epage:
For reference, I'm the maintainer of the primary Rust TOML library and a member of the Cargo team though I'm not speaking on behalf of them.

While I agree that we should revert the more controversial parts of 1.1 to get it out and re-examine how decisions are made (maybe my Rust bias but I feel like a more formalized RFC process should be adopted due to the ramifications of changes)., I don't think the spec and the conformance test suite should be limited by the decisions of an individual implementation maintainer.   I know for Cargo, we've "bet big" on TOML but if it came to it, we could fork if needed.  Thats different (and I watch this repo in the hopes that I can help avoid that).

If an implementation is needed to match the spec, I could work to add unstable support for 1.1 to Rust's TOML parser.  It might be a few weeks out as I am on vacation and have some other obligations to wrap up.

pradyunsg:
> @pradyunsg In light of the conversation here over the past eight months, and in related feedback on the changes since #924, what is your take on this PR?

Let's revert this change now, and revisit this for a TOML 1.2.0.

pradyunsg:
I've merged this in the spirit of "let's lean toward action over inaction here" + given that this is the direction I'd prefer for us to take here. Holler if you got concerns by filing a new issue please!

PR #1002, opened 2023-10-27
Extended bare key ranges to include all emojis

### LINKED ISSUE #954: Not all emojis work as bare keys
I was writing test cases for this, and using a pirate flag (🏴‍☠️) doesn't work; this is:

         CPoint  Dec    UTF8        HTML       Name (Cat)
    '🏴' U+1F3F4 127988 f0 9f 8f b4 &#x1f3f4;  WAVING BLACK FLAG (Other_Symbol)
    '�'  U+200D  8205   e2 80 8d    &zwj;      ZERO WIDTH JOINER (Format)
    '☠'  U+2620  9760   e2 98 a0    &#x2620;   SKULL AND CROSSBONES (Other_Symbol)

The flag and ZWJ is fine, but the skull and crossbones isn't allowed in the current range.

Seems confusing since most emojis work. Took me quite a bit of time to figure when modifying my parser to support this because I just assumed I missed something, but turns out it's just not in the allowed range:

    unquoted-key-char =/ %x2070-218F / %x2460-24FF          ; include super-/subscripts, letterlike/numberlike forms, enclosed alphanumerics
    unquoted-key-char =/ %x2C00-2FEF / %x3001-D7FF          ; skip arrows, math, box drawing etc, skip 2FF0-3000 ideographic up/down markers and spaces

Looking at the U+2500..U+2bff range, I don't really see why we need to skip a lot of these things.

---

I know we discussed this before, but I still think we should either allow only letters+numbers or just allow almost everything (with a few exceptions); the current behaviour is just confusing. The examples uses an emoji as an example and ZWJ is explicitly allowed, so you'd expect all emojis to work, but turns out only *some* emojis work. It just so happened by chance that "pirate flag" was the first emoji I tried, but there are probably others as well and with ZWJ combinations it'll be a whack-a-mole.

Either way, IMHO we should support all emojis or none. Many other ZWJ combinations do work fine; 🏳️‍🌈 (U+1F3F3 ZWJ U+1F308) or 🏴󠁧󠁢󠁷󠁬󠁳󠁿 is okay, but 🏳️‍⚧️ isn't (as U+26A7 isn't in the allowed range). In a quick test it seems all flags work, except two.

_Originally posted by @arp242 in https://github.com/toml-lang/toml/issues/891#issuecomment-1383980000_

ChristianSi:
Also explain better what's allowed in bare keys and remove the emoji example, since (though it's possible) we don't advice using emojis as bare keys.

Fixes #954.

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -110,7 +110,7 @@ are also allowed, while whitespace and private use characters are forbidden. It
 is advisable to use only words (in arbitrary languages) as bare keys, as they
 are generally accepted, while not all symbols and punctuation marks are. If you
 want to use a bare key made up of several words, use a suitable separator
-character (such as a underscore or hyphen) between the words, as spaces are not
+character (such as an underscore or hyphen) between the words, as spaces are not
 allowed. Note that bare keys are allowed to be composed of only digits, e.g.
 1234, but are always interpreted as strings.
 
--- a/toml.md
+++ b/toml.md
@@ -104,10 +104,15 @@ first = "Tom" last = "Preston-Werner" # INVALID
 A key may be either bare, quoted, or dotted.
 
 **Bare keys** may contain any letter-like or number-like Unicode character from
-any Unicode script, as well as ASCII digits, dashes and underscores.
-Punctuation, spaces, arrows, box drawing and private use characters are not
-allowed. Note that bare keys are allowed to be composed of only ASCII digits,
-e.g. 1234, but are always interpreted as strings.
+any Unicode script, as well as digits, dashes and underscores. Various symbols
+(such as emojis or arrows) and punctuation marks from outside the ASCII range
+are also allowed, while whitespace and private use characters are forbidden. It
+is advisable to use only words (in arbitrary languages) as bare keys, as they
+are generally accepted, while not all symbols and punctuation marks are. If you
+want to use a bare key made up of several words, use a suitable separator
+character (such as a underscore or hyphen) between the words, as spaces are not
+allowed. Note that bare keys are allowed to be composed of only digits, e.g.
+1234, but are always interpreted as strings.
 
 &#x2139;&#xfe0f; The exact ranges of allowed code points can be found in the
 [ABNF grammar file][abnf].
@@ -118,7 +123,6 @@ bare_key = "value"
 bare-key = "value"
 1234 = "value"
 Fuß = "value"
-😂 = "value"
 汉语大字典 = "value"
 辭源 = "value"
 பெண்டிரேம் = "value"
--- a/toml.abnf
+++ b/toml.abnf
@@ -58,8 +58,11 @@ unquoted-key-char =/ %xB2 / %xB3 / %xB9 / %xBC-BE       ; superscript digits, fr
 unquoted-key-char =/ %xC0-D6 / %xD8-F6 / %xF8-37D       ; non-symbol chars in Latin block
 unquoted-key-char =/ %x37F-1FFF                         ; exclude GREEK QUESTION MARK, which is basically a semi-colon
 unquoted-key-char =/ %x200C-200D / %x203F-2040          ; from General Punctuation Block, include the two tie symbols and ZWNJ, ZWJ
-unquoted-key-char =/ %x2070-218F / %x2460-24FF          ; include super-/subscripts, letterlike/numberlike forms, enclosed alphanumerics
-unquoted-key-char =/ %x2C00-2FEF / %x3001-D7FF          ; skip arrows, math, box drawing etc, skip 2FF0-3000 ideographic up/down markers and spaces
+unquoted-key-char =/ %x2070-21FF / %x2300-24FF          ; skip math operators
+unquoted-key-char =/ %x25A0-268B / %x2690-2757          ; skip box drawing, block elements, and some yin-yang symbols
+unquoted-key-char =/ %x2762-2767 / %x2776-27E5          ; skip some Dingbat punctuation
+unquoted-key-char =/ %x2801-297F                        ; skip some math brackets and arrows, and braille blank
+unquoted-key-char =/ %x2B00-2FFF / %x3001-D7FF          ; skip various math operators and symbols, and ideographic space
 unquoted-key-char =/ %xF900-FDCF / %xFDF0-FFFD          ; skip D800-DFFF surrogate block, E000-F8FF Private Use area, FDD0-FDEF intended for process-internal use (unicode)
 unquoted-key-char =/ %x10000-EFFFF                      ; all chars outside BMP range, excluding Private Use planes (F0000-10FFFF)
 


ChristianSi:
This is an alternative fix for #954. It extends the bare key range to allow arbitrary emojis and generally many harmless symbols that were so far excluded. At the same time it continues to allow arbitrary words in arbitrary languages in bare keys, which I consider very important for reasons for fairness and proper internationalization. And the pragmatic approach of simply defining ranges is preserved, instead of switching to an approach based on Unicode character classes that would make implementation considerably more complicated for those who don't have access to a fully-fledged Unicode support library.

The language of the written spec is also adapted to better describe what's now allowed in bare keys, and the emoji example has been deleted. (Using more or less arbitrary emojis as bare keys is now possible, but it's not something we recommend, as words make more meaningful keys.)

The extended ranges still try to exclude, in so far as reasonably possible, characters that are "problematic" in bare keys since they look similar to TOML's own meaningful punctuation (especially quotation marks, hashes, equals signs, commas, parentheses). This is just a "reasonable best effort" and not meant to be totally comprehensive – truly ensuring that visual inspection of a file gives identical results to what a parser sees is in any case impossible with Unicode (even in the total absence of bare keys).

ChristianSi:
Here's a list of all the new characters added by this PR: https://gist.github.com/ChristianSi/f3d97247c79d234326c47779227b1ff0.

Note that I didn't include the first few emojis (#️ – hash sign, *️ – asterisk, ©️ – copyright, ®️ – registered, ‼️ – double exclamation mark, ⁉️ – exclamation question mark), since they seem more like punctuation or special markers than emojis to me, and of course the hash sign is reserved for comments in TOML. All the other emojis listed in https://unicode.org/Public/UNIDATA/emoji/emoji-data.txt should be included, however.

Let me know what you think!

ChristianSi:
@pradyunsg Can you take a look, please? Time to get 1.1 closer a release candidate!

arp242:
It's not really about "emoji" but about consistent character ranges; i.e. "this type is allowed, and this type isn't".

Here's just the first thing I looked it. It currently skips:

	% uni print 2768..2775
	'❨'  U+2768  10088  e2 9d a8    &#x2768;   MEDIUM LEFT PARENTHESIS ORNAMENT (Open_Punctuation)
	'❩'  U+2769  10089  e2 9d a9    &#x2769;   MEDIUM RIGHT PARENTHESIS ORNAMENT (Close_Punctuation)
	'❪'  U+276A  10090  e2 9d aa    &#x276a;   MEDIUM FLATTENED LEFT PARENTHESIS ORNAMENT (Open_Punctuation)
	'❫'  U+276B  10091  e2 9d ab    &#x276b;   MEDIUM FLATTENED RIGHT PARENTHESIS ORNAMENT (Close_Punctuation)
	'❬'  U+276C  10092  e2 9d ac    &#x276c;   MEDIUM LEFT-POINTING ANGLE BRACKET ORNAMENT (Open_Punctuation)
	'❭'  U+276D  10093  e2 9d ad    &#x276d;   MEDIUM RIGHT-POINTING ANGLE BRACKET ORNAMENT (Close_Punctuation)
	'❮'  U+276E  10094  e2 9d ae    &#x276e;   HEAVY LEFT-POINTING ANGLE QUOTATION MARK ORNAMENT (Open_Punctuation)
	'❯'  U+276F  10095  e2 9d af    &#x276f;   HEAVY RIGHT-POINTING ANGLE QUOTATION MARK ORNAMENT (Close_Punctuation)
	'❰'  U+2770  10096  e2 9d b0    &#x2770;   HEAVY LEFT-POINTING ANGLE BRACKET ORNAMENT (Open_Punctuation)
	'❱'  U+2771  10097  e2 9d b1    &#x2771;   HEAVY RIGHT-POINTING ANGLE BRACKET ORNAMENT (Close_Punctuation)
	'❲'  U+2772  10098  e2 9d b2    &lbbrk;    LIGHT LEFT TORTOISE SHELL BRACKET ORNAMENT (Open_Punctuation)
	'❳'  U+2773  10099  e2 9d b3    &rbbrk;    LIGHT RIGHT TORTOISE SHELL BRACKET ORNAMENT (Close_Punctuation)
	'❴'  U+2774  10100  e2 9d b4    &#x2774;   MEDIUM LEFT CURLY BRACKET ORNAMENT (Open_Punctuation)
	'❵'  U+2775  10101  e2 9d b5    &#x2775;   MEDIUM RIGHT CURLY BRACKET ORNAMENT (Close_Punctuation)

But there are many parenthesis; I marked the allowed ones with here:

    % uni search 'LEFT PARENTHESIS'
    N    '('  U+0028  40     28          &lpar;     LEFT PARENTHESIS (Open_Punctuation)
    Y    '◌'  U+1AC1  6849   e1 ab 81    &#x1ac1;   COMBINING LEFT PARENTHESIS ABOVE LEFT (Nonspacing_Mark)
    Y    '◌'  U+1AC3  6851   e1 ab 83    &#x1ac3;   COMBINING LEFT PARENTHESIS BELOW LEFT (Nonspacing_Mark)
    Y    '⁽'  U+207D  8317   e2 81 bd    &#x207d;   SUPERSCRIPT LEFT PARENTHESIS (Open_Punctuation)
    Y    '₍'  U+208D  8333   e2 82 8d    &#x208d;   SUBSCRIPT LEFT PARENTHESIS (Open_Punctuation)
    Y    '⎛'  U+239B  9115   e2 8e 9b    &#x239b;   LEFT PARENTHESIS UPPER HOOK (Math_Symbol)
    Y    '⎜'  U+239C  9116   e2 8e 9c    &#x239c;   LEFT PARENTHESIS EXTENSION (Math_Symbol)
    Y    '⎝'  U+239D  9117   e2 8e 9d    &#x239d;   LEFT PARENTHESIS LOWER HOOK (Math_Symbol)
    N    '❨'  U+2768  10088  e2 9d a8    &#x2768;   MEDIUM LEFT PARENTHESIS ORNAMENT (Open_Punctuation)
    N    '❪'  U+276A  10090  e2 9d aa    &#x276a;   MEDIUM FLATTENED LEFT PARENTHESIS ORNAMENT (Open_Punctuation)
    Y    '⹙'  U+2E59  11865  e2 b9 99    &#x2e59;   TOP HALF LEFT PARENTHESIS (Open_Punctuation)
    Y    '⹛'  U+2E5B  11867  e2 b9 9b    &#x2e5b;   BOTTOM HALF LEFT PARENTHESIS (Open_Punctuation)
    Y    '﴾'  U+FD3E  64830  ef b4 be    &#xfd3e;   ORNATE LEFT PARENTHESIS (Close_Punctuation)
    Y    '︵' U+FE35  65077  ef b8 b5    &#xfe35;   PRESENTATION FORM FOR VERTICAL LEFT PARENTHESIS (Open_Punctuation)
    Y    '﹙' U+FE59  65113  ef b9 99    &#xfe59;   SMALL LEFT PARENTHESIS (Open_Punctuation)
    Y    '（' U+FF08  65288  ef bc 88    &#xff08;   FULLWIDTH LEFT PARENTHESIS (Open_Punctuation)
    Y    '�'  U+E0028 917544 f3 a0 80 a8 &#xe0028;  TAG LEFT PARENTHESIS (Format)

So basically all except these that are specifically excluded.

"You can have super- and subscript parens, small parens, and fullwidth parens, but not regular parens or ornamental parens" is unexplainable, just as "you can use this smiling emoji but not that other smiling emoji" is.

ChristianSi:
@arp242: It's actually quite simple to explain: "You can use arbitrary words in arbitrary languages as bare keys. If you want to use more than one word as a bare key, use dashes or underscores to connect them, as whitespace is not allowed. Better don't use other characters in bare keys, as that may or may not work."

That's in fact more or less how I explain it now in the README.

These additional non-letters are simply allowed in order to simplify the range definitions, it's not that anybody is supposed to use them. So there is no point in, and no need for, detailed explanations.

arp242:
> Better don't use other characters in bare keys, as that may or may not work.

I can "explain" everything with this.

People don't read specs, nor should they. They try stuff and see what happens. What you get now is that they try something, which works, and then they change it to something else and that doesn't work, which makes no sense.

ChristianSi:
Why should people try using something like LEFT PARENTHESIS UPPER HOOK in a key and then, when it works, move on to MEDIUM LEFT PARENTHESIS ORNAMENT, only to be disappointed that it doesn't work? That doesn't strike me as a very likely scenario. I think you're worrying too much.

ChristianSi:
@pradyunsg Are you around to take a look?

ChristianSi:
@pradyunsg Kind reminder that this is still open, and that maybe you need additional maintainers to support you?

ChristianSi:
@pradyunsg Ping?

abelbraaksma:
> And the pragmatic approach of simply defining ranges is preserved, instead of switching to an approach based on Unicode character classes that would make implementation considerably more complicated for those who don't have access to a fully-fledged Unicode support library.

I think this is the key improvement here. Allowing only certain character classes was never gonna work (different versions of Unicode issues, or, if specified, way too specific to be useful long-term in any practical way).

My opinions have already been discussed in the linked issue. The main point was this one: blowing up the spec for the sake of one single character class specification. While this change includes many characters that arguably shouldn't be part of a bare key, ultimately, that's up to users (and there are many languages, including certain .NET languages, that support all characters in type and member names, so there's precedent): if they want to write code that includes emojis or brackets, they can. 

The more pragmatic and welcome change here is to simply be inclusive of other scripts and languages (which was the aim of the original PR that extended bare keys). Anything else is just what fishermen call "by-catch" ;).

Great work for getting this ready!

toml.md
```diff
@@ -104,10 +104,15 @@ first = "Tom" last = "Preston-Werner" # INVALID
...
-e.g. 1234, but are always interpreted as strings.
+any Unicode script, as well as digits, dashes and underscores. Various symbols
+(such as emojis or arrows) and punctuation marks from outside the ASCII range
+are also allowed, while whitespace and private use characters are forbidden. It
+is advisable to use only words (in arbitrary languages) as bare keys, as they
+are generally accepted, while not all symbols and punctuation marks are. If you
+want to use a bare key made up of several words, use a suitable separator
+character (such as an underscore or hyphen) between the words, as spaces are not
+allowed. Note that bare keys are allowed to be composed of only digits, e.g.
+1234, but are always interpreted as strings.
```
abelbraaksma, line 115: I like this. We tried to come up with some advisary language in the earlier attempt to extend bare keys, but decided against it. I think it is good to have it in 👍.

abelbraaksma (approved):

PR #1058, opened 2025-04-14
Clarify integer size and float precision levels

### LINKED ISSUE #1029: Required int/float sizes implementations need to support?
The spec says:

> Arbitrary 64-bit signed integers (from −2^63 to 2^63−1) should be accepted and handled losslessly. If an integer cannot be represented losslessly, an error must be thrown.
>
> Floats should be implemented as IEEE 754 binary64 values.

"Should" can mean any number of things.

Does this mean "must" or "it's also okay to accept less"?

---

Practically speaking, int64 is annoying in JS as it only has floats, which is only accurate up to int54 or thereabouts. You can use BigInt and some implementations do this, but it's not very fluid.

This is also an issue for JSON compatibility; which like JS only has ints.

I'm not sure if there are other cases where int64 posses issues? But based on the JSON alone I'd say that it should be changed to "MUST accept int32, RECOMMEND accepting at least int64".

---

For float64 I'm not sure if there are environments or use-cases where it doesn't work? Maybe embedded use cases @marzer? Otherwise I'd be in favour of just requiring float64.

arp242:
"Should" is one of those ambiguous terms that can mean any number of things. We never strictly define what "should" means.

This clarifies things so that it's obvious that any size can be supported, but that 64bit sizes are recommended.

Fixes #1029

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@
 
 - Clarify sub-millisecond precision is allowed ([#805]).
 
+- Clarify that parsers are free to support any int or float size ([#1058]).
+
 [#790]: https://github.com/toml-lang/toml/pull/790
 [#796]: https://github.com/toml-lang/toml/pull/796
 [#805]: https://github.com/toml-lang/toml/pull/805
@@ -47,6 +49,7 @@
 [#904]: https://github.com/toml-lang/toml/pull/904
 [#929]: https://github.com/toml-lang/toml/pull/929
 [#950]: https://github.com/toml-lang/toml/pull/950
+[#1058]: https://github.com/toml-lang/toml/pull/1058
 
 ## 1.0.0 / 2021-01-11
 
--- a/toml.md
+++ b/toml.md
@@ -464,13 +464,12 @@ oct2 = 0o755 # useful for Unix file permissions
 bin1 = 0b11010110
 ```
 
-Arbitrary 64-bit signed integers (from −2^63 to 2^63−1) should be accepted and
-handled losslessly. If an integer cannot be represented losslessly, an error
-must be thrown.
+Implementations are free to support any integer size. It's recommended that at
+least 64-bit signed integers (from −2^63 to 2^63−1) are accepted and handled
+losslessly. If an integer cannot be represented losslessly, an error must be
+thrown.
 
 ## Float
-
-Floats should be implemented as IEEE 754 binary64 values.
 
 A float consists of an integer part (which follows the same rules as decimal
 integer values) followed by a fractional part and/or an exponent part. If both a
@@ -530,6 +529,9 @@ sf4 = nan  # actual sNaN/qNaN encoding is implementation-specific
 sf5 = +nan # same as `nan`
 sf6 = -nan # valid, actual encoding is implementation-specific
 ```
+
+Implementations are free to support any precision level. It's recommended that
+at least IEEE 754 binary64 values are supported.
 
 ## Boolean
 


toml.md
```diff
@@ -479,13 +479,17 @@ oct2 = 0o755 # useful for Unix file permissions
...
+least 64-bit signed integers (from −2^63 to 2^63−1) are accepted and handled
+losslessly. If an integer cannot be represented losslessly, an error must be
+thrown.
 
 ## Float
 
-Floats should be implemented as IEEE 754 binary64 values.
+Floating point numbers are expressed as IEEE 754 values. Implementations are
+free to support any float size. It's recommended that at least binary64 values
+are accepted and handled losslessly. If an floating point cannot be represented
```
pradyunsg, line 491:
```suggestion
are accepted and handled losslessly. If a floating point cannot be represented
```

Copy paste error? :P

pradyunsg (dismissed):
A couple of notes:

1. The current language permitted using arbitrary precision floating point values, which the language explicitly excludes with "are expressed as ...".

2. I'd suggest dropping the current first line and add the implementation notes at the end of the section, so that we don't have dense language at the start of the section. Relatedly, suggestion on the language:

  > Implementations are free to support any precision level. It's recommended that at IEEE 754 binary64 values) are accepted and handled losslessly. If a floating point cannot be represented losslessly, an error must be thrown.

pradyunsg:
> "Should" is one of those ambiguous terms that can mean any number of things. We never strictly define what "should" means.

FWIW, I have always read should as meaning "this expectation is to" (similar to recommended). I'm on-board with switching to recommended here since that's a clearer indicator of what the intention is here than "should".

pradyunsg:
> 2\. add the implementation notes at the end of the section

@arp242 I think you missed this bit from my review (unless you have another push lined up). 😅

pradyunsg:
And we should probably add a changelog entry as well.

arp242:
I kept in the reference to "IEEE 754" because that has a bunch of specific behaviour wrt. rounding and such, and it's what was already there. On the other hand, I know a bunch of implementations that *don't* use it already, so probably okay to drop that.

> I think you missed this bit from my review (unless you have another push lined up).

No, just WIP; wanted to see what it reads like when rendered (for some reason that always seemed different). But yes, at the end is better.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 - Seconds in Date-Time and Time values are now optional.
 - Allow non-English scripts in unquoted (bare) keys
 - Clarify newline normalization in multi-line literal strings.
+- Clarify that parser are free to support any int or float size (#1058).
```
pradyunsg, line 15: *parsers are

pradyunsg (commented):
LGTM, with a smol English-related note.

pradyunsg:
And... this needs another rebase. 😅

arp242:
Dine

ChristianSi:
I'm a bit confused by what's meant with the lossless representation of floats. Many decimals cannot be exactly represented as binary64 – say 0.1 will actually be converted into a very slightly different value. So must a TOML parser throw an error when encountering such decimals? That doesn't sound very practical! But then, what exactly ''is'' the intend of the new wording and when ''must'' an error be thrown?

arp242:
That was inspired partly by the similar language for ints, and partly by https://github.com/FourierTransformer/tinytoml/issues/6 where TOML like this:

    pi = 3.141592653589793

Would get round-tripped to:

    pi = 3.1415926535898

If we don't want to enforce any minimum size, then we should probably have *something* on what to do if a float doesn't fit the supported size, but not sure what the best way to phrase that would be. "Losslessly" is probably not the best word here 🤔

ChristianSi:
Yes, I'm afraid the "lossless" wording will not work for floats, which are inherently lossy. I also think the sentence "Implementations are free to support any precision level" does not make much sense, since it seems to think about the number of decimal digits (which are visible in the file), but inherently floats aren't decimal. And we need to decide whether we really want to allow *any* level of precision, no matter how low (probably not!) or rather enforce binary64 as the minimum. I think the latter makes much more hence. Hence we could replace the three sentences with something like:

> Implementations must accurately accept and represent all values conforming to the IEEE 754 binary64 format; they may support floating-point values of higher precision.

ChristianSi (changes requested):
The float-related language needs to be clarified, see comment.

toml.md
```diff
@@ -545,6 +544,10 @@ sf5 = +nan # same as `nan`
 sf6 = -nan # valid, actual encoding is implementation-specific
 ```
 
+Implementations are free to support any precision level. It's recommended that
+at least IEEE 754 binary64 values are accepted and handled losslessly. If a
+floating point cannot be represented losslessly, an error must be thrown.
+
```
pradyunsg, line 550:
```suggestion
Implementations are free to support any precision level. It's recommended that
at least IEEE 754 binary64 values are accepted and stored without losing bits.
If a floating point cannot be accepted and stored, an error must be thrown.
```

pradyunsg: (an attempt at addressing concerns from the review above)

ChristianSi: @pradyunsg Not sure that'll help – what does "losing bits" means? In the file, floats are actually stored as decimals, so there is no clear "bit" equivalent, and conversion into a float representation will in most cases be slightly lossy. Also, how to decide which values "cannot be accepted and stored"?

arp242:
I changed it to *"If a floating point cannot be represented losslessly beyond imprecision inherent to the format, an error must be thrown."*

I spent some time reading up on IEEE 754 and I couldn't really find any better language. This should probably be fairly clear to most readers?

ChristianSi:
@arp242 It's not clear to me and I think something similar to what I proposed above (https://github.com/toml-lang/toml/pull/1058#issuecomment-2823653871) would make more sense.

First, we now say "Implementations are free to support any precision level", which means that an implementation that just supports 32-bit floats (Java/C *float*, Go *float32*) or even 16-bit floats (*numpy.float16*) would be acceptable (if not recommended). But do we really want that?

Also what does "If a floating point cannot be represented losslessly beyond imprecision inherent to the format" mean? I must admit I don't have the foggiest idea. Which floats would a 32-bit implementation have to reject? Can you give an example? And which floats would an IEEE 754 binary64-compatible implementation have to reject – are there any?

arp242:
> First, we now say "Implementations are free to support any precision level", which means that an implementation that just supports 32-bit floats (Java/C float, Go float32) or even 16-bit floats (numpy.float16) would be acceptable (if not recommended). But do we really want that?

That's what came out of #1029

arp242:
Well, I ended up just removing that sentence. I can't really think of a better way to phrase it, and it wasn't really what I wanted to address here either.

toml.md
```diff
@@ -545,6 +544,9 @@ sf5 = +nan # same as `nan`
 sf6 = -nan # valid, actual encoding is implementation-specific
 ```
 
+Implementations are free to support any precision level. It's recommended that
+at least IEEE 754 binary64 values are accepted.
```
ChristianSi, line 548: I would add the "and handled losslessly" from above here too. Otherwise the "accepted" would be pretty meaningless, since it could be interpreted as "accepted, but internally converted to a lower precision".

ChristianSi:
This is starting to look pretty good, though I have suggested a small wording change.

Personally I would still favor making 64-bit int/float support a strict requirement, as I had [argued for](https://github.com/toml-lang/toml/issues/1029#issuecomment-2119881302) in #1029. Every modern implementation should manage that, especially maybe on tiny embedded systems, which are unlikely to use TOML anyway. But if we don't, this PR is now pretty decent.

toml.md
arp242: But then we're back to your [original comment](https://github.com/toml-lang/toml/pull/1058#issuecomment-2823653871) that "handled losslessly" doesn't really work for IEEE floats?

ChristianSi:
@arp242 All right, let's stick with "recommended" for now.

toml.md
ChristianSi: Hmm, well, the problem already lies in the "accepted". What's accepted, i.e. parsed, is originally never a float, but rather a decimal written as string. So I guess what we mean is not that they are "accepted", but they there are used internally, meaning that any decimals will be parsed with sufficient precision.

So how about changing it to "It's recommended that at least IEEE 754 binary64 values are *supported*" ? That should do the trick, I think.

ChristianSi: @arp242 So what do you think about changing "accepted" to "supported" to finalize this?

ChristianSi (approved):
Looks fine now!

ChristianSi:
@pradyunsg Can you take another look and ideally approve this? I think it's now the last issue that's blocking TOML 1.1 from being released :flushed:

ChristianSi:
@arp242  What do you think about merging this? Let's do this!

(I might do it in a week if nobody objects until then, but feel free to do it now.)

ChristianSi:
I have merged it now, since the PR is fine and there were no objections.

PR #1059, opened 2025-04-19
Expand changelog

arp242:
One of my difficulties in updating TOML implementations in the past was figuring out what exactly I need to change, as the changelog has typically been extremely concise.

This expands things with commit references and some examples for changes in syntax, which should make things a lot easier for people looking to update TOML 1.0 implementations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,16 +2,51 @@
 
 ## unreleased
 
-- Clarify that comments never affect the tables produced by parsers.
-- Clarify Unicode and UTF-8 references.
-- Allow newline after key/values in inline tables.
-- Allow trailing comma in inline tables.
-- Clarify where and how dotted keys define tables.
-- Add new `\e` shorthand for the escape character.
-- Add \x00 notation to basic strings.
-- Seconds in Date-Time and Time values are now optional.
-- Allow non-English scripts in unquoted (bare) keys
-- Clarify newline normalization in multi-line literal strings.
+- Allow newlines and trailing commas in inline tables ([#904]).
+
+  Previously an inline table had to be on a single line and couldn't end with a
+  trailing comma. This is now relaxed so that the following is valid:
+
+      tbl = {
+          key      = "a string",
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,16 +2,41 @@
...
-- Clarify where and how dotted keys define tables.
-- Add new `\e` shorthand for the escape character.
-- Add \x00 notation to basic strings.
-- Seconds in Date-Time and Time values are now optional.
-- Allow non-English scripts in unquoted (bare) keys
-- Clarify newline normalization in multi-line literal strings.
+- Allow newlines and trailing commas in inline tables (18348af).
+
+  Previously an inline table had to be on a single line and couldn't end with a
+  trailing comma. This is now relaxed so that this is valid:
```
pradyunsg, line 8:
```suggestion
  trailing comma. This is now relaxed so that the following is valid:
```

CHANGELOG.md
```diff
@@ -2,16 +2,41 @@
...
+- Add `\x00` notation to basic strings for codepoints <255 (7e8a748):
+
+      null = "null byte: \x00"
+
+- Add `\e` escape for the escape character (e56ab35):
+
+      csi = "\e["
+
+- Seconds in datetime and time values are now optional (7e8a748). These are now
+  valid:
```
pradyunsg, line 26:
```suggestion
- Seconds in datetime and time values are now optional (7e8a748). The
  following are now valid:
```

pradyunsg (commented):
I haven't double-checked the hashes, and I think it would be better to use the relevant issue/PR numbers instead (as well as include a link for that).

arp242:
> I haven't double-checked the hashes, and I think it would be better to use the relevant issue/PR numbers instead (as well as include a link for that).


I think there was a reason I did it like that, but I don't recall what it was (I originally did this years ago, but never bothered submitting a PR).

pradyunsg (approved):

PR #1063, opened 2025-06-04
Remove spurious line referring to code points

ChristianSi:
It's a leftover from the backed-out attempt to introduce Unicode bare keys and so completes #979.

@arp242, @pradyunsg : Can you take a look and approve?


arp242:
Cheers

arp242:
I think your PR at #1002 is now also redundant by the way?

PR #1064, opened 2025-06-08
Clarify string descriptions

arp242:
This came out of seeing what (if anything) we want to merge out of #875; it makes the following copy editing changes:

- Clearly list allowed codepoints at the start of every string type.

- "Unicode" on its own doesn't necessarily mean anything; UTF-16 or UCS-2 is "Unicode". Perhaps a bit pedantic, but "UTF-8" or "codepoints" are "more correct".

  Similarly, a "character" or "Unicode character" is quite a tricky thing to define. Multiple codepoints can be one "character". Most of the time "codepoint" is really what's intended.

- Don't link to some random page for base64 decode. Guess we could link to Wikipedia, but seems a but redundant to me.

Fixes #875

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -260,12 +260,11 @@ The above TOML maps to the following JSON.
 ## String
 
 There are four ways to express strings: basic, multi-line basic, literal, and
-multi-line literal. All strings must contain only Unicode characters.
+multi-line literal. All strings must be encoded as UTF-8.
 
-**Basic strings** are surrounded by quotation marks (`"`). Any Unicode character
-may be used except those that must be escaped: quotation mark, backslash, and
-the control characters other than tab (U+0000 to U+0008, U+000A to U+001F,
-U+007F).
+**Basic strings** are surrounded by quotation marks (`"`). Any codepoint may be
+used except those that must be escaped: quotation mark, backslash, and the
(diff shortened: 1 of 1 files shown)


ChristianSi:
@arp242 , @pradyunsg What do you think?

ChristianSi:
@arp242 How do you want to proceed here? I think my suggestions for improvement make sense, don't you agree?

If we have trouble finding an unified version here, we could also consider just closing this. The current version is basically already fine and there is no urgent need to change anything.

And this now the last issue blocking TOML 1.1, and there is certainly on good reason to allow such a minor matter, which doesn't even change anything in the language, to block the next release, which is already so much overdue!

ChristianSi:
OK, if no one suggests another course of action, I'll close this in a week or so. For 1.1 the used language should be fine, and if it turns out helpful, we can always revise it for 1.2 or later.

arp242:
> OK, if no one suggests another course of action, I'll close this in a week or so.

Yes that's okay; I mostly wanted to include anything from #875 worth keeping, but it's not a big deal, and the changes are too minor and unimportant to spend my time discussing to be honest.

(Sorry for very late reply, I was busy with other stuff earlier this year)

PR #1075, opened 2025-12-19
remove redundant time precision requirements

andrius-puksta-sensmetry:
Removed redundant statements about millisecond precision, erroneously introduced in https://github.com/toml-lang/toml/pull/805


arp242:
Thanks; I'm not sure if it was always wrong or if I messed up with the rebase + merge of an old PR. But it's pretty clear the intent was to *replace* that paragraph all three times rather than add a new one.

ChristianSi:
Thanks for handling this, @arp242 !

PR #1086, opened 2026-04-30
Fix duplicate keys in doc introduced in 1.1

sjorsdonkers:
1.1 introduced optional seconds for time and datetimes. The examples that were added are nicely updated to be sequential within the code block, but caused keys to duplicate with following code blocks. This is inconsistent with the rest of the document which avoids duplicated keys. Also it just looks odd to randomly have a odt4 again.

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -558,15 +558,15 @@ For the sake of readability, you may replace the T delimiter between date and
 time with a space character (as permitted by RFC 3339 section 5.6).
 
 ```toml
-odt4 = 1979-05-27 07:32:00Z
+odt5 = 1979-05-27 07:32:00Z
 ```
 
 One exception to RFC 3339 is permitted: seconds may be omitted, in which case
 `:00` will be assumed. The offset immediately follows the minutes.
 
 ```toml
-odt5 = 1979-05-27 07:32Z
-odt6 = 1979-05-27 07:32-07:00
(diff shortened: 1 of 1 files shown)


arp242:
I think I'd prefer to outright remove the numeric continuity between examples. That is:

	```toml
	odt = 1979-05-27 07:32:00Z
	```

	[.. text ..]


	```toml
	odt1 = 1979-05-27 07:32Z
	odt2 = 1979-05-27 07:32-07:00
	odt3 = 1979-05-27 07:32Z
	odt4 = 1979-05-27 07:32-07:00
	```

Unless there's some value in keeping it? Seems like a pain to maintain and easy to break.

I don't mind doing that (boring) change myself – don't expect you to change it all.

sjorsdonkers:
Feel free to take it any direction you prefer.

Your point regarding maintainability are valid and evidently real.

The smallest argument in favor of keeping unique variable names across blocks is that you can currently test the correctness of all examples in the docs, by putting them in a single file, like [datetime examples](https://github.com/sjorsdonkers/toml-jai/blob/main/examples/data/datetime.toml). One could obviously just create a file per block. Some blocks in the doc contain purposefully invalid examples, so there is very little to this argument.

arp242:
Okay, lets just do this as it's the smallest change for now.

Issue #432, opened 2016-09-16
What's the point of empty quoted keys?


Lucretia:
> A bare key must be non-empty, but an empty quoted key is allowed (though discouraged).

Seems completely pointless to allow this, why not just disallow it?

mk-pmb:
I'm sure there are a lot more keys that you'd deem pointless. How about keys consisting of just space characters? We should consider a whitelist approach here, and only allow keys that our chosen wise experts deem valuable.

BurntSushi:
@mk-pmb I'd appreciate it if you could be more constructive in future comments. As it stands, your comment isn't very helpful.

In any case, disallowing empty keys would be a breaking change, and I don't think we should be making any breaking changes at this point.

Lucretia:
@BurntSushi didn't answer the question. I wanted to know what they are useful for? I cannot see a point to this.

BurntSushi:
I don't have a good answer for you. If I did, I would have given you one.

mk-pmb:
Use case example: In [xmlattrdict](https://www.npmjs.com/package/xmlattrdict) I use the empty string key for the tag name. Ideally I'd use `null` as a key, to indicate that the entry is in the dict just for convenience even though it's not really a key-value-pair like the attributes. I can't use actuall null in ES5, so the empty string was the closest equivalent I could find.

Lucretia:
@mk-pmb thanks, think  i'll experiment with one of the libs.

PR #1095, opened 2026-06-29
Clarify the reason empty keys are allowed

arp242:
TOML allows empty (quoted) keys:

	"" = 123

Over the years I've seen a number of people express surprise that it's allowed.

The reason it's allowed is because TOML tables are intended to map to hash tables/dictionaries in programming languages, and they allow empty keys. So TOML does too.

You can still argue that it's a mistake to allow it in TOML, but we can't change that now and this at least clarifies the thinking behind why it's allowed.

### DIFF
--- a/toml.md
+++ b/toml.md
@@ -129,8 +129,11 @@ to use bare keys except when absolutely necessary.
 'quoted "value"' = "value"
 ```
 
-A bare key must be non-empty, but an empty quoted key is allowed (though
-discouraged). You cannot use multi-line strings to define quoted keys.
+You cannot use multi-line strings to define quoted keys.
+
+TOML tables correspond to hash tables (also known as dictionaries) which usually
+permit any string value. For that reason empty quoted keys are allowed, though
+discouraged. A bare key must be non-empty.
 
 ```toml
 = "no key name"           # INVALID
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 ## unreleased
 
-- Clarify that tables typically cannot be nested infinitely [#1087]).
+- Clarify that tables typically cannot be nested infinitely ([#1087]).
+
+- Clarify the reason for allowing empty keys ([#1095]).
 
 [#1087]: https://github.com/toml-lang/toml/pull/1087
+[#1095]: https://github.com/toml-lang/toml/pull/1095
 
 ## 1.1.0 / 2025-12-18
 


toml.md
```diff
@@ -129,8 +129,11 @@ to use bare keys except when absolutely necessary.
 'quoted "value"' = "value"
 ```
 
-A bare key must be non-empty, but an empty quoted key is allowed (though
-discouraged). You cannot use multi-line strings to define quoted keys.
+You cannot use multi-line strings to define quoted keys.
+
+TOML tables are intended to map to hash tables or dictionaries, as a consequence
+empty quoted keys are allowed (though discouraged). A bare key must be
```
ChristianSi: The rationale feels a bit underspecified. Many languages and data structures that have hash tables or dictionaries also allow numbers, booleans, and other non-string values as keys, but TOML keys are intentionally more restricted than that. So, to prevent further possible confusion, I’d suggest making the rationale a bit more explicit, perhaps along these lines:

> TOML tables are intended to map to hash tables or dictionaries, which commonly permit empty string keys. For that reason, empty quoted keys are allowed, though discouraged. A bare key must be non-empty.

ChristianSi (changes requested):
This may ultimately be better suited for a FAQ or rationale page, but since TOML does not yet have one, a short explanation in the spec itself seems appropriate enough. However, I’d be a bit more explicit about the underlying reasoning; see my suggested change.

toml.md
arp242: How about:

> TOML tables correspond to hash tables or dictionaries which usually permit any string value. For that reason empty quoted keys are allowed, though discouraged. A bare key must be non-empty.

"intended to map to" → "correspond to" as it seems a bit less of a mouthful.

"commonly permit empty string keys" → "usually permit any string value" as that's a bit more general.

ChristianSi: @arp242 Sure, sounds good!

arp242: Okay, changed it with `hash tables or dictionaries` → `hash tables (also known as dictionaries)` because on re-reading the "or" could be interpreted to mean "one or the other" rather than "also known as".

cyyynthia:
> This may ultimately be better suited for a FAQ or rationale page, but since TOML does not yet have one, a short explanation in the spec itself seems appropriate enough

I also agree that the rationale is purely informational and I have mixed feelings about it being part of the spec document. It's at the same time valuable info, and at the same time noise that doesn't really have normative value. Perhaps a `<details>` block would be the best of both worlds? 🤔

ChristianSi:
@cyyynthia: For just one sentence added, a `<details>` block seems overkill, though.

ChristianSi (approved):

arp242:
Personally I think the "why" of these sort of things should be part of the main specification text. Over the years I've found that knowing *why* something is the way it is hugely helps actually understanding and remembering things. Otherwise it just become rote memorisation and I kind of suck at that.

Within reason, of course. A potential FAQ could expand on some of the things in the specification. But I do think it's helpful to explain the *why*, when it might be useful, as concise as possible, regardless of whether we have a FAQ or not.

I would certainly not be in favour of footnotes or `<detail>` tags. I avoid that in my own writing and generally dislike it as everyone reads them anyway and they give a disjointed reading experience. The exception being Sir Terry's footnotes because they're just so funny (I'd be okay with footnotes if they're sufficiently entertaining).

dimbleby:
> The reason it's allowed is because TOML tables are intended to map to hash tables/dictionaries in programming languages, and they allow empty keys. So TOML does too.

Is there evidence for this as an explanation?  Or is it a story we are retro-fitting ten years later?

If it were a good explanation, wouldn't non-string keys also be allowed?

The relevant phrase was I think introduced at https://github.com/toml-lang/toml/pull/387,  which nominally was just moving text around.  I don't find any discussion of whether or not to allow empty strings - perhaps something exists elsewhere? - so it is hard at this distance to say what the reasoning was.

FWIW: no explanation was offered at https://github.com/toml-lang/toml/issues/432, which was nearly contemporary.

arp242:
It's in the text already: "Tables (also known as hash tables or dictionaries)". Older versions are even more explicit: "TOML is designed to map unambiguously to a hash table".

dimbleby:
The question was not: is TOML designed to map to a hash table?  I agree that it is.  But this would hold _regardless_ of whether empty strings were allowed as keys.

The question was more: is this either a historically accurate or currently useful explanation for why empty strings should be allowed as keys?  I find it unnecessarily involved, eg as both I and - I now see - ChristianSi have experienced, talk of hash tables raises questions in reader minds about non-string keys.

To my mind a simpler explanation is just: empty strings are allowed because an empty string is a perfectly good string and there was no compelling reason to make an exception for it.

Whether that deserves a sentence in the spec is unclear to me.

ChristianSi:
@dimbleby: I agree – that’s a perfectly plausible explanation, and "TOML keys are conceptually strings" is a simpler way to put it than "TOML tables are conceptually hash tables whose keys are arbitrary strings." Both arrive at the same endpoint, but the detour through hash tables seems unnecessary.

PR #1103, opened 2026-08-28
Clarify Array of Tables do not need to be grouped together

### LINKED ISSUE #1100: Clarify if array of table entries must be grouped together
It is not clear from the spec if the values in an array of tables must be grouped together in a file or if they can be interleaved. Specifically, is the following syntax allowed:

```TOML
[[alpha]]
val = 1
[[alpha]]
val = 2
[[beta]]
val = 11
[[beta]]
val = 22
[[alpha]]
val= 3
[[beta]]
val = 33
```

where the intended JSON equivalent is:

```JSON
{
  "alpha": [
    {
      "val": 1
    },
    {
      "val": 2
    },
    {
      "val": 3
    }
  ],
  "beta": [
    {
      "val": 11
    },
    {
      "val": 22
    },
    {
      "val": 33
    }
  ]
}
```

Or, is it required that all the elements be together like the following:

```TOML
[[alpha]]
val = 1
[[alpha]]
val = 2
[[alpha]]
val= 3
[[beta]]
val = 11
[[beta]]
val = 22
[[beta]]
val = 33
```

arp242:
There was already such a paragraph for tables, so add one for array of tables as well.

In addition, remove the IMHO needless "The last syntax that has not yet been described allows" throat-clearing on the section, and instead of clarifying what [[..]] does with JSON,  use TOML inline arrays and tables. That removes the need for "you can also use inline arrays" section.

Fixes #1100

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,13 @@
 - Clarify that comparison for quoted keys is ordinal and doesn't do Unicode
   normalization ([#993])
 
-- Clarify that tables typically cannot be nested infinitely [#1087]).
+- Clarify that tables typically cannot be nested infinitely ([#1087]).
+
+- Clarify that arrays of tables do not need to be grouped together ([#1103]).
 
 [#993]: https://github.com/toml-lang/toml/pull/993
 [#1087]: https://github.com/toml-lang/toml/pull/1087
+[#1103]: https://github.com/toml-lang/toml/pull/1103
 
 ## 1.1.0 / 2025-12-18
 
--- a/toml.md
+++ b/toml.md
@@ -909,11 +909,11 @@ type.name = "Nail"
 
 ## Array of Tables
 
-The last syntax that has not yet been described allows writing arrays of tables.
-These can be expressed by using a header with a name in double brackets. The
-first instance of that header defines the array and its first table element, and
-each subsequent instance creates and defines a new table element in that array.
-The tables are inserted into the array in the order encountered.
+Arrays of tables can be expressed by using a header with a name in double
+brackets. The first instance of that header defines the array and its first
+table element, and each subsequent instance creates and defines a new table
+element in that array. The tables are inserted into the array in the order
+encountered.
 
 ```toml
 [[product]]
@@ -929,16 +929,14 @@ sku = 284758393
 color = "gray"
 ```
 
-In JSON land, that would give you the following structure.
+You can also use inline tables; the following is identical:
 
-```json
-{
-  "product": [
-    { "name": "Hammer", "sku": 738594937 },
+```toml
+product = [
+    { name = "Hammer", sku = 738594937 },
     {},
-    { "name": "Nail", "sku": 284758393, "color": "gray" }
-  ]
-}
+    { name = "Nail",   sku = 284758393, color = "gray" },
+]
 ```
 
 Any reference to an array of tables points to the most recently defined table
@@ -967,7 +965,7 @@ name = "banana"
 name = "plantain"
 ```
 
-The above TOML maps to the following JSON.
+The above TOML maps to the following JSON:
 
 ```json
 {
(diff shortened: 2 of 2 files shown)


toml.md
```diff
@@ -1024,12 +1022,34 @@ shape = "round"
 color = "green"
 ```
 
-You may also use inline tables where appropriate:
+Defining arrays of tables out-of-order is discouraged.
```
ChristianSi, line 1039: "Discouraged" implicitly allows it, but I think an explicit clarification would be even better. Hence I'd suggest changing this to something like:

> Instances of the same array of tables need not be grouped together, but it is recommended to do so whenever possible.

This would also avoid the "out-of-order" wording, which could be considered confusing here. (Arrays of tables are always ordered, so they can't be defined "out-of-order" in the same way as, say, child tables can be defined before their parents.)

toml.md
```diff
@@ -1024,12 +1022,34 @@ shape = "round"
...
+```toml
+# RECOMMENDED
+[store]
+name = "Fruit 'n Veg Emporium"
+[[fruit]]
+name = "apple"
+[[fruit]]
+name = "pear"
+[[vegetable]]
+name = "broccoli"
```
ChristianSi, line 1071: Change to "courgette" to use the same order as above.

toml.md
```diff
@@ -1024,12 +1022,34 @@ shape = "round"
...
+[[vegetable]]
+name = "courgette"
```
ChristianSi, line 1074: Change to "broccoli" to use the same order as above.

toml.md
```diff
@@ -915,16 +915,14 @@ sku = 284758393
 color = "gray"
 ```
 
-In JSON land, that would give you the following structure.
+You may also use inline tables; the following is identical:
```
ChristianSi: I'd consider changing "the following is identical" to something like "the following represents the same data". The reason is that they are identical when the given examples are complete TOML files, but not if there is other data before or after them.

(If there's a table header before it, `product =` would be placed in it, while `[[product]]` is always a top-level array of tables. And the `[[product]]` syntax allows additional `[[product]]` headers after the three given ones, while the `product =` syntax doesn't.)

toml.md
```diff
@@ -915,16 +915,14 @@ sku = 284758393
...
 
-```json
-{
-  "product": [
-    { "name": "Hammer", "sku": 738594937 },
+```toml
+product = [
+    {name = "Hammer", sku = 738594937},
```
ChristianSi: I'd consider adding spaces within the braces, as that's the style used elsewhere for inline tables:

> { name = "Hammer", sku = 738594937 },

(And likewise for the third entry below.)

ChristianSi (changes requested):
Looks good overall, I'd just change some details for clarity and consistency.

CHANGELOG.md
```diff
@@ -2,9 +2,12 @@
 
 ## unreleased
 
-- Clarify that tables typically cannot be nested infinitely [#1087]).
+- Clarify that tables typically cannot be nested infinitely ([#1087]).
+
+- Clarify Array of Tables do not need to be grouped together ([#1103]).
```
ChristianSi: Better "Clarify that arrays of tables ..."

toml.md
```diff
@@ -1024,12 +1022,34 @@ shape = "round"
...
+Defining arrays of tables out-of-order is discouraged.
 
 ```toml
-points = [ { x = 1, y = 2, z = 3 },
-           { x = 7, y = 8, z = 9 },
-           { x = 2, y = 4, z = 8 } ]
+# VALID BUT DISCOURAGED
+[[fruit]]
+name = "apple"
+[store]
```
ChristianSi, line 1046: I'd consider adding an empty line before each table and array-of-table header (throughout both examples), as that's the style used elsewhere.

ChristianSi:
After we merge this, it would also be good to add some valid test cases to toml-test. The "VALID BUT DISCOURAGED" example added here could be one; additional tests could be taken from https://github.com/toml-lang/toml/issues/1100#issue-5185516777 and https://github.com/toml-lang/toml/issues/1100#issuecomment-5384206145.

toml.md
arp242: I copied this line from tables and dotted keys ("Defining tables out-of-order is discouraged." and "Defining dotted keys out-of-order is discouraged.")

Personally I prefer the simpler shorter language over your proposal, and with the example I think it's fairly clear.

toml.md
arp242: I rather prefer the simpler language of "is identical" over "represents the same data" here too. I feel readers can understand that (obviously) there are differences between the syntax structures, and that in some cases one might prefer one or the other.

CHANGELOG.md
```diff
@@ -5,10 +5,13 @@
 - Clarify that comparison for quoted keys is ordinal and doesn't do Unicode
   normalization ([#993])
 
-- Clarify that tables typically cannot be nested infinitely [#1087]).
+- Clarify that tables typically cannot be nested infinitely ([#1087]).
+
+- Clarify that Array of Tables do not need to be grouped together ([#1103]).
```
ChristianSi: You forgot to change this to lower-case and pluralized "arrays of tables". Or is there a reason for the singular capitalized form I don't get?

toml.md
ChristianSi: All right. Let's just hope this won't lead to further confusion in the future.

ChristianSi (changes requested):
I'd still suggest a tiny change to the CHANGELOG. Otherwise this is good to go!

CHANGELOG.md
arp242: Just as a general "capitalise terms with specific meaning". I don't feel strongly one way or the other – I thought your previous comments was just about adding "that".

qpj675mn8w-gif (commented):
###

ChristianSi (approved):
Looks fine now!

Issue #427, opened 2016-08-20
Suggestion: Add support for SI and IEC binary number suffixes


jdfergason:
I'm breaking this out from ticket #292 as I think it's a very useful feature that would be highly valued in scientific applications.

**Proposal**

For floating point and integer numbers allow SI and binary suffix modifiers.  This suffix acts as a multiplier on the base value.  The following table lists the supported suffixes.

| Decimal |  | Binary |  |
| --- | --- | --- | --- |
| **Suffix** | **Value** | **Suffix** | **Value** |
| k | 1000 | Ki | 1024 |
| M | 1000<sup>2</sup> | Mi | 1024<sup>2</sup> |
| G | 1000<sup>3</sup> | Gi | 1024<sup>3</sup> |
| T | 1000<sup>4</sup> | Ti | 1024<sup>4</sup> |
| P | 1000<sup>5</sup> | Pi | 1024<sup>5</sup> |
| E | 1000<sup>6</sup> | Ei | 1024<sup>6</sup> |
| Z | 1000<sup>7</sup> | Zi | 1024<sup>7</sup> |
| Y | 1000<sup>8</sup> | Yi | 1024<sup>8</sup> |

**Examples**

```
5k = 5000
10.3Mi = 10,800,332.8
```

**Motivation**

This would be very useful in scientific applications where this notation is common.  However, it is also useful outside of the scientific community, for example when specifying the maximum disk space to allow.

rmunn:
If this suggestion is adopted, I would also suggest allowing uppercase K (in addition to lowercase k) for 1000. Although uppercase K is not an official SI unit, if uppercase K is *not* allowed then it will cause confusion. All the *other* suffixes follow the pattern "remove the lowercase i from the binary suffix and you get the decimal suffix". If Ki -> K is allowed, then that pattern holds at all times and there will be less user confusion in the long run.

rmunn:
Also, the `10.3Mi = 10,800,332.8` example surprises me. I would expect `10.3Mi` to produce an int, not a float. More generally, I have never come across a situation where I wanted to express a *floating-point* value using binary suffixes. The rule in my head is "If there's a binary suffix, it's an int". Therefore, I would expect `10.3Mi` to be "the integer value closest to 10.3 * 1024<sup>2</sup>", or possibly "10.3 * 1024<sup>2</sup> rounded down to an int". (I don't know if `round` semantics or `floor` semantics would be least surprising to others.)

[...]

Issue #292, opened 2015-01-24
Few ideas


phaux:
Let me begin with a little story: About half a year ago I was inspired by TOML to make another simple language for storing structured data. I modified an existing TOML parser and even wrote a spec, but never actually used it... Until now. Recently I decided to reconsider using TOML, so I came here and reread the spec. It still doesn't have all the features that I need for more complicated structures, but it's much closer to my needs. So close, actually, that I decided to make my language a _strict superset of TOML_. Then I thought to myself: Why not just try to propose these changes to TOML itself? And here I am.

You can play with [this web-based TOML<->JSON converter](http://peg.arcanis.fr/S0SUi/2/) to test all of these features (multi-line strings and dates not yet implemented).
But first, take look at this:

```
# part of actual config for my RESTful API generator
[user.property.id]
type = str
validate.match = /[0-9A-Z_a-z]+/
normalize.strip_whitespace

[user.property.email]
type = email

[user.property.avatar]
type = picture
validate.filesize.max = 5M
normalize.transcode.small = { w=32, h=32, type="png" }
normalize.transcode.big =   { w=64, h=64, type="png", no_zoom }
```

Here's a list of things that I'd like to add (ordered by importance, most needed first):
1.  **Allow keys in key-value pairs to actually be paths** like `a.b.c = 42`. Very useful when you need to define an object that just happens to have only one or two properties.
   Of course you could just use underscores (`validate_filesize_max`) and split it on the application side, but I think it should be handled by the parser, so the example application can just take all the properties of the `validate` object (which is usually only one) and interpret them as it wishes.
2.  **Inline Objects.** Seriously, we need this. It fits in the language definition: It's "obvious": People are already familiar with such notation. It definately is minimal: adding this to an existing TOML parser written in PEG took me literally 5 seconds, yeah, I was surprised as well.
   There are situations (like `user.property.avatar.normalize.transcode` in the example), that just can't be handled in TOML without repeating tons of variable names.
3.  **First-class regex.** Do I need to justify that? Would be very cool to have. Could go between `/`s and "bare regex" would just be a full regex beginning with `^` and ending with `$`
4.  **Allow bare string as values** to make keys and strings more consistent. Not sure about multi-line strings as keys, tho. That would be silly, but would basically unify definition of key and string.
[...]

PR #1105, opened 2026-09-15
Add number suffixes for durations and bytesizes

### LINKED ISSUE #514: Feature request: Add a duration/timedelta type

### LINKED ISSUE #912: Add Filesize
It would be nice to have Filesizes in TOML.
```
test = 1MB # Results in 1.000 * 1.000 = 1.000.000
test2 = 1MiB # Results in 1.024 * 1.024 = 1.048.576
```
Things like max Filesize etc. are used in many configurations and it should be easy to implement for the different parsers.

arp242:
This adds the concept of "number suffixes" for duration and bytesizes. The problem this aims to solve is that:

1. Calculating these numbers is non-trivial; you can't "just" add zeros to turn seconds in to minutes, or a KB in to a MB. You need to do some arithmetic.

2. In many cases, it's not really clear what unit is being used, with multiple reasonable interpretations:

	    cache-size = 128  # Bytes? KB? MB?
	    timeout    = 100  # Milliseconds? Seconds?

   This can be solved with an awkward key name (cache-size-kb, timeout-seconds) or an almost mandatory comment. Neither are great.

It adds some complexity but makes TOML more "obvious". Quite a lot of configuration files have this built-in (see below), and in cases where it's not it's often inconvenient to *not* have it.

I made one proposal for both durations and bytesizes as they're essentially the same feature: adding a suffix to a number, and generalized rules for both would be nice.

I don't expect we'll ever want to add any other suffix types. They're rarely used in the context of TOML, in most cases it's also decimal and a matter of "just add zeroes". Just regular groupings are fine:

	weight          = 80        # Rarely used in TOML, "80 kg" or "80 lbs" might be clearer but is not needed.
	max_connections = 1_000_000 # Clearly a million, no need for generalised SI suffixes.

Some notes
----------
- You can only use one suffix; especially for durations "1 hour 30 minutes" or "1h30m" might be useful sometimes but isn't supported. This makes the implementation a bit easier and gives parity with bytesize suffixes. There was a bit of discussion on this in the issue. I'm not opposed to a "1 hour 30 minute" format and it does read a bit more natural, but it's a bit of a special case for parsers so I'd lean towards using fractions.

  It's possible for this to be changed/extended in a future version if it does prove to be too awkward: nothing in this proposal is preventing that.

- I disallowed inf and nan because there is no functional difference between "inf", "inf days", "inf minutes". Infinity is infinity. And this just looks horrible and is annoying to parse:

      inf-day = infday
      nan-day = nanday

- Negative values don't seem especially useful, but are allowed.

  I've gone back and forth on this: in my original write-up from 2023 it was allowed "because why not", then when updating it last week I forbade it together with inf/nan because it seemed a bit nonsensical, but then when writing the ABNF I allowed it again as it would be annoying to rewrite the ABNF.

  Don't necessarily see much harm in allowing it, and probably makes some implementations a bit easier since existing number parsing routines can be used without modifications (except for inf/nan).

- Unit names are currently always case-sensitive. Most other keywords are case-sensitive, and it avoids ambiguities in some cases: is 1M a minute or megabyte? 1m is more obviously a minute.

Notes on bytesizes
------------------
- Sizes are always as powers of 2 (times 1024, not 1000). This is what pretty much everyone else does with "1MB" or "1M". An explicit "KiB" is just confusing, and "MB is 1024 and M is 1000" even more so (especially with "m" also meaning "minute"). If you want base-10 units then use digit groupings.

  In practice almost no one wants base-10 units.

- Fractions are disallowed for bytes because "half a byte" doesn't really mean anything in most context and is nonsensical. A byte is almost always the smallest addressable size.

Notes on durations
------------------
- DST and leap «anything» are not an issue; this is something that should be handled in code when adding or subtracting a duration combined with a datetime point.

- The smallest durations is milliseconds, rather than nanoseconds. There are two reasons for that: 1) at least Python and JavaScript have millisecond as the lowest time unit in their stdlib, and 2) a float64 in ms can accurately hold all reasonable natural numbers: up to 285k years, rather than only 104 days with ns. This makes implementing a bit easier, as you can "just" use a float64 even for cache-time = 52 weeks. For this reason it also disallows fractions for milliseconds.

  I can't recall ever using ns in config files anyway. Note that nothing is preventing us from adding it later if there really is a demand.

- Originally I also allowed longer suffixes such as "day", "hour", etc:

      one-day       = 1 day
      two-days      = 2 days
      two-days-also = 2 day

  I ended up removing that, as I felt "more than one way to write something" is generally a bad idea. And most other config files don't support it.

  I do still rather like how natural "1 day" reads though. So idk. But I'd rather opt to start with the least possible, and extend later if need be.

Implementation notes and compatibility
--------------------------------------
Durations should usually parse to some duration type. We already have datetimes and practically anything that supports datetime also supports some notion of durations.

So if someone makes an error like so:

	timeout = 60s  # Correct use: parses to duration type.
	timeout = 60   # Incorrect use: parses to int.

The application sees an unexpected type and can issue an error.

---

Bytesizes are more tricky as there typically isn't a concept of a "bytesize type", so the natural thing to do is using a regular int or float. This causes problems in cases such as:

	cache-size = 128 MB # Correct use: parses to int 134217728
	cache-size = 128    # Incorrect use: parses to int 128

In languages such as Go or Rust this tends to be okay because you typically parse to a typed struct:

	// toml package would do something like:
	// type Bytesize int64

	// Users can then use:
	type Config struct {
		CacheSize toml.Bytesize
	}

And it can (IMHO *should*) refuse to parse "128" to a Bytsize, or "128MB" to an int, requiring the use of the correct type.

But in languages like Python or Ruby there is no type information and it just parses to a dict or Hash:

	{ 'cache-size': 134217728 } # cache-size = 128 MB
	{ 'cache-size': 128 }       # cache-size = 128

And now the wrong thing happens, silently. This can also be the case in e.g. Go, because you *can* parse to an untyped map[string]any which has the same problem.

So for this reason, there is a recommendation to ensure applications can see if a suffix is used, and may issue an error.

This also allows "upgrading" keys; for example a real-world example from dnscrypt-proxy:

	log_files_max_size = 10   # Maximum log files size in MB - Set to 0 for unlimited.

With this, they can keep treating the int values as MB, while also allowing newer bytesize values.

Current usage
-------------
I did a survey of some common software, based on "what I could think of" and looking at the top 500 packages in https://popcon.debian.org:

| Config       | Time units                    | Fractions  | Size units                             | Fractions         |
| ------       | ----------                    | ---------  | ----------                             | ---------         |
| [Apache]     | ms ⹋                          |            | †                                      |                   |
| [Caddy]      | ns, us/µs, ms, s, m, h, d     | 1.5h 1h30m | 1000: kB, mB - 1024: kiB, miB          | ?                 |
| [MariaDB]    | †                             |            | For some options; but can't find docs  | ?                 |
| [OpenSSH]    | s, m, h, d, w                 | 1h30m      | ?: K, M, G                             | ?                 |
| [Postfix]    | s, m, h, d, w                 | no         | †                                      | †                 |
| [PostgreSQL] | us, ms, s, min, h, d          | no         | 1024: B, kB, MB, GB, TB (w/ fractions) | 1.5M (round to K) |
| [Redis]      | †                             |            | 1000: k, m, g - 1024: kb, mb, gb       | ?                 |
| [Varnish]    | ms, s, m, h, d, w, y          | 1.5h       | n/a                                    |                   |
| [git]        | †                             |            | 1024: k, m, g                          | no                |
| [haproxy]    | us, ms, s, m, h, d            | no         | 1024: k, m, g (no fractions)           | -                 |
| [nginx]      | ms, s, m, h, d, w, M (30d), y | 1h30m      | ?: k/K, m/M                            | ?                 |
| [samba]      | †                             |            | †                                      |                   |
| [systemd]    | ‡                             |            | n/a                                    |                   |

n/a: Not applicable; there are no settings that could use this unit.
†:   Unit is implied.
‡:   Unit is in the key name (e.g. podInitialBackoffSeconds, JobTimeoutSec).
⹋:   Unit is usually implied, but a few values do allow changing the unit.

In many cases where a unit isn't supported, it would be better if it was. For example:

	Postfix: lmdb_map_size = 16777216        (16MB)
	Samba:   deadtime = 10080                (1 week)
	Apache:  LimitRequestBody =  1073741824  (1GB)

Some are also inconsistent; e.g. Redis's `lfu-decay-time` is in minutes, but `repl-timeout` is in seconds.

Fixes #514
Fixes #912

Previous: #292 #427 #717

[Apache]: https://httpd.apache.org/docs/2.4/mod/core.html#timeout
[Caddy]: https://caddyserver.com/docs/conventions#durations
[MariaDB]: https://mariadb.com/kb/en/server-system-variables/#connect_timeout
[OpenSSH]: https://man.openbsd.org/sshd_config#TIME_FORMATS
[Postfix]: http://www.postfix.org/postconf.5.html
[PostgreSQL]: https://www.postgresql.org/docs/current/config-setting.html
[Redis]: https://redis.io/docs/management/config-file/
[Varnish]: https://varnish-cache.org/docs/trunk/reference/vcl.html#durations
[caddy-size]: https://caddyserver.com/docs/caddyfile/directives/request_body
[git]: https://git-scm.com/docs/git-config
[haproxy]: http://docs.haproxy.org/2.4/configuration.html#2.5
[nginx]: https://nginx.org/en/docs/syntax.html
[samba]: https://www.samba.org/samba/docs/current/man-html/smb.conf.5.html#idm2437
[systemd]: https://www.freedesktop.org/software/systemd/man/systemd.unit.html#JobTimeoutSec=


Skgland:
> Notes on bytesizes
Sizes are always as powers of 2 (times 1024, not 1000). This is what pretty much everyone else does with "1MB" or "1M". An explicit "KiB" is just confusing, and "MB is 1024 and M is 1000" even more so (especially with "m" also meaning "minute"). If you want base-10 units then use digit groupings.

I strongly disagree with this argumentation.
How is KiB confusing it can only mean 2^10 compared to MB which due to common misuse could refer to both  10^6 and 2^20.

> Quote from JEDEC Standard 100B.01, page 8:
> The definitions of kilo, giga, and mega based on powers of two are included only to reflect common usage. IEEE/ASTM SI 10-1997 states "This practice frequently leads to confusion and is deprecated." 
> - https://en.wikipedia.org/wiki/JEDEC_memory_standards#cite_note-6

> The SI prefixes refer strictly to powers of 10. They should not be used to indicate powers of
2 (for example, one kilobit represents 1000 bits and not 1024 bits). The names and symbols
for prefixes corresponding to powers of 2 are the following: kibi (Ki) for 210, mebi (Mi) for
220, gibi (Gi) for 230, tebi (Ti) for 240, pebi (Pi) for 250, exbi (Ei) for 260, zebi (Zi) for 270, yobi
(Yi) for 280, robi (Ri) for 290, and quebi (Qi) for 2100 (see IEC 80000-13 for details).
> - https://www.bipm.org/documents/20126/41483022/SI-Brochure-9-EN.pdf#page=29

The misuse of SI-prefixed shouldn't be propagated further.

> In practice almost no one wants base-10 units.

While I would agree with the sentiment that most would want base-2 units rather than base-10, 
I don't think this is a good argument for using the old/deprecated/ambiguous spelling.
When the recommended/unambiguous spelling is just one extra letter (an extra `i`) away.

CHANGELOG.md
epage, line 1: I thought there was some talk of adding a more formal proposal process. While Rust's RFC process is not ideal (and I've been exploring how to improve it), more formally acknowledging topics certain topics I find a big help for these kind of dicussions, especially if it can be done in a staged approach where higher level details are worked out without distractions of the small details

toml.abnf
```diff
@@ -155,6 +155,27 @@ special-float = [ minus / plus ] ( inf / nan )
...
+)
+bytesize-suffix = (
+    %x42    / ; B
+    %x4B.42 / ; KB
+    %x4D.42 / ; MB
+    %x47.42 / ; GB
+    %x54.42 / ; TB
+    %x50.42   ; PB
+)
+duration = ( dec-int / hex-int / oct-int / bin-int / float-number ) ws duration-suffix
```
epage, line 176: Last you posted, you were skeptical of adding this (https://github.com/toml-lang/toml/issues/514#issuecomment-1732430974) but I'm not seeing any acknowledgment of that and what changed

toml.md
```diff
@@ -547,6 +552,99 @@ sf6 = -nan # valid, actual encoding is implementation-specific
...
+
+    ms
+    s      = 1000 milliseconds
+    m      = 60 seconds
+    h      = 60 minutes
+    d      = 86'400 seconds
+    w      = 7 days
+
+Because there is no specific start time leap days and leap seconds are ignored.
+They may be inserted by the application if appropriate.
```
epage, line 617: Cargo has found months helpful for longer term config, like garbage collection.

The current design of this makes that a bit messier as there is no precise conversion. Cargo decided not to worry, that if someone is doing units in terms of months, they are unlikely looking for precision. However, that is unlikely to generalize to TOML.

One design alternative is to use fixed tags like byte sizes, rather than specifying the conversions between units

toml.abnf
arp242: The *"It is strongly encouraged that TOML implementations allow applications to see if a unit suffix was used [..]"*-paragraph addresses my concerns from 2023.

toml.md
```diff
@@ -547,6 +552,99 @@ sf6 = -nan # valid, actual encoding is implementation-specific
...
+
+Fractions are not allowed for milliseconds:
+
+```toml
+second-and-a-half = 1.5s
+
+very-short        = 0.5ms  # INVALID
+ms-and-a-half     = 1.5ms  # INVALID
+```
```
epage, line 626: > The smallest durations is milliseconds, rather than nanoseconds. There are two reasons for that: 1) at least Python and JavaScript have millisecond as the lowest time unit in their stdlib, and 2) a float64 in ms can accurately hold all reasonable natural numbers: up to 285k years, rather than only 104 days with ns. This makes implementing a bit easier, as you can "just" use a float64 even for cache-time = 52 weeks. For this reason it also disallows fractions for milliseconds.

From a user and parser perspective, special cases `ms` like this is annoying. In other cases, like numbers, the encoding is only a suggestion. Why not the same here?

epage, line 626: Does this need to also prevent `1e-10ms`?

toml.abnf
```diff
@@ -155,6 +155,27 @@ special-float = [ minus / plus ] ( inf / nan )
...
+bytesize = ( dec-int / hex-int / oct-int / bin-int / float-number ) ws size-suffix
```
epage, line 177: Hex and octal durations feels a bit strange.

epage, line 177: From a parser and tooling perspective, it is really frustrating that date times use a space. I was hoping any future work would avoid it. This makes it even worse by allowing 0+ spaces or tabs.

epage, line 177: Like TOML, Rust uses `_` as a digit separator. It also supports it as a unit separator (`1_usize`).  This provides an alternative to users for space (like `T` for datetimes) that provides a visual distinction.

Should that also be allowed here?

epage, line 177: TOML delegates to a spec for datetime, should it do the same for durations, like [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations)?

I can fully understand the answer being "those are not intuitive to read or write". Asking more for completeness.

toml.md
arp242: Personally I'd say that e.g. `8w`, `26w`, or `52w` for "2 months", "half a year", or "a year" should be "good enough" if you don't care about exact precision?

toml.md
```diff
@@ -547,6 +552,99 @@ sf6 = -nan # valid, actual encoding is implementation-specific
...
+    PB     = 1024 TB
+
+Fractions are not allowed for bytes:
+
+```toml
+mb-and-a-half   = 1.5MB
+
+byte-and-a-half = 1.5B   # INVALID
+very-small      = 0.5B   # INVALID
+```
```
epage, line 646: Again, does this also disallow `1e-100KB`?

toml.abnf
arp242: *and binary

I don't disagree, but it was done to generalise `«any number value» «any suffix value»` as much as possible.

toml.md
```diff
@@ -547,6 +552,99 @@ sf6 = -nan # valid, actual encoding is implementation-specific
...
+### Bytesize
+
+Byte sizes are supported with the following suffixes:
+
+    B      byte
+    KB     = 1024 bytes
+    MB     = 1024 KB
+    GB     = 1024 MB
+    TB     = 1024 GB
+    PB     = 1024 TB
```
epage, line 637: Is it intentional that this doesn't cover how to do anything more than parsing? The PR talks about how to deserialize these, including effectively talking about this being a tag.

epage: @eksortso seemed to be against the tag approach at https://github.com/toml-lang/toml/issues/912#issuecomment-1747819454

epage, line 637: > So for this reason, there is a recommendation to ensure applications can see if a suffix is used, and may issue an error.

This is describing a tag. Why not provide a general tag mechanism, allowing for applications to have any units they want and avoid having to get TOML into the details of each type and litigating which types are supported?

toml.abnf
arp242: I'm not sure if I follow why this would be hard to parse? Isn't "skip any number of whitespace" already a feature in many parts of TOML?

That said, `cache-size = 128         MB` is fairly silly and I don't mind making it one (optional) space only.

I also don't have very strong feelings on outright forbidding any whitespace, so that's okay with me too. But it seems to me that `cache-size = 128 MB` is a fairly natural way of writing things and accepting that seems within TOML's stated goal of "minimal configuration file format that's easy to read due to obvious semantics".

toml.md
epage: Part of the intent of this is to avoid people having to do that exact math.

As a reader, `12w` and `26w` is very unintuitive to read.

toml.abnf
epage: That would make sense if this was being done in a fully general way, especially if arbitrary tags are supported (https://github.com/toml-lang/toml/pull/1105#discussion_r4039061821) but this is tailoring to the application already with what units are supported and how small those units can become.

toml.md
arp242: That would significantly affect the portability of TOML documents, with different implementations choosing to support different tags, possibly in different ways.

toml.abnf
epage: >  Isn't "skip any number of whitespace" already a feature in many parts of TOML?

I guess it depends on how you tokenize. For datetimes, My assumption would be to have the unit be part of the token and arbitrary whitespace invites arbitrary lookahead.

More particularly, with the latest Rust `toml` parser,
- I have a very light weight lexer that tries to do no lookahead or know specific type details and I had to special case things for datetimes
- I added error recovery and have to try to detect when someone meant an invalid datetime vs some other situation (like a missing terminator)

toml.md
epage: This sounds less like a factor of the TOML implementation than the application's schema which is already specialized.

epage, line 637: *(starting a thread to make the conversation easier to follow)*

Like @Skgland at https://github.com/toml-lang/toml/pull/1105#issuecomment-5713275209, I'm hearing some strong concerns from Rust project members about this, particularly how confusing it is.  When someone sees `5MB`, users likely can't predict what the syntax prescribes for that and they likely don't know where to look it up. Using `5MiB` at least removes the question, and parsers can check for that and point users in the right direction in their error message.

arp242: No other config file format uses that. No other commonly used tool uses it. Practically all tooling uses M or MB and it's all fine.

No one is going to be confused about it unless you're intentionally looking to be confused by things. It's the de-facto default for practically everything.

> Using 5MiB at least removes the question

And replaces it with the question "what does 5MiB mean? And how does it differ from 5MB?", especially from non-technical people.

arp242: I would not be in favour of having to rely on some external schema to be able to interpret what values in a document might mean.

And if you rely on some external schema then you might as well use `key = "12MB"` in a string, and saving these two characters is barely worth it IMO so then this entire proposal becomes pointless.

Similar previous issue: #603

toml.abnf
arp242: ISO8601 durations are atrocious. I would rather not have any durations in TOML than ISO8601 durations. None of the config files I surveyed use it – I've never seen it used anywhere outside of ical (where it's okay as it's not a human-readable format).

toml.md
arp242: I guess, but personally I'd say it's "good enough".

So the downside is that we'd have to use longer names; we can't use `m` as that's already used for minutes and "`m` for minute, `M` for month". is just confusing IMO. Then we'd also have to accept plurals probably as  `k = 2 day` or `k = 1 days` reads badly. I don't think it's worth it.

toml.md
arp242: Also, nothing here is necessarily preventing arbitrary tags or suffixes. A section saying *"any other [a-zA-Z] text after a number if implementation-dependent, it must issue an error if it doesn't recognise this suffix"*, or something to that effect, could be added in a backwards-compatible way. I would not be in favour of that personally, but the door does exist.

CHANGELOG.md
arp242: I can't recall any discussions about that, just me saying that I feel it would be good if maybe one or two implementations exist before merging non-trivial proposals, and/or maybe some tests in toml-test. Because sometimes actually doing the work may discover problems that were not obvious from the text/ABNF itself. 

I don't see anything about a staged approach in https://github.com/rust-lang/rfcs? Not sure if I missed that? Just 0000-template.md with some questions basically (most of which are answered by the commit message here)?

toml.md
arp242: You kind of need to have a special case here anyway because `0.5B` is completely nonsensical and doesn't mean anything, and if you have that then you might as well do it for `0.5ms` also.

That said, I don't mind making millisecond fractions implementation-dependent, although I would prefer keeping implementation-dependent behaviours at a minimum.

toml.md
epage: `e` in floats would cause problems with that form of extension.

> I would not be in favour of having to rely on some external schema to be able to interpret what values in a document might mean.

I'm confused by this because you already have to.

toml.abnf
epage: `uv` uses ISO 8601 for their cooldowns. Its the only reason I know it exists. It was intentional that I didn't copy it for Cargo's cooldowns (besides we already had a duration format for GC).

toml.md
epage: Huh, I had overlooked minutes and now wonder if users will be confused as to what `m` means

toml.abnf
arp242: Looks like they support both "friendly duration" and ISO8601: https://github.com/astral-sh/uv/blob/5d64ede21e9e835e59ecd85329ebefd43faa8211/docs/concepts/resolution.md?plain=1#L861-L863

toml.md
arp242: If you mean "what types of values can this key accept", then yes. But that's a different thing than "will this particular value even be supported by this particular implementation, and if so, how?"

toml.md
joshtriplett: I think the restriction against using values smaller than milliseconds, just to allow using float64 milliseconds or similar, is conflating implementation limitations with format limitations. By contrast, the spec text for integers says:

> Implementations are free to support any integer size. It's recommended that at least 64-bit signed integers (from −2^63 to 2^63−1) are accepted and handled losslessly. If an integer cannot be represented losslessly, an error must be thrown.

Floats likewise say:
> Implementations are free to support any precision level.

I think it would make sense to 1) allow microseconds and nanoseconds, 2) have similar language for durations, that it is *strongly recommended* to use a storage format that can handle precision of at least integer nanoseconds and duration of at least thousands of years.

Also, JavaScript's "Temporal" should fix all of the JavaScript-specific concerns referenced in the motivation.

toml.md
joshtriplett: @arp242 I really, really sympathize. I hate `MiB` too. But ambiguity is awful.

If the concern is partly not wanting to have to *write* `MiB`, have you considered allowing e.g. 4k, 5M, 10G, 42T? I've widely seen that used as an abbreviation convention that always means power-of-two units. You could have `M` and `MiB`, both meaning power-of-two, and just reject `MB` as ambiguous.

toml.md
cyyynthia: I'm used to `mo` being the classic unit symbol for `month`. If ambiguity is a concern then only accepting `min` and `mo` could be a solution; `min` being the accepted unit symbol for minutes under the SI.

toml.md
cyyynthia: > You kind of need to have a special case here anyway because `0.5B` is completely nonsensical and doesn't mean anything

The problem is that a simple value like `1.4 KB` is also nonsensical under a power-of-two scheme. It translates to 1433.6 bytes, which is nonsense as any non-integer value.

I'd advocate for a simpler approach, requiring parsers to simply round down values in bytes. `1.4 KB` becomes `1433 B`, `0.5 B` becomes `0 B`, etc. The latter remains accepted for the same rationale as why empty strings make for valid quoted keys. The format is meant to be simple and pragmatic, but not necessarily pedantic.

> if you have that then you might as well do it for `0.5ms` also.

I'm not sure I'm too fond of this argument, I don't think adding an extra restriction makes a lot of sense; once again seeing the empty-quoted-key example I'd consider it to fall under the things permitted "because there's not a compelling usability or technical argument to not just let it slide as permitted".

toml.abnf
cyyynthia: > From a parser and tooling perspective, it is really frustrating that date times use a space. I was hoping any future work would avoid it.

But it's prettier.............. 🥺

Jokes aside, I sympathise with the sentiment. Not being able to treat a whitespace as a direct end-of-value marker makes the DFA much more complex and harder to isolate in small areas that are easier to work with and test. That being said, as a user I really appreciate the comfort of being able to use a space here and there.

I think allowing zero or a singular space between the number and the unit is an acceptable tradeoff. It strictly caps the number of lookahead peeks needed to 1 character; and that's likely to be a cold path: there only needs to be a lookup if we encounter a space, but I feel like we're a lot more likely to encounter a newline instead. The branch predictor will likely be able to predict reasonably well that it doesn't need to branch, making the cost virtually null - and the single-char "limit" caps the cost to something imho more than acceptable.

toml.md
joshtriplett: One example of precedent here is `du` and `df`:

> The SIZE argument is an integer and optional unit (example: 10K is 10*1024).  Units are K,M,G,T,P,E,Z,Y,R,Q (powers of 1024) or KB,MB,... (powers of 1000).  Binary prefixes can be used, too: KiB=K, MiB=M, and so on.

Another is `git`; from `man git-config`:

> int: canonicalize values as simple decimal numbers. An optional suffix of k, m, or g will cause the value to be multiplied by 1024, 1048576, or 1073741824 upon input.

toml.abnf
```diff
@@ -155,6 +155,27 @@ special-float = [ minus / plus ] ( inf / nan )
 inf = %x69.6E.66  ; inf
 nan = %x6E.61.6E  ; nan
 
+;; Duration and bytesize suffixes
+
+duration-suffix = (
+    %x6D.73 ; ms
```
ChristianSi: Add '/' after each alternative, except for the last. Right now they would be concatenated instead of becoming alternatives, as intended.

toml.abnf
```diff
@@ -155,6 +155,27 @@ special-float = [ minus / plus ] ( inf / nan )
...
+    %x64    ; d
+    %x77    ; w
+)
+bytesize-suffix = (
+    %x42    / ; B
+    %x4B.42 / ; KB
+    %x4D.42 / ; MB
+    %x47.42 / ; GB
+    %x54.42 / ; TB
+    %x50.42   ; PB
```
ChristianSi: I would suggest dropping the 'B' by itself (it would multiply with 1, with is a no-op), and dropping it also from the other suffixes, reducing them to K, M, G, T, P. This somewhat addresses the problem that SI prefixes are meant to be decimal rather than binary, as these forms would be used as standalone *suffixes* rather than prefixed to a unit. SI would also require lower-case 'k' – maybe it's good to keep the upper case here to further emphasize that these are *not* SI prefixes.

toml.abnf
ChristianSi: I would drop the hex-int / oct-int / bin-int forms, as unlikely to be useful. Adding them in a future version would still be possible if a good use case for hexadecimal durations could be found (which i somewhat doubt), but for now let's keep it simple and start with the forms known to be useful.

toml.abnf
ChristianSi: `size-suffix` should read `bytesize-suffix`.

And i would suggest changing `ws` in both types to `unit-sep = [ %x20 ]`, i.e. to allow only a single optional space before the unit. That would make the types more similar to the date-time types, where a single space or 't/T' are allowed as separators, but not a tab or multiple spaces.

toml.md
```diff
@@ -547,6 +552,99 @@ sf6 = -nan # valid, actual encoding is implementation-specific
 Implementations are free to support any precision level. It's recommended that
 at least IEEE 754 binary64 values are supported.
 
+## Number Suffixes
+
+Any number (integer or float) may be suffixed with a unit to make representing
```
ChristianSi: Tweak accordingly if not allowing octal etc. durations, as suggested above.

toml.md
```diff
@@ -547,6 +552,99 @@ sf6 = -nan # valid, actual encoding is implementation-specific
...
+time durations and byte sizes easier.
+
+To use a suffix, write any number followed by a suffix, optionally separated by
+any amount of whitespace:
```
ChristianSi: Tweak if this a restricted to a single, optional space, as suggested above.

toml.md
```diff
@@ -547,6 +552,99 @@ sf6 = -nan # valid, actual encoding is implementation-specific
...
+
+```toml
+a = 1h
+b = 1 h
+c = 1    h
```
ChristianSi: Personally I'd find this late, dangling letter unexpected and possibly confusing. If we only allow a single optional space here, that issue cannot arise.

toml.md
```diff
@@ -547,6 +552,99 @@ sf6 = -nan # valid, actual encoding is implementation-specific
...
+```
+
+It is strongly encouraged that TOML implementations allow applications to see if
+a unit suffix was used, for example by ensuring it parses to a separate type.
+This is so that applications can distinguish between plain number values and
+suffixed values:
+
+```toml
+sz1 = 128 MB # Correct usage, usually represented as int 134217728.
+sz2 = 128    # Valid TOML but incorrect usage, parses to int 128
```
ChristianSi: I'd change this to something like "possibly incorrect", since of course it might be just as intended.

toml.md
ChristianSi: I would simply omit these remarks about leap days and leap seconds, since, as you say, the issue will never arise with durations as such. It may only arise when a add or subtract a duration from a date or date-time, but such operations are outside of the scope of the TOML spec. Also that adding 1 d to 2024-02-28 should result in 2024-02-29 rather than 2024-03-01 seems obvious enough, so I doubt we have to point it out.

toml.md
```diff
@@ -547,6 +552,99 @@ sf6 = -nan # valid, actual encoding is implementation-specific
...
+    s      = 1000 milliseconds
+    m      = 60 seconds
+    h      = 60 minutes
+    d      = 86'400 seconds
+    w      = 7 days
+
+Because there is no specific start time leap days and leap seconds are ignored.
+They may be inserted by the application if appropriate.
+
+Fractions are not allowed for milliseconds:
```
ChristianSi: Like @epage I would simply drop this prohibition. Half a byte doesn't make much sense, but half a millisecond is perfectly well defined and any implementation that can handle floats can handle it. I t would also be odd to forbid `0.5 ms`, but allow `0.0005 s`, which means the same thing.

toml.md
```diff
@@ -547,6 +552,99 @@ sf6 = -nan # valid, actual encoding is implementation-specific
...
+    KB     = 1024 bytes
+    MB     = 1024 KB
+    GB     = 1024 MB
+    TB     = 1024 GB
+    PB     = 1024 TB
+
+Fractions are not allowed for bytes:
+
+```toml
+mb-and-a-half   = 1.5MB
```
ChristianSi: This can still easily lead to fractional bytes, e.g. `0.3 K[B]` = `307.2 [B]`. I'd make it clear that these size suffixes may use fractional numbers (except for 'B' itself, if we keep but), but should be rounded to the nearest whole int by the implementation, hence `0.3 K[B]` represents `307 [B]`.

toml.abnf
ChristianSi: Again, better drop the drop the hex-int / oct-int / bin-int forms, especially because of the ambiguity. The use case for file sizes may be a bit stronger than for durations, but I don't see it as particularly strong.

toml.md
```diff
@@ -547,6 +552,99 @@ sf6 = -nan # valid, actual encoding is implementation-specific
...
+### Duration
+
+A duration is a time period without a specific start time. Accepted duration
+suffixes are:
+
+    ms
+    s      = 1000 milliseconds
+    m      = 60 seconds
+    h      = 60 minutes
+    d      = 86'400 seconds
```
ChristianSi: I'd write this as 86,400 seconds, as is usual in English.

ChristianSi (changes requested):
I think in general these are very useful additions, but I'd tweak some of the details – see my comments.

toml.md
cyyynthia: Why not simply write it as `24 hours`, like all the previous suffixes being defined using the unit that precedes it?

